PR #178, opened 2014-12-16
Add --scheduler option in manage command to prepare pluggable schedulers

francisbouvier:
Hi,

I've made some changes in order to add pluggable schedulers in the future.
Scheduler is an interface and SwarmScheduler its default implementation.

My idea is to add other scheduler, e.g. an ApiScheduler to connect to external services.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -22,7 +22,7 @@ import (
 
 type context struct {
 	cluster       *cluster.Cluster
-	scheduler     *scheduler.Scheduler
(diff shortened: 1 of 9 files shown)


vieux:
@francisbouvier thank for this, but it's a bit too abstract for us to merge. Be sure we are going to discuss the scheduler API publicly and I'm going to include you in the discussions.

vieux:
See #213 for reference

PR #185, opened 2014-12-18
Persistent store and Virtual ID support.

aluzzardi:
Eventually, in order to support rebalancing, Swarm will need to store the requested state in a persistent way.
Also, rebalancing containers will require them to be re-created on a different node, resulting in a different container ID.

In order to keep IDs consistent across node moves, Swarm will generate a "Virtual ID" that is exposed through the API and guaranteed to never change.
The API takes care of remapping the Virtual ID back into the ID.


vieux:
I'm closing this one for now.

PR #216, opened 2015-01-03
Clarify README

lukemarsden:
Some minor improvements to the README to clarify some stumbling blocks I had first time I tried it:

* Make it easier for Go noobs like me to use `go get`.
* Make it clearer that `$SWARM_TOKEN` should be dynamic.

Recreating this PR because I didn't have signed-off-by in my previous PR's commit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,26 +30,26 @@ This can be achieved by starting Docker with the `-H` flag (e.g. `-H 0.0.0.0:237
 
 ```bash
 # create a cluster
-$ SWARM_TOKEN=`swarm create`; echo $SWARM_TOKEN
-6856663cdefdec325839a4b7e1de38e8
+$ swarm create
+6856663cdefdec325839a4b7e1de38e8 # <- this is your unique <cluster_id>
 
 # on each of your nodes, start the swarm agent
 #  <node_ip> doesn't have to be public (eg. 192.168.0.X),
 #  as long as the other nodes can reach it, it is fine.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -25,16 +30,16 @@ This can be achieved by starting Docker with the `-H` flag (e.g. `-H 0.0.0.0:237
 
 ```bash
 # create a cluster
-$ swarm create
```
vieux: @aluzzardi what do you think ? I'm afraid it'll confuse people even more.

lukemarsden: I wanted to stop people copying and pasting the example cluster ID, because that would be insecure and confusing, right?

My proposed approach does have the problem, however, that many of the commands you need to run get run in a different context (ie, on a different machine). So the environment variable isn't there.

Thinking about it, there are already some manually interpolated variables in this shell transcript. Let me do another commit to suggest a different way of spelling it.

README.md
```diff
@@ -26,25 +31,25 @@ This can be achieved by starting Docker with the `-H` flag (e.g. `-H 0.0.0.0:237
 ```bash
 # create a cluster
 $ swarm create
-6856663cdefdec325839a4b7e1de38e8
+6856663cdefdec325839a4b7e1de38e8 # <- this is your unique <cluster_id>
```
lukemarsden: How's this, and the following lines in the diff, @vieux?

vieux: I like it better, @aluzzardi ?

aluzzardi: Looks good

vieux:
LGTM

PR #218, opened 2015-01-05
Proposal: report error when two Docker's IDs are duplicated

chanwit:
There is a corner case that a VM is duplicated from another. If `Docker` is already initialized on the source machine, the duplicated one will contain the same `~/.docker/key.json` file. This makes `Swarm` not adding the new node but the error report is difficult to trace the problem.

This PR makes `Swarm` a bit more verbose on this case. It reports two IP of machines that share the same ID, so the problem can be easier to fix.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -50,8 +50,7 @@ func (c *Cluster) AddNode(n *Node) error {
 
 	if old, exists := c.nodes[n.ID]; exists {
 		if old.IP != n.IP {
-			log.Errorf("ID duplicated: %s", n.ID)
-			log.Errorf("IP [%s] and [%s] share the same ID", old.IP, n.IP)
+			log.Errorf("ID duplicated. %s shared by %s and %s", n.ID, old.IP, n.IP)
 		}
 		return ErrNodeAlreadyRegistered
 	}
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -48,7 +48,11 @@ func (c *Cluster) AddNode(n *Node) error {
 	c.Lock()
 	defer c.Unlock()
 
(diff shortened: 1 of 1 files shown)


cluster/cluster.go
```diff
@@ -48,7 +48,11 @@ func (c *Cluster) AddNode(n *Node) error {
 	c.Lock()
 	defer c.Unlock()
 
-	if _, exists := c.nodes[n.ID]; exists {
+	if old, exists := c.nodes[n.ID]; exists {
+		if old.IP != n.IP {
+			log.Errorf("ID duplicated: %s", n.ID)
+			log.Errorf("IP [%s] and [%s] share the same ID", old.IP, n.IP)
```
phemmer: my own personal opinion is that this should generate a single log entry. When digging through logs, I find it unpleasant to get a matching entry, and then have to go look at surrounding lines for the complete message (`grep`, get the time, then `less` through the file to that time).

Perhaps something like:

    log.Errorf("ID duplicated. %s shared by %s and %s", n.ID, old.IP, n.IP)

vieux: I agree on the one line log entry.

chanwit:
fixed as suggested by @phemmer and ping @vieux

vieux:
LGTM

@aluzzardi do you see cases when a machine comes back or something and we shouldn't see the warning ?

Otherwise, feel free to merge

aluzzardi:
@vieux I think you may be more familiar than me with this: Wasn't there a case where you tried to `AddNode` anyway and checked the error? I think that code changed massively since you implemented the new discovery system.

vieux:
I think it's good like this.

PR #220, opened 2015-01-06
add more fields to /version

### LINKED ISSUE #219: Expose API Version for compatibility with Ansible
Ansible is a very effective tool for orchestration with Docker containers. Unfortunately, it does not work with Swarm as it checks for the API version in use [1] which is not exposed by Swarm [2]. I think that the Swarm server should respond with it's minimum API compatibility level in a similar manner to the Docker daemon.

[1] https://github.com/ansible/ansible-modules-core/blob/devel/cloud/docker/docker.py#L535
[2] https://gist.github.com/technolo-g/cf82abbb4eda212f26ce

vieux:
close #219

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -17,6 +17,8 @@ import (
 	"github.com/samalba/dockerclient"
 )
 
+const APIVERSION = "1.16"
+
 type context struct {
 	cluster       *cluster.Cluster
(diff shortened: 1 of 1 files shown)


technolo-g:
:+1: This fixes the issue (#219) that I reported.

PR #222, opened 2015-01-06
make use of the "pullImage" parameter of node.Create method

mountkin:
The 'pullImage' parameter of node.Create method is not used currently.


aluzzardi:
Thanks!

I know that the entire `Create` function is not tested at all, but would you mind adding a test for this?

You could do it by using a mock client in `node_test.go` (there are already a few examples on how to achieve this).

mountkin:
Hi, @aluzzardi 
The test is added.

PR #223, opened 2015-01-06
Proposal: ZooKeeper discovery

chanwit:
This PR implements the ZooKeeper discovery service. It's been tested with ZooKeeper 3.3.x and 3.4.x.

### DIFF
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -76,23 +76,40 @@ func (s *ZkDiscoveryService) Watch(callback discovery.WatchCallback) {
 
 func (s *ZkDiscoveryService) Register(addr string) error {
 	newpath := path.Join(s.path, addr)
-	exists, _, err := s.conn.Exists(newpath)
+
+	// check existing for the parent path first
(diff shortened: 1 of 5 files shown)


discovery/zookeeper/zookeeper.go
```diff
@@ -0,0 +1,115 @@
...
+	var nodes []*discovery.Node
+	for _, addr := range addrs {
+		nodes = append(nodes, discovery.NewNode(addr))
+	}
+
+	return nodes, nil
+}
+
+func (s *ZkDiscoveryService) Watch(callback discovery.WatchCallback) {
+	for _ = range time.Tick(time.Duration(s.heartbeat) * time.Second) {
```
vieux: Can you use zookeeper watches instead of a timer here please ?

Like we do in etcd.

chanwit:
Hi @vieux I've implemented a proper watcher.

To confirm that the watcher kicks, Please enable Debug logging  and check a log entry saying:
`[ZK] Watch triggered`.

vieux:
LGTM

vieux:
Thank you!

PR #224, opened 2015-01-06
fix proxy function with TLS

### LINKED ISSUE #212: proxy function doesn't configure TLSClientConfig
When using swarm with TLS I get the error: **" x509: certificate signed by unknown authority "** when using *"docker start"* and it also happens with *"docker logs"*.

The problem is that proxy function in utils.go doesn't use the tlsConfig.

This is what is missing is to configure the tls transport in the proxy function:

`  httpTransport := &http.Transport{`
`    TLSClientConfig: tlsConfig,`
`  }`
`  client := &http.Client{Transport: httpTransport}`

After making these changes "docker start" worked correctly.

## here is full-list of changes I made to get it work:

**api/api.go**

first I needed to relay cluster variable to proxy function

        func proxyContainer(c *context, w http.ResponseWriter, r *http.Request) {
            container, err := getContainerFromVars(c, mux.Vars(r))
            if err != nil {
                    httpError(w, err.Error(), http.StatusNotFound)
                    return
            }

            if err := proxy(c.cluster, container, w, r); err != nil {
            }
        }

**api/utils.go**
 
from the cluster variable I read the TlsConfig

     func proxy(c *cluster.Cluster, container *cluster.Container, w http.ResponseWriter, r *http.Request)   error {

        httpTransport := &http.Transport{
                TLSClientConfig: c.TlsConfig,
        }

        // Use a new client for each request
        client := &http.Client{Transport: httpTransport}

        // RequestURI may not be sent to client
        r.RequestURI = ""

        parts := strings.SplitN(container.Node().Addr, "://", 2)
        if len(parts) == 2 {
                r.URL.Scheme = parts[0]
                r.URL.Host = parts[1]
        } else {
                r.URL.Scheme = "http"
                r.URL.Host = parts[0]
        }

        log.Debugf("[PROXY] --> %s %s", r.Method, r.URL)
        resp, err := client.Do(r)
        if err != nil {
                return err
        }
        w.WriteHeader(resp.StatusCode)
        io.Copy(w, resp.Body)

        return nil
    }

**cluster/cluster.go**

TlsConfig needed to be public for proxy function to be able to access it.

        type Cluster struct {
            sync.RWMutex
            TlsConfig     *tls.Config
            eventHandlers []EventHandler
            nodes         map[string]*Node
        }

  ....

      if err := n.Connect(c.TlsConfig); err != nil {

vieux:
Close #212

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -220,7 +220,7 @@ func proxyContainerAndForceRefresh(c *context, w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	if err := proxy(container, w, r); err != nil {
+	if err := proxy(c.tlsConfig, container, w, r); err != nil {
 		httpError(w, err.Error(), http.StatusInternalServerError)
 	}
 
@@ -249,7 +249,7 @@ func proxyHijack(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := hijack(container, w, r); err != nil {
+	if err := hijack(c.tlsConfig, container, w, r); err != nil {
 		httpError(w, err.Error(), http.StatusInternalServerError)
 	}
 }
--- a/api/api.go
+++ b/api/api.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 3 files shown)


aluzzardi:
Tested, it works!

Without this PR:
```
Error response from daemon: Post https://dev:4242/v1.16/containers/4e/start: x509: certificate signed by unknown authority
FATA[0000] Error: failed to start one or more containers
```

With this PR applied the container starts.

LGTM

PR #227, opened 2015-01-07
add 'balanced' scheduler strategy

phemmer:
This adds a 'balanced' strategy. This is a very basic strategy which will evenly distribute containers among the docker hosts.

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -76,7 +76,7 @@ var (
 	}
 	flStrategy = cli.StringFlag{
 		Name:  "strategy",
-		Usage: "placement strategy to use [binpacking, random]",
+		Usage: "placement strategy to use [balanced, binpacking, random]",
 		Value: "binpacking",
 	}
 	flFilter = cli.StringSliceFlag{
--- a/scheduler/strategy/balanced.go
+++ b/scheduler/strategy/balanced.go
@@ -33,6 +33,10 @@ func (p *BalancedPlacementStrategy) PlaceContainer(config *dockerclient.Containe
 
 		var total = cpuScore + memoryScore + containerScore
 
+		if cpuScore > 100 || memoryScore > 100 {
+			continue
+		}
+
 		scores = append(scores, &balancedScore{node: node, score: total})
 	}
 
--- a/scheduler/strategy/balanced.go
+++ b/scheduler/strategy/balanced.go
@@ -1,0 +1,70 @@
+package strategy
+
+import (
+	"sort"
+
+	"github.com/docker/swarm/cluster"
+	"github.com/samalba/dockerclient"
+)
+
+type BalancedPlacementStrategy struct {
+	overcommitRatio int64
+}
+
+func (p *BalancedPlacementStrategy) Initialize() error {
+	return nil
+}
+
+func (p *BalancedPlacementStrategy) PlaceContainer(config *dockerclient.ContainerConfig, nodes []*cluster.Node) (*cluster.Node, error) {
+	scores := balancedScores{}
+
+	for _, node := range nodes {
+		nodeMemory := node.UsableMemory()
+		nodeCpus := node.UsableCpus()
+
+		// Skip nodes that are smaller than the requested resources.
+		if nodeMemory < int64(config.Memory) || nodeCpus < config.CpuShares {
+			continue
+		}
+
+		var cpuScore = (node.ReservedCpus() + config.CpuShares) * 100 / nodeCpus
+		var memoryScore = (node.ReservedMemory() + config.Memory) * 100 / nodeMemory
+		var containerScore = int64(len(node.Containers())) + 1
+
+		var total = cpuScore + memoryScore + containerScore
+
+		scores = append(scores, &balancedScore{node: node, score: total})
+	}
+
+	if len(scores) == 0 {
+		return nil, ErrNoResourcesAvailable
+	}
+
+	sort.Sort(scores)
+
+	return scores[0].node, nil
+}
+
+type balancedScore struct {
+	node  *cluster.Node
+	score int64
+}
+
+type balancedScores []*balancedScore
+
+func (s balancedScores) Len() int {
+	return len(s)
+}
+
+func (s balancedScores) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+func (s balancedScores) Less(i, j int) bool {
+	var (
+		ip = s[i]
+		jp = s[j]
+	)
+
+	return ip.score < jp.score
+}
--- a/scheduler/strategy/balanced_test.go
+++ b/scheduler/strategy/balanced_test.go
@@ -1,0 +1,256 @@
+package strategy
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/docker/swarm/cluster"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestBalancedPlaceContainerMemory(t *testing.T) {
+	s := &BalancedPlacementStrategy{}
+
+	nodes := []*cluster.Node{}
+	for i := 0; i < 2; i++ {
+		nodes = append(nodes, createNode(fmt.Sprintf("node-%d", i), 2, 1))
+	}
+
+	// add 1 container 1G
+	config := createConfig(1, 0)
+	c1node, err := s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.NoError(t, c1node.AddContainer(createContainer("c1", config)))
+
+	assert.Equal(t, c1node.ReservedMemory(), 1024*1024*1024)
+
+	// add another container 1G
+	config = createConfig(1, 1)
+	c2node, err := s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.NoError(t, c2node.AddContainer(createContainer("c2", config)))
+
+	assert.Equal(t, c2node.ReservedMemory(), 1024*1024*1024)
+
+	// check that both containers ended on different nodes
+	assert.NotEqual(t, c1node.ID, c2node.ID, "")
+	assert.Equal(t, len(c1node.Containers()), len(c2node.Containers()), "")
+}
+
+func TestBalancedPlaceContainerCPU(t *testing.T) {
+	s := &BalancedPlacementStrategy{}
+
+	nodes := []*cluster.Node{}
+	for i := 0; i < 2; i++ {
+		nodes = append(nodes, createNode(fmt.Sprintf("node-%d", i), 1, 2))
+	}
+
+	// add 1 container 1CPU
+	config := createConfig(0, 1)
+	c1node, err := s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.NoError(t, c1node.AddContainer(createContainer("c1", config)))
+
+	assert.Equal(t, c1node.ReservedCpus(), 1)
+
+	// add another container 1CPU
+	config = createConfig(0, 1)
+	c2node, err := s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.NoError(t, c2node.AddContainer(createContainer("c2", config)))
+	assert.Equal(t, c2node.ReservedCpus(), 1)
+
+	// check that both containers ended on different nodes
+	assert.NotEqual(t, c1node.ID, c2node.ID, "")
+	assert.Equal(t, len(c1node.Containers()), len(c2node.Containers()), "")
+}
+
+func TestBalancedPlaceContainerHuge(t *testing.T) {
+	s := &BalancedPlacementStrategy{}
+
+	nodes := []*cluster.Node{}
+	for i := 0; i < 50; i++ {
+		nodes = append(nodes, createNode(fmt.Sprintf("node-%d", i), 1, 1))
+	}
+
+	// add 100 container 1CPU
+	for i := 0; i < 100; i++ {
+		node, err := s.PlaceContainer(createConfig(0, 1), nodes)
+		assert.NoError(t, err)
+		assert.NoError(t, node.AddContainer(createContainer(fmt.Sprintf("c%d", i), createConfig(0, 100))))
+	}
+
+	// add 100 container 1G
+	for i := 100; i < 200; i++ {
+		node, err := s.PlaceContainer(createConfig(1, 0), nodes)
+		assert.NoError(t, err)
+		assert.NoError(t, node.AddContainer(createContainer(fmt.Sprintf("c%d", i), createConfig(1, 0))))
+	}
+}
+
+func TestBalancedPlaceContainerOvercommit(t *testing.T) {
+	s, err := New("balanced")
+	assert.NoError(t, err)
+
+	nodes := []*cluster.Node{createNode("node-1", 0, 1)}
+	nodes[0].Memory = 100
+
+	config := createConfig(0, 0)
+
+	// Below limit should still work.
+	config.Memory = 90
+	node, err := s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.Equal(t, node, nodes[0])
+
+	// At memory limit should still work.
+	config.Memory = 100
+	node, err = s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.Equal(t, node, nodes[0])
+
+	// Up to 105% it should still work.
+	config.Memory = 105
+	node, err = s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.Equal(t, node, nodes[0])
+
+	// Above it should return an error.
+	config.Memory = 106
+	node, err = s.PlaceContainer(config, nodes)
+	assert.Error(t, err)
+}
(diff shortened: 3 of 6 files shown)


MrGossett:
+1

vieux:
@phemmer don't you need the overcommit parameter here as well ?

Docker doesn't return the exact amount of memory available, a little less (the system returns a little less)

phemmer:
Not sure. I opted to leave it out as I thought the strategy should be basic.

The idea behind the binpacking algorithm is to fill a single 'bin' to it's max, and then continue to the next one, so overcommit is necessary there as you need to know maximum capacity (I'm sure you're aware of this, just adding here for completeness).

This algorithm is meant to be a lot simpler in that it evenly distributes everything, so a maximum isn't as important, and it won't refuse to start a container due to lack of available resources. Because it won't refuse to start a container, overcommit doesn't make much sense.

My thoughts on why it shouldn't be limited is that I think it would be unexpected for swarm to refuse to start containers, as this is not how docker behaves. There's also currently nothing that warns the user that they are approaching the maximum capacity. So a sudden refusal to start containers can be very bad.

I can easily add this in if desired. Perhaps with an insanely high overcommit so the same effect is reached.

vieux:
This thing is, if all the machines in your cluster and you want to run a container with -m 2g with your strategy, it won't work.

phemmer:
I think you have an incomplete thought: "if all the machines in your cluster" what?

vieux:
@phemmer sorry, updated

vieux:
but we are thinking about moving the overcommit outside of the strategies, it would be a top level thing.

phemmer:
Ah, yes, because of the pre-check node filter. Valid point, I'll add it in.

phemmer:
Ok, added in.

However one comment on the overcommitness used by swarm is that it has a critical difference from linux's overcommit. In linux, `100` overcommit means no overcommit, and to use exactly how much memory is available. Swarm is treating `0` as no overcommit. The advantage of using `100` as no overcommit is that if you want to prevent the system from launching something that uses *nearly* all the available resource, you can, by setting the overcommit value to `90` or so.
I think this would be a good idea for swarm to adopt, but I kept this strategy using the same behavior as binpack for consistency.

vieux:
if we use this, it means our default whould be `105` ?

phemmer:
You could keep the scale if you wanted, and use `1.00` as no overcommit, or you can use `100`.
So either `1.05` or `105` would be the equivalent of the current `0.05`.

chanwit:
+1

I'm about to propose the "least running containers" strategy to balance my cluster. Hopefully I can use yours instead of inventing a new one.

phemmer:
Well there is one thing that might be unexpected. This strategy doesn't consider whether the container is running or not. If you have 10 stopped containers on one node, but 0 running, and 5 running containers on another node, it will place the new container on the one with 5 containers.
This was done as it's how the binpack strategy behaves. Though it might be a good idea to add a flag to control the behavior.

phemmer:
Will rebase for #228 and work on adding some tests.

vieux:
thanks @phemmer, sorry about that.

vieux:
@phemmer could you please add some tests similar to binpacking_test.go ?

flags.go
```diff
@@ -57,7 +57,7 @@ var (
 	}
 	flStrategy = cli.StringFlag{
 		Name:  "strategy",
-		Usage: "placement strategy to use [binpacking, random]",
+		Usage: "PlacementStrategy to use [balanced, binpacking, random]",
```
vieux: While you add the tests, can you switch this back to `placement strategy` (lowercase + space)

phemmer:
Sorry for the delay. Rebased, node.go fixed, and tests added.

phemmer:
I'm also still uncertain about the overcommit thing. As mentioned earlier, this scheduler will allow you to add 2x 2gb containers to a 3gb node, and this isn't how binpack behaves.
I don't much like the idea of having it refuse to launch containers if they exceed the limits, but the reasons for doing so are pretty strong, and consistency is another big factor.

I see a few possible solutions:
* Leave as is (perhaps open a ticket to continue discussion on the matter)
* Enforce the limit, and add an option to toggle it.
* Have the binpack scheduler fall back to the balanced scheduler if all nodes are full, so that it gets the same behavior.

aluzzardi:
@phemmer Can you please clarify?

phemmer:
@aluzzardi Beyond the explanation already provided, i'm not sure how. Perhaps an example:

Lets say you have 2 nodes, each with 1.5gb memory.
You start one container using 1gb, it goes to node A.
You start another container using 1gb, it goes to node B.
You start a third container using 1gb, it goes to node A.
Node A's total reserved memory is now 2gb.

aluzzardi:
@phemmer With how much overcommit?

phemmer:
none, or even the default `0.05`. But less than 150%.

vieux:
Ok, In my opinion, it should work the same way as binpacking: the 3rd commit shouldn't start because there is not enough ressource.

// what do you think of the name "spread" instead of balanced ?

phemmer:
But then why doesn't docker behave that way? Docker will happily let you launch a container even if it exceeds the node's capacity.

As for the name, i'm not too fond of "spread", as I think the term is rather ambiguous. The "random" strategy spreads containers mostly evenly based on count, so the name should differentiate how the strategy behaves.

aluzzardi:
@phemmer Bin packing won't let you do that, and so shouldn't Balanced - right?

phemmer:
@aluzzardi bin packing wont let you because its part of the bin packing algorithm. The whole concept of bin packing is to fill a bin without overflowing it.
If the argument is that because binpacking does this, then so should other strategies, then why doesn't random do it?

phemmer:
I'll update the PR to add this behavior as an option, but I'm still in favor of deafulting to 'off'. My reasons are up above, but I'll just recap:
* Docker does not do this
* There is nothing that warns the user they are approaching maximum capacity.
* It would violate the rule of least surprise. I do not think admins would expect docker to start refusing to launch containers.

aluzzardi:
@phemmer Swarm won't refuse to run containers if you don't ask for any kind of specific resources.

The goal of Swarm is to handle cluster resources, therefore it has to refuse to run containers if there are not enough resources available, in the same way that it refuses to schedule a container when requesting a port that is not available.

phemmer:
But there's still a use for requesting resources. If I start a container with 10gb memory, and I have 2 nodes in my cluster, one with 5gb, and another with 15gb, it should chose the one with 15gb.
Then there are many reasons I may want it to launch another 10gb container. Perhaps the container only spikes up to 10gb, and doesn't use it all the time, or perhaps the 2 10gb containers I launched share memory, so they're not using 20gb total.

aluzzardi:
There might be eventually a global flag to discard resource usage, but right now I believe it is a sensible default to take resource usage into account and refuse to start containers when not enough resources are available.

abronan:
I'm late to the party but throwing my 2 cents :smile: 

@phemmer I think that a Clustering system should not let you do that because there are critical performance requirements for containers running in a production environment. For example, I don't want my Web containers to feel unresponsive because someone else decided to use the Swarm manager to deploy a container and ignore the resource usage, thus delivering far higher latency or worse, crashing my application in some way because it can't handle the load anymore or does not have enough memory.

Imho discarding resource usage should be an option (as proposed by @aluzzardi, with a global flag for example) used for development environment clusters and either your entire cluster allows to overcommit on resource usage or you use the default behavior by refusing to deploy a container if those resources are not available.

> Perhaps the container only spikes up to 10gb, and doesn't use it all the time, or perhaps the 2 10gb containers I launched share memory, so they're not using 20gb total.

Agreed but this is a sensitive subject with a solution that is far from being trivial. If you want to support this use case, you'll need to constantly analyze resource usage at peak hours and make sense of the data to take a decision (or make predictions) on which container will be running or not at a specific time of the day. If you launch 2 containers using 10GB on a 15GB machine, one of the two containers will need to pause for the other to handle peak traffic efficiently.

Overall, being strict and defensive on the resource usage wastes some resources, but ensures that the entire Cluster is running containers with critical production apps safely. Or we'll see containers crashing every now and then because they do not have enough memory left for example.

phemmer:
@abronan 

> I don't want my Web containers to feel unresponsive because someone else decided to use the Swarm manager to deploy a container and ignore the resource usage

Swarm won't place it on an over-capacity node unless it has no other choice. If you're not monitoring your cluster to ensure that it has enough resources to run the containers your telling it to, or if you've given control of your cluster to someone who is not appropriately using it, that's not swarm's issue.

> you'll need to constantly analyze resource usage at peak hours

You should be doing this anyway. In fact you should be doing it 24/7, not just at peak hours. This also isn't territory restricted to clustering. What happens in a non-clustered system, and you're running an application, and that application starts using more resources than it usually does? You're going to need to move your app to a bigger box.

> Overall, being strict and defensive on the resource usage wastes some resources, but ensures that the entire Cluster is running containers with critical production apps safely.

Actually what you propose can do the exact opposite. It can cause service interruption. In mission critical systems, if you need to launch container X, and swarm refuses to start it because the system is full, your service could be completely down. I would much rather have service A be a little slow, than have service B down entirely because my clustering system refused to start it.

&nbsp;

I'm not saying your arguments don't have merit. They do, and I'm well aware of them. But it's my opinion you're wanting swarm to do things that aren't necessarily in it's realm of responsibility.

----

On another subject, the 'overflow' option does not make sense as a global option, as it does not apply to all strategies. As mentioned with the binpacking strategy, you cannot have overflow. It's core to the algorithm.

abronan:
@phemmer I fully understand your arguments. This is really subject to personal preferences anyway. Thus the proposal to have a global flag (even applied only to a subset of strategies that are using overflow, excluding bin-packing that does not use it, if this makes more sense).

> Swarm won't place it on an over-capacity node unless it has no other choice

Swarm will then put you into risk by running a container on a node that does not have the available resources. Shouldn't Swarm at least give you a safety-belt? An app being slow is not a issue, we could live with that and we already did using Docker while deploying dozens of containers on a single machine. But I'm mostly thinking about the worst case here.

> If I tell swarm to put the system over capacity, it's my responsibility to deal with the consequences.

True, but shouldn't Swarm at least avoid you getting into a bad situation where some of your apps could go unstable or crash? Isn't there a way to give the user the choice to say: "I'm well aware of the risk and I want my entire Cluster to behave like a single Docker machine"? Thus if there is no other choice, let me deploy a container (or dozens) even if the resources are not available.

> Actually what you propose can do the exact opposite. It can cause service interruption.

True but I was kind of foreseeing a replication parameter that would allow containers to be replicated a given amount of time through agents (kind of what Kubernetes is doing with Pods). This way you don't run into service interruption. But I'm off-topic here.

--------
To sum-up for me this would be: 

- Shouldn't Swarm protect the user against bad Cluster resource usage? 
- If not what are the arguments against that behavior?
- Is there a way to give the choice to the user to hold tight with some kind of safety-belt or get rid of it?

phemmer:
Commit has been added to disable overflow. Unfortunately the ability to have per-strategy options has been removed, so the behavior cannot be controlled. Meaning that I will be running my own fork of swarm until the issue gets resolved :-(

aluzzardi:
@phemmer Thank you for raising those points.

I understand what you are saying and your points are valid, but early on in the design of Swarm we agreed that we will guarantee resource allocation across the cluster for different reasons.

Here are some of them:

> But there's still a use for requesting resources. If I start a container with 10gb memory, and I have 2 nodes in my cluster, one with 5gb, and another with 15gb, it should chose the one with 15gb.

But if the node that has 15gb has already another container running, constantly using its 10gb allocation, it actually only has 5gb available as the other one. What would be the point in requesting resources that cannot be honored?

> Actually what you propose can do the exact opposite. It can cause service interruption. In mission critical systems, if you need to launch container X, and swarm refuses to start it because the system is full, your service could be completely down. 

In that case, you wouldn't start your containers by explicitly requesting resources.

I find that specifying memory requirements affects the container in the same way it affects swarm:

If you request to create a container and there not enough resources in the cluster, swarm will refuse to schedule and it may cause service disruption.

If you try to use more memory than you requested, one of your processes in the container will end up OOM'ing, which may cause a service disuption.

vieux:
In any case, we going to postpone the merge of this PR after the RC, so we can try to find a solution that works for everybody.

rgbkrk:
Looking forward to this one so long as it doesn't overcommit. For JupyterHub's [dockerspawner](https://github.com/jupyter/dockerspawner) and [tmpnb](https://github.com/jupyter/tmpnb) we expect to fill up some expected amount of memory, pooling them in advance. We'd rather not see it overcommit. If it's configurable though, that's a different thing.

vieux:
@jhamrick are you using this PR as is, or tweaked in some ways ?

jhamrick:
@vieux I'm not actually using this PR; I just made a small modification to the binpacking strategy (just reverses the sort order, so it does more of a round robin thing). I may switch to this strategy once the PR is merged, though.

mota:
If I may suggest, why not considering composability through a pipeline of applicable strategies?

That is, each strategy returns a set of potential candidates that are passed to the next one. In fine, the first element of the resulting set (into which members are considered equivalent) is chosen.

It would add a little more complexity within each strategy but would remove the need to repeat code.

In the cli point of view, this could be expressed as such:
```
swarm manage --strategies=binpacking,balanced ...
```
in which case the binpacking strategy would be applied before the balanced strategy.

tnachen:
@mota I'm not sure strategies really compose since they often have competing priorities , so at least I don't see a good use case for it yet. 
And IMO it becomes harder to implement strategy, since each strategy shouldn't simply take a list of candidates from the last one, lots of them still has look at the global state (ie: balanced needs to balance across cluster) as a whole and then trying to see if any of the ones matches the passed in candidates.
I'm more in favor with a single strategy, and if really need to support configurations per strategy that favors different scenarios.

mota:
@tnachen You're right, strategies interfere with each other.

Nevertheless, I don't agree with your idea of a one monolithic strategy that is applied statefully.

In the first place because filters are applied before strategies. They cannot operate on a cluster as a whole, but only on a subset of it.

Secondly because I think it is best to let the user chose which priority he or she values the most.

However, I'd like to revise my paper and add a little subtlety regarding the implementation I propose.

The idea would be not to associate to each node a pipeline score at each pipeline execution.
Each strategy would still be free to remove unsatisfying candidates, but would add to the remaining nodes a value to their score.

At the end of the pipeline process, the node with the best score is chosen.

That way, the `balanced` strategy would be implemented such as it only adds to each node's score a value of its own.
Therefore, the current implementation of `balanced` could be either matched with `balanced,binpacking` or `binpacking,balanced`.

tnachen:
Since a filter is user defined, it's supporting cases where users explicitly can prune the selection to what they want, and that sunset is what I refer to as global state.

I think what's missing in your proposal is a concrete use case that deems this necessary, I'm not against the idea, but I'm hoping to keep the scheduler simple to begin with as it can become very hard to reason with what you described.

tnachen:
And balanced with binpack composed
Together doesn't make much sense to me too.

phemmer:
@mota Perhaps you can provide some examples of how you expect a combination of schedulers to work. Because binpacking is pretty much the complete opposite of balanced. I don't see how they can co-exist.

mota:
@phemmer sure thing.

Maybe my view is flawed but as things evolve, I don't see strategies as a code of conduct but more as a best-effort behavior.

First thing, I think the elimination of incapable nodes in terms of resource usage should be made in the filtering phase, not in the strategy phase as it is done currently.

Thus, what I propose is simply to tune the behavior of the scheduler. Let's say you want to favor spreading instead of stacking, you should use the `balanced` behavior. If in contrary you favor stacking, then you use the `binpacking` behavior. And finally, if you want to reach a best effort candidate, you apply both.

Tell me if I'm wrong but as I see it, the implementation of `balanced` you provide is pretty much the `binpacking` strategy with one added dimension.

bacongobbler:
The [strategy README](https://github.com/docker/swarm/blob/master/scheduler/strategy/README.md) should be updated as well

scheduler/strategy/balanced.go
```diff
@@ -0,0 +1,74 @@
...
+		}
+
+		scores = append(scores, &balancedScore{node: node, score: total})
+	}
+
+	if len(scores) == 0 {
+		return nil, ErrNoResourcesAvailable
+	}
+
+	sort.Sort(scores)
```
bacongobbler: Instead of re-implementing the wheel here, you can just call `sort.Sort(sort.Reverse(scores))` on a `scores` structure. That should clean up a lot of the boilerplate

aluzzardi:
@phemmer I just merged https://github.com/docker/swarm/pull/458 since it's more recent and merges properly (there have been many changes to the scheduler since this PR was opened).

Does it fits your needs?

vieux:
ping @phemmer  ?

phemmer:
Sorry, haven't had a chance to actually build and use it. But from looking at the implementation, this appears to have the same effect as the strategy in this PR, so I think it's good.

vieux:
cool, @phemmer I'm closing this, please comment if you have any issue.

PR #228, opened 2015-01-08
Move overcommit outside of binpacking

vieux:
Related to comments in #227

### DIFF
--- a/cluster/node.go
+++ b/cluster/node.go
@@ -21,14 +21,14 @@ const (
 	requestTimeout = 10 * time.Second
 )
 
-func NewNode(addr string, overcommitRatio int64) *Node {
+func NewNode(addr string, overcommitRatio float64) *Node {
 	e := &Node{
 		Addr:            addr,
 		Labels:          make(map[string]string),
 		ch:              make(chan bool),
 		containers:      make(map[string]*Container),
 		healthy:         true,
-		overcommitRatio: overcommitRatio - 100,
+		overcommitRatio: int64(overcommitRatio * 100),
 	}
 	return e
 }
--- a/cluster/node.go
+++ b/cluster/node.go
@@ -270,11 +270,11 @@ func (n *Node) ReservedCpus() int64 {
 	return r
 }
 
-func (n *Node) OverCommitedMemory() int64 {
+func (n *Node) UsableMemory() int64 {
 	return n.Memory + (n.Memory * n.overcommitRatio / 100)
 }
 
-func (n *Node) OverCommitedCpus() int64 {
+func (n *Node) UsableCpus() int64 {
 	return n.Cpus + (n.Cpus * n.overcommitRatio / 100)
 }
 
--- a/cluster/node.go
+++ b/cluster/node.go
@@ -21,13 +21,14 @@ const (
 	requestTimeout = 10 * time.Second
 )
 
-func NewNode(addr string) *Node {
+func NewNode(addr string, overcommitRatio int64) *Node {
 	e := &Node{
-		Addr:       addr,
-		Labels:     make(map[string]string),
-		ch:         make(chan bool),
-		containers: make(map[string]*Container),
-		healthy:    true,
+		Addr:            addr,
+		Labels:          make(map[string]string),
+		ch:              make(chan bool),
+		containers:      make(map[string]*Container),
+		healthy:         true,
+		overcommitRatio: overcommitRatio - 100,
 	}
 	return e
 }
@@ -43,11 +44,12 @@ type Node struct {
 	Memory int64
 	Labels map[string]string
 
-	ch           chan bool
-	containers   map[string]*Container
-	client       dockerclient.Client
-	eventHandler EventHandler
-	healthy      bool
+	ch              chan bool
+	containers      map[string]*Container
+	client          dockerclient.Client
+	eventHandler    EventHandler
+	healthy         bool
+	overcommitRatio int64
 }
 
 // Connect will initialize a connection to the Docker daemon running on the
@@ -266,6 +268,14 @@ func (n *Node) ReservedCpus() int64 {
 	}
 	n.RUnlock()
 	return r
+}
+
+func (n *Node) OverCommitedMemory() int64 {
+	return n.Memory + (n.Memory * n.overcommitRatio / 100)
+}
+
+func (n *Node) OverCommitedCpus() int64 {
+	return n.Cpus + (n.Cpus * n.overcommitRatio / 100)
 }
 
 func (n *Node) Create(config *dockerclient.ContainerConfig, name string, pullImage bool) (*Container, error) {
--- a/flags.go
+++ b/flags.go
@@ -50,10 +50,10 @@ var (
 		Name:  "tlsverify",
 		Usage: "use TLS and verify the remote",
 	}
-	flOverCommit = cli.IntFlag{
+	flOverCommit = cli.Float64Flag{
 		Name:  "overcommit, oc",
 		Usage: "overcommit to apply on resources",
-		Value: 105,
+		Value: 0.05,
 	}
 	flStrategy = cli.StringFlag{
 		Name:  "strategy",
--- a/flags.go
+++ b/flags.go
@@ -50,10 +50,15 @@ var (
 		Name:  "tlsverify",
 		Usage: "use TLS and verify the remote",
 	}
+	flOverCommit = cli.IntFlag{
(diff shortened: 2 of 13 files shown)


vieux:
sorry @phemmer you'll have to update your PR again, but I think it's for the best.

ping @aluzzardi

flags.go
```diff
@@ -50,10 +50,15 @@ var (
 		Name:  "tlsverify",
 		Usage: "Use TLS and verify the remote",
 	}
+	flOverCommit = cli.IntFlag{
+		Name:  "overcommit, oc",
+		Usage: "OverCommit to apply on resouces",
```
tnachen: s/resouces/resources/g
Also is this the help text that's going to show up in the cli? If it is I think more description what and why are we overcommiting will be nice.

manage.go
```diff
@@ -98,7 +98,7 @@ func manage(c *cli.Context) {
 		}
 	}()
 
-	s, err := strategy.New(c.String("strategy"))
+	s, err := strategy.New(c.String("strategy"), int64(c.Int("overcommitratio")))
```
tnachen: the name is either overcommit or oc right?

scheduler/strategy/binpacking.go
```diff
@@ -17,10 +16,9 @@ type BinPackingPlacementStrategy struct {
 	ratio int64
 }
 
-func (p *BinPackingPlacementStrategy) Initialize(opts string) error {
-	overcommitRatio, err := strconv.ParseFloat(opts, 64)
-	p.ratio = int64(overcommitRatio * 100)
-	return err
+func (p *BinPackingPlacementStrategy) Initialize(ratio int64) error {
+	p.ratio = ratio
```
tnachen: I wonder if overcommitRatio is more obvious than ratio

tnachen:
@vieux hope you don't mind me reviewing the commits, I'm trying to get to know what's going on swarm

vieux:
@tnachen, not at all, thanks!

manage.go
vieux: yes, good catch I always forgot this, this bad design from the cli pkg IMO

flags.go
vieux: @tnachen, what would you suggest ?

```
NAME:
   manage - manage a docker cluster

USAGE:
   command manage [command options] [arguments...]

OPTIONS:
   --strategy 'binpacking'				PlacementStrategy to use [binpacking, random]
   --filter, -f '--filter option --filter option'	Filter to use [constraint, health, port]
   --discovery 						DiscoveryService to use [token://<token>, etcd://<ip1>,<ip2>/<path>, file://path/to/file, consul://<addr>/<path>] [$SWARM_DISCOVERY]
   --host, -H '--host option --host option'		ip/socket to listen on [$SWARM_HOST]
   --heartbeat, --hb '25'				time in second between each heartbeat
   --overcommit, --oc '105'				OverCommit to apply on resources
   --tls						Use TLS; implied by --tlsverify=true
   --tlscacert 						Trust only remotes providing a certificate signed by the CA given here
   --tlscert 						Path to TLS certificate file
   --tlskey 						Path to TLS key file
   --tlsverify						Use TLS and verify the remote
   --api-enable-cors, --cors				enable CORS headers in the remote API
```

manage.go
vieux: sorry,  I read to fast, you're right.

flags.go
tnachen: It actually isn't really a overcommit flag after all, since you're no longer padding the resources but actually assigning the final resources to schedule with the ratio. If I put in 50 I'm actually under instead of overcommitting. 

I wonder if instead of putting in a 105 value, we should just let the user put in 5, and this extra 5% is the overcommit resources so we don't ever have the possibility to go under.

Then in the comment it can say "percentage of container resources to overcommit"

flags.go
```diff
@@ -50,10 +50,15 @@ var (
 		Name:  "tlsverify",
 		Usage: "Use TLS and verify the remote",
 	}
+	flOverCommit = cli.IntFlag{
+		Name:  "overcommit, oc",
+		Usage: "OverCommit to apply on resources",
+		Value: 105,
```
tnachen: Wouldn't this be better if we prevent under committing? Overcommit actually takes the percentage of resources to overcommit, and in the code we simply overcommit + 100?

phemmer:
I think with this change, we should remove any reference to overcommit from the strategy itself, and put it in `cluster.Node`. For example, in the binpacking strategy, the only place overcommit is used is to set `nodeMemory`. Instead we can just create `(*Node) Memory()` (or `MemoryOvercommit()`) and call it, which will give us the memory value with overcommit baked in.

flags.go
phemmer: @tnachen part of this change stems from [this comment](https://github.com/docker/swarm/pull/227#issuecomment-69090266) in #227. The concept is based off linux's memory overcommit.

This bit of the change was actually made to allow undercommit. There are cases where you might not want a container to get allocated to a box where the container will consume almost all of the box's resources. In this case you would set overcommit to a value less than `100`.

tnachen: If undercommit is actually desired we definitely should change the flag to not call it overcommit. 

However, I'm not really sure I understand your motivation though, if I don't want a container to be consume all of the box's resources, we should instead advertise less than the maximum memory/cpu instead of undercommiting right? Since lower resource usage can also lead to more binpacked containers to fill the gaps.

vieux:
@phemmer good idea

flags.go
vieux: @aluzzardi what do you think ? do we want to allow undercommit ?

aluzzardi:
@vieux: I'm not sure about undercommiting.

There clearly is a use case where you don't want to use all of of the machine's memory (especially if it's a mix of containers and non-containers).

However, ideally this should be provided on a machine-by-machine basis rather than globally on a cluster, but I don't think there is a nice way to do that (memory available is reported by Docker and we cannot add an extra flag to override that).

vieux:
@aluzzardi ok, then do want to revert to use `1.05` but keep the overcommit outside like in the PR ?

tnachen:
@aluzzardi @vieux or in the scheduler there can be a reserved space per machine that is auto deducted from each node's reported resources.

Undercommit definitely won't address the resource reservation problem, simply because it doesn't stop scheduling another container on the resources you saved.

phemmer:
@tnachen 
> Undercommit definitely won't address the resource reservation problem, simply because it doesn't stop scheduling another container on the resources you saved

actually it does: https://github.com/docker/swarm/blob/master/scheduler/strategy/binpacking.go#L51
This line will make the strategy only pick a node if the new container, plus existing containers, don't exceed the node's capacity. (that's core to the binpack algorithm)

tnachen:
@phemmer so help me understand if I'm missing something here.

Let's say if we have a node that has 2gb mem 2 cpus, and I have 4 containers each requires 1 gb 1 cpu.
Normally I can only run 2 containers without any adjustment to resources.

But if I undercommit the task resources (flag Overcommit = 50), then it becomes 0.5 gb and 0.5 cpu per container, won't I actually launch 4 containers?

phemmer:
@tnachen No, if you undercommit, you will only be able to launch 1 of your containers. Undercommit pretends as if the box has less resources than it actually does. So if you box has 2gb, an undercommit of 90% will act as if your box has 1.8gb, and the scheduler will refuse to launch the second container.

tnachen:
I see, i thought it's applied on the container's resources (which doesn't make much sense).
Anyhow I think a global undercommit or overcommit could be useful but also a bit limited, since a single ratio can mean different resources for different nodes that has different resources.

cluster/node.go
```diff
@@ -267,6 +269,14 @@ func (n *Node) ReservedCpus() int64 {
 	return r
 }
 
+func (n *Node) OverCommitedMemory() int64 {
```
aluzzardi: How about simply `AvailableMemory`? Today we are doing overcommitting, tomorrow we may be doing something different.

It might be ambiguous as it might look like "free memory" rather than "how much total memory I can use", but I'd rather not include overcommit in the name.

phemmer: How about "unreserved"? I really dislike "available" for the confusion you mentioned.

vieux:
@aluzzardi I removed undercommit from this PR and kept 0.05. Let's discuss about undercommit in another PR.

Can you merge this if it LGTY (it's blocking #227)

aluzzardi:
LGTM

PR #231, opened 2015-01-09
Cleanup: zookeeper discovery's watcher

chanwit:
This PR is to improve the ZooKeeper's discovery a bit by calling callback 1 more time before getting into the event channel loop. As zk.ChildrenW() returns a list of nodes for the first time, `Swarm` does not need to wait until the next event get fired.

A unit test is also added.

### DIFF
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -56,21 +56,30 @@ func (s *ZkDiscoveryService) Fetch() ([]*discovery.Node, error) {
 		return nil, err
 	}
 
-	var nodes []*discovery.Node
+	return s.createNodes(addrs), nil
+}
+
+func (s *ZkDiscoveryService) createNodes(addrs []string) (nodes []*discovery.Node) {
+	nodes = make([]*discovery.Node, 0)
+	if addrs == nil {
+		return
+	}
+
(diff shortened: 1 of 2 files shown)


discovery/zookeeper/zookeeper.go
```diff
@@ -56,21 +56,30 @@ func (s *ZkDiscoveryService) Fetch() ([]*discovery.Node, error) {
...
 func (s *ZkDiscoveryService) Watch(callback discovery.WatchCallback) {
 
-	_, _, eventChan, err := s.conn.ChildrenW(s.path)
+	addrs, _, eventChan, err := s.conn.ChildrenW(s.path)
 	if err != nil {
 		log.Debugf("[ZK] Watch aborted")
 		return
 	}
+	nodes := s.createNodes(addrs)
+	callback(nodes)
```
vieux: do we need to do this here ?

maybe swarm should call watch before calling fetch in the first place ?

chanwit: Well, I would say it's not a direct swarm issue, actually. This is specific to the ZooKeeper's API.

The `zk.ChildrenW` returns both the addresses and the watch channel at the same time. It's kinda waste to throw away the first-time result.

vieux:
LGTM thanks

PR #232, opened 2015-01-09
Proposal: add a `list` discovery service to give a list of ips

vieux:
Right now it feels weird that you can pass ips as argument to swarm manage, it's documented nowhere.
Also, when you start `swarm manage` without any argument it seems to hang.

I propose to change this to a new discovery system so

`swarm manage 1.1.1.1:1111 2.2.2.2:2222` will become `swarm manage --discovery list://1.1.1.1:1111,2.2.2.2:2222`

and `swarm manage` will output `FATA[0000] --discovery required to manage a cluster`


vieux:
ping @aluzzardi

discovery/README.md
```diff
@@ -124,6 +124,20 @@ $ swarm list --discovery zk://<zookeeper_addr1>,<zookeeper_addr2>/<path>
 http://<node_ip:2375>
 ```
 
+###### Using a static list of ips
+
+```bash
+# start the manager on any machine or your laptop
+$ swarm manage --discovery list://<node_ip1:2375>,<node_ip2:2375> -H=<swarm_ip:swarm_port>
```
tnachen: What's the motivation behind the list:// prefix? why not just node:1234,node2:2345 ?

vieux: I guess to be consistant with `file://`, `token://` etc but you're right we could to an exception

aluzzardi:
I like the concept, I just don't like the name `list`.

How about `nodes`, `static`, ... ?

vieux:
@aluzzardi what do you think about @tnachen proposal, using `--discovery node:1234,node2:2345`
it show it's a special discovery service.

vieux:
also, if we merge this `--discovery` will be required for every command. What about removing the flag ?

```console
$ swarm manage node:1234,node2:2345
#or
$ swarm manage nodes://node:1234,node2:2345

$ swarm manage token://48534832908742389074236
$ swarm join etcd://127.0.0.1/swarm
$ swarm list zk://127.0.0.1,127.0.0.2/swarm
....
```

aluzzardi:
I agree for removing `--discovery` since it's becoming useless now.

Not sure about `nodes://node1,node2` vs `node1,node2`, I think I'm okay for either way.

discovery/README.md
vieux: `nodes://` is now optional, thanks for the feedback @tnachen

vieux:
I've decided to leave `--discovery` in this PR mainly for the usage, if we remove it we need to rework the usage, otherwise it's becoming not clear.

Let's work on this in another PR

PR #233, opened 2015-01-11
Proposal: constraint & affinity filter expression enhancement

chanwit:
This PR proposes the use of `!=` for negate and `/regexp/` for direct regexp matching to improve the constraint filters. It supports `==` and relative comparisons `>=` and `<=` for range matching (string only).

For example,

*  `constraint:name==node1` will match `node1`.
*  `constraint:name!=node1` will match all nodes, expect `node1`.
*  `constraint:name!=n*` will match all, except nodes whose name starting with `n`.
*  `constraint:region!=us*` will match all, except nodes in `us` regions like `us-west` or `us-east`.
*  `constraint:name==/node\d/` will match all nodes named by `node` plus 1 digit.
*  `constraint:name==/node[12]/` will match nodes named by `node1` and `node2`.
*  `constraint:node!=/node-[01]-id/` will match all, except nodes id `node-0-id` and `node-1-id`.
*  `constraint:name!=/foo\[bar\]/` will match all nodes, except those with name `foo[bar]`. You can see the use of escape characters here.
*  `constraint:name!=/(?i)node/` will match all nodes name case-insensitivity, so `NoDe` and `NODE` are also matched.
* `constraint:group<=3` will match all nodes in group less than or equal to "3". This is string, not numeric, comparison.
* `constraint:kernel>=3.0` will match all nodes having label kernel greater than or equal to "3.0". This is also string, not numeric, comparison.

### DIFF
--- a/scheduler/filter/README.md
+++ b/scheduler/filter/README.md
@@ -150,15 +150,22 @@ As you can see here, the containers were only scheduled on nodes with the redis 
 #### Constraint Expression Syntax
 
 Additionally, you can use a not (`!`) to negate and a regular expression in the form of `/regexp/` for specifying a constraint.
+Relative comparisons, `>=` and `<=` are also supported, but they are limited to `string` comparison only.
+
 For example,
 
 * `constraint:name=node1` will match nodes named with `node1`.
-* `constraint:name=!node1` will match all nodes, except `node1`.
+* `constraint:name==node1` will also match nodes named with `node1`. Note that `==` also allowed.
+* `constraint:name!=node1` will match all nodes, except `node1`.
+* `constraint:name=!node1` will match all nodes, except `node1` (alternative syntax).
 * `constraint:region=!us*` will match all nodes outside the regions prefixed with `us`.
 * `constraint:name=/node[12]/` will match nodes named `node1` and `node2`.
 * `constraint:name=/node\d/` will match all nodes named with `node` + 1 digit.
 * `constraint:node=!/node-[01]-id/` will match all nodes, except those with ids `node-0-id` and `node-1-id`.
 * `constraint:name=!/foo\[bar\]/` will match all nodes, except those with name `foo[bar]`. You can see the use of escape characters here.
+* `constraint:name=/(?i)node1/` will match all nodes named with `node1` case-insensitive. So 'NoDe1' or 'NODE1' will also matched.
+* `constraint:kernel>=3.0` will match all nodes with label `kernel` greater than or equal to "3.0". This is the string, not numeric, comparison.
+* `constraint:group<=3` will match all nodes with `group`  less than or equal to "3". This is also the string, not numeric, comparison.
 
 ## Port Filter
 
--- a/scheduler/filter/README.md
+++ b/scheduler/filter/README.md
@@ -152,12 +152,13 @@ As you can see here, the containers were only scheduled on nodes with the redis 
 Additionally, you can use a not (`!`) to negate and a regular expression in the form of `/regexp/` for specifying a constraint.
 For example,
 
-* `name=node1` will match nodes named with `node1`.
-* `name=!node1` will match all nodes, except `node1`.
-* `region=!us*` will match all nodes outside the regions prefixed with `us`.
-* `name=/node[12]/` will match nodes named `node1` and `node2`.
-* `name=/node\d/` will match all nodes named with `node` + 1 digit
-* `node=!/node-[01]-id/` will match all nodes, except those with ids `node-0-id` and `node-1-id`
+* `constraint:name=node1` will match nodes named with `node1`.
+* `constraint:name=!node1` will match all nodes, except `node1`.
+* `constraint:region=!us*` will match all nodes outside the regions prefixed with `us`.
+* `constraint:name=/node[12]/` will match nodes named `node1` and `node2`.
+* `constraint:name=/node\d/` will match all nodes named with `node` + 1 digit.
+* `constraint:node=!/node-[01]-id/` will match all nodes, except those with ids `node-0-id` and `node-1-id`.
+* `constraint:name=!/foo\[bar\]/` will match all nodes, except those with name `foo[bar]`. You can see the use of escape characters here.
 
 ## Port Filter
 
--- a/scheduler/filter/README.md
+++ b/scheduler/filter/README.md
@@ -45,7 +45,7 @@ Once the nodes are registered with the cluster, the master pulls their respectiv
 Let's start a MySQL server and make sure it gets good I/O performance by selecting nodes with flash drives:
 
 ```
-$ docker run -d -P -e constraint:storage=ssd --name db mysql
+$ docker run -d -P -e constraint:storage==ssd --name db mysql
 f8b693db9cd6
 
 $ docker ps
@@ -59,7 +59,7 @@ In this case, the master selected all nodes that met the `storage=ssd` constrain
 Now we want to run an `nginx` frontend in our cluster. However, we don't want *flash* drives since we'll mostly write logs to disk.
 
 ```
-$ docker run -d -P -e constraint:storage=disk --name frontend nginx
+$ docker run -d -P -e constraint:storage==disk --name frontend nginx
 f8b693db9cd6
 
 $ docker ps
@@ -161,16 +161,14 @@ You can use a not (`!`) to negate and a regular expression in the form of `/rege
 Relative comparisons, `>=` and `<=` are also supported, but they are limited to `string` comparison only.
 
 For example,
-* `constraint:name=node1` will match nodes named with `node1`.
-* `constraint:name==node1` will also match nodes named with `node1`. Note that `==` also allowed.
+* `constraint:name==node1` will match nodes named with `node1`.
 * `constraint:name!=node1` will match all nodes, except `node1`.
-* `constraint:name=!node1` will match all nodes, except `node1` (alternative syntax).
-* `constraint:region=!us*` will match all nodes outside the regions prefixed with `us`.
-* `constraint:name=/node[12]/` will match nodes named `node1` and `node2`.
-* `constraint:name=/node\d/` will match all nodes named with `node` + 1 digit.
-* `constraint:node=!/node-[01]-id/` will match all nodes, except those with ids `node-0-id` and `node-1-id`.
-* `constraint:name=!/foo\[bar\]/` will match all nodes, except those with name `foo[bar]`. You can see the use of escape characters here.
-* `constraint:name=/(?i)node1/` will match all nodes named with `node1` case-insensitive. So 'NoDe1' or 'NODE1' will also matched.
+* `constraint:region!=us*` will match all nodes outside the regions prefixed with `us`.
+* `constraint:name==/node[12]/` will match nodes named `node1` and `node2`.
+* `constraint:name==/node\d/` will match all nodes named with `node` + 1 digit.
+* `constraint:node!=/node-[01]-id/` will match all nodes, except those with ids `node-0-id` and `node-1-id`.
+* `constraint:name!=/foo\[bar\]/` will match all nodes, except those with name `foo[bar]`. You can see the use of escape characters here.
+* `constraint:name==/(?i)node1/` will match all nodes named with `node1` case-insensitive. So 'NoDe1' or 'NODE1' will also matched.
 * `constraint:kernel>=3.0` will match all nodes with label `kernel` greater than or equal to "3.0". This is the string, not numeric, comparison.
 * `constraint:group<=3` will match all nodes with `group`  less than or equal to "3". This is also the string, not numeric, comparison.
 
--- a/scheduler/filter/README.md
+++ b/scheduler/filter/README.md
@@ -95,11 +95,11 @@ CONTAINER ID        IMAGE               COMMAND             CREATED             
 87c4376856a8        nginx:latest        "nginx"             Less than a second ago   running             192.168.0.42:80->80/tcp         node-1      front
 ```
 
-Using `-e affinity:container=front` will schedule a container next to the container `front`.
-You can also use IDs instead of name: `-e affinity:container=87c4376856a8`
+Using `-e affinity:container==front` will schedule a container next to the container `front`.
+You can also use IDs instead of name: `-e affinity:container==87c4376856a8`
 
 ```
-$ docker run -d --name logger -e affinity:container=front logger
+$ docker run -d --name logger -e affinity:container==front logger
  87c4376856a8
 
 $ docker ps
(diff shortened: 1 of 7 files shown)


aluzzardi:
I like it, except we might want to use operators (== and !=) instead of negation (=, =!).

@vieux?

vieux:
@aluzzardi I'm still undecided, for we we could even support both (`=` and `==` same, `!=` and `=!` same)

I'll let you make the call.

thaJeztah:
Is there a way to distinguish between using regexp and a regular compare? possibly edge-case, but what if I named something `foo[bar]`. Not a good example, but just to illustrate

vieux:
I guess right now when it's starts and ends with a `/` it's a regex, otherwise it's simple globbing

chanwit:
@aluzzardi @vieux when it comes to environment variables, I feel right with a single `=`
and `!` in this syntax will be used to negate the expression behind it.

but I leave the decision to you guys anyway.

chanwit:
@thaJeztah, @vieux is right. Everything between two `/`s will be a regexp. This means you can also escape whatever characters you'd like there.

thaJeztah:
Okay, sounds good to me. Wasn't sure other cases were treated as well (and could become troublesome).

Thanks for the info

chanwit:
@thaJeztah I've added an extra test to cover your case and also updated the README.

thaJeztah:
Great!

thaJeztah:
A possible alternative could be to use a special operator for regexes; iirc, PostgreSQL uses `~` for this. That would also allow other character than `/` to start/end the Regex, which could be helpful matching against strings with many slashes.

This is just a suggestion I was thinking of when reading this PR originally. The current implementation should work for most cases, but wanted to suggest it just in case.

chanwit:
@thaJeztah  So it would be `constraint:name=~foo\[bar]\`, for example ?

thaJeztah:
Typing on my phone now, not sure if they also use `/...../`. In which case it would be `=~/^regex-here$/` not sure.

chanwit:
For `~` syntax, I ping @aluzzardi and @vieux for addition comments.

thaJeztah:
Looked up the syntax that's used there; See [documentation](http://www.postgresql.org/docs/current/static/functions-matching.html#FUNCTIONS-POSIX-TABLE)

The reason I'm adding all this "noise" is that I'm looking at a wider scope; not sure if that is the right thing to do, but I'm thinking if Docker *itself* would start using regex-matching for (for example) labels, it would be a good thing to have a syntax that can be re-used there.

chanwit:
@thaJeztah I am more than happy to help implement this. But I am not the ones who can decide which feature to be put into swarm, anyway :) Let us see the maintainers' opinions.

vieux:
@thaJeztah what's bothering you about `/......./` ?

thaJeztah:
@vieux no *big* concerns, apart from possibly allowing a bit more flexibility (https://github.com/docker/swarm/pull/233#issuecomment-69705637); using `~` wouldn't restrict the regex to use a hard-coded `/` for quoting, which could make it easier to write regexes *containing* slashes.

Also, the option to have case-insensitive matching as illustrated in the [PostgreSQL docs](http://www.postgresql.org/docs/current/static/functions-matching.html#FUNCTIONS-POSIX-TABLE) could be a nice addition.

So, to summarise; I have no *big* objections with the PR as it is, I wanted to add some alternative options to consider, before it being "set in stone".

chanwit:
To make this PR compact and actionable, I'd to summary the points we've discussed. So it's going to be easier for us to conclude things.

1. Do we agree with the proposed syntax `constraint:name=!/regexp/` ? (I do, of course)
2. Do we agree with the alternative syntax `constraint:name!=/regexp/` ? (I am neutral and I'm OK to implement it)
3. Do we agree with the leading `~` syntax, so we can add an extra option like insensitivity to an expression, like` constraint:name=!~*/regexp/`? (I am neutral and I'm OK to implement it)

@aluzzardi @vieux @thaJeztah Could you please response to this one-by-one. Thank you.

aluzzardi:
@chanwit: My opinion (but feel free to discard):
1/ I like `/pattern/` better than `=~ pattern` (I don't like using `~`)
2/ Regarding != vs =!, I'm mostly neutral. However, I feel like `!=` is more powerful. Someday we might have `kernel>=3.0` for instance.

Another thing: Right now we compose the regexps in the code (we add `^`, `$`, etc to do globalling). Perhaps we should simply use the regular expression directly.

chanwit:
@aluzzardi thank you for comments. I'm convinced with `kernel>=3.0`. It has a real value!

vieux:
Pretty much the same as @aluzzardi  for me, 

I don't like `~` and `!=` seems better as he said.

Regarding the @aluzzardi's another thing: if I'm not mistaking, we are only adding `^` and `$` for the matching, and I would like to keep it. Regexp are very powerful, but I feel most of the users will use simple globbing like `-e constraint:region=us*`

My another thing: as you said in #243, as your PR will likely be merged before, can you refactor and create the `filter/utlils.go` right away ?

Thanks!

chanwit:
I have this PR supports `!=`, `==` also `>=` and `<=`. Matching codes are refactored into `filter/utils.go`.

scheduler/filter/README.md
```diff
@@ -79,6 +79,26 @@ Those tags are sourced from `docker info` and currently include:
...
 * operatingsystem
 
+#### Constraint Expression Syntax
+
+Additionally, you can use a not (`!`) to negate and a regular expression in the form of `/regexp/` for specifying a constraint.
+Relative comparisons, `>=` and `<=` are also supported, but they are limited to `string` comparison only.
+
+For example,
+
+* `constraint:name=node1` will match nodes named with `node1`.
```
tnachen: Do we have restrictions what characters we allow labels to have? we want to also make sure labels are not allowed to use any of the operator we add

chanwit: I'm not sure but think that restriction is implemented in docker?  maybe @vieux can answer this.

scheduler/filter/utils.go
```diff
@@ -0,0 +1,89 @@
...
+	gte
+	lte
+)
+
+func parse(k, v string) (string, string, comparison, bool) {
+	// default comparison mode
+	mode := equ
+
+	// support case of constraint:k==v
+	// try to check = on both sides, "k= : = : v" and "k : = : =v", to make sure it works everywhere
```
tnachen: I'm not sure I understand the cases you have here?

chanwit: Using "=", it's possible for an entry "k==v" to be split as:
1. "k=" as key and "v" as value
2. "k" as key and "=v" as value
I just make sure to cover these cases.

scheduler/filter/utils.go
```diff
@@ -0,0 +1,89 @@
...
+	if strings.HasSuffix(k, "=") {
+		k = strings.TrimSuffix(k, "=")
+	} else if strings.HasPrefix(v, "=") {
+		v = strings.TrimPrefix(v, "=")
+	}
+
+	if strings.HasPrefix(v, "!") {
```
tnachen: I personally think this will become harder to maintain when we add more operators and more cases to handle.
How about instead of splitting the constraint on "=" and then trying to find each operator in either key/val, you define all the operators (ie: "=", "==", ">=", "<=", "!=", "=!", and each operator has a matching mode.
Then all you have to do something like the following:

func findMode(constraint string) {
for(op <- operators) {
 if strings.Contains(constraints, op.value) {
    return op.mode
 }
}

 err....
}

chanwit: I see. Thank you for your comments. I'll refactor this next time I add another operator.
Which operator do you think we should add more anyway?

chanwit:
make comments more clear for double-equal parsing as suggested by @tnachen

scheduler/filter/README.md
vieux: maybe we should restrict keys and values to alphanum only ? 

@aluzzardi ?

chanwit:
Feature complete for now.
ping @vieux @aluzzardi

scheduler/filter/utils.go
```diff
@@ -0,0 +1,92 @@
...
+	return k, v, mode, useRegex
+}
+
+func extractEnv(key string, env []string) map[string]string {
+	values := make(map[string]string)
+	for _, e := range env {
+		if strings.HasPrefix(e, key+":") {
+			value := strings.TrimPrefix(e, key+":")
+			parts := strings.SplitN(value, "=", 2)
+			if len(parts) == 2 {
```
thaJeztah: Should this be `>= 2` if values themselves contain `=`? The value should then be the remainder parts concat-ed with `=`.

scheduler/filter/utils.go
```diff
@@ -0,0 +1,92 @@
...
+				values[strings.ToLower(parts[0])] = strings.ToLower(parts[1])
```
thaJeztah: Not sure if `ToLower` (at least, on the *value*) is the right thing to do here. I'd keep the value as-is and decide if matches should be case-insensitive during the actual filtering.

Perhaps white-space  *should* be trimmed from value and key to allow `something == something`?

scheduler/filter/constraint_test.go
```diff
@@ -108,3 +112,279 @@ func TestConstrainteFilter(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Len(t, result, 2)
 }
+
+func TestConstraintNotExpr(t *testing.T) {
+	var (
```
thaJeztah: Do we need a positive/negative test-case using spaces around the operator? `foo == bar`?

thaJeztah: ^^ that's a general note! just picked a test to out a comment on :smile:

scheduler/filter/utils.go
chanwit: I think 2 is fine here. It's the code I've copied from @vieux. The value containing `=` will not raise any problem. We just detect the first `=` as the delimiter.

scheduler/filter/utils.go
chanwit: > Not sure if ToLower (at least, on the value) is the right thing to do here.
May be you're right. I'll double check this. Thanks !

> Perhaps white-space should be trimmed from value and key to allow something == something?
I think this is not the case, as each entry must go through the environment variable map. So it's impossible to put some spaces there. @vieux  what do you think?

scheduler/filter/README.md
chanwit: +1

aluzzardi: Agreed.

chanwit: should we put the restriction in this PR? Or you'd prefer a separate one?

chanwit: It seems the validation can be placed nicely in the refactoring `filter/utils.go`. I'll go for it.

scheduler/filter/utils.go
tnachen: Not sure, but I can adding wild cards is going to be a common ask
key==*abc*

chanwit: Your case is already covered by the syntax. I'll add a test case to confirm it.

scheduler/filter/constraint_test.go
chanwit: it's fixed properly in the new commit. thanks!

chanwit:
I think it is good for now. @vieux could you please review?

aanand:
I'm concerned that these constraints are **implicitly** on host labels - further down the line when containers have labels too, that's going to be confusing (does `name` mean "host name" or "container name")?

chanwit:
@aanand `name` is a container name.

aanand:
@chanwit I'm confused then - your examples in the description at the top talk about names belonging to "nodes", and seems to strongly imply that nodes are hosts (e.g. they exist in regions, they have a kernel version).

chanwit:
@aanand my bad, it's a docker instance, not a container.
They are a set labels exposed by a docker instance.
Sorry for any confusion.

aanand:
No worries. So here's how I understand it:

- With this PR, you can schedule a container somewhere in a constrained set of hosts using `-e constraint:name=foo`, `-e constraint:region=us-east`, `-e constraint:kernel>=3.0`, etc
- With #243, you can schedule a container somewhere based on other containers using `-e affinity:container=backend`, `-e affinity:container!=redis-master`, etc

So I have two questions:

- What will the syntax look like for container-label-based affinity? (This should arguably be discussed in #243)
- How do we make sure that it's obvious that `constraint` uses host labels and `affinity` uses container labels?

I don't think `constraint:name` is sufficiently self-describing. Perhaps `constraint` should just be `node`? `-e node:name=node1`, `-e node:region=us-east`, etc.

In any case, I'm anxious that the syntax for node constraints and container affinity - when dealing with names, labels or whatever else - is designed as a whole, not as if they were two unrelated features.

chanwit:
@aanand I think labels (any string after `:`) for `constraint` and `affinity` are not directly related. They're just going to share the same parser from `filter/utils.go`.

For the reason why it's called `constraint`, @vieux may be the right person to answer you. I am not the one who inventing it. This PR is just to enhance @vieux's early work.

chanwit:
@aanand anyway I feel to agree with you for `constraint:name` to be changed to `node:name` and may be `constraint:node` would also be change to `node:id`, which better reflects their meaning.

@vieux any though?

vieux:
@aanand as of right now, affinity won't use container labels are they are no such thing, we are using container and image names/ID 

I my opinion `constraint` of `affinity` don't relate to what information we use (`host labels, container names`) but more about what it does:

`affinity` -> I want to be next to a container or next to an image
`constraint` -> I want to be in any host that match this

@aluzzardi any thought ?

vieux:
@chanwit after chatting with @aluzzardi we believe it would be much easier if we only supports:

`key==value` and `key!=value` so we can split by `==` and `!=` in the code.

We are not supporting `=` or `=!`

vieux:
Can you also please rebase ? thanks :D

chanwit:
RAAC (rebased and the adventure continues)

chanwit:
> @chanwit after chatting with @aluzzardi we believe it would be much easier if we only supports:
> key==value and key!=value so we can split by == and != in the code.
> We are not supporting = or =!

No problem. I'll ping you guys after it's done.

chanwit:
@vieux code has been largely refactored. it's now easily to add/remove operators if needed (thanks to @tnachen suggestions)

please have a review.

thaJeztah:
Yay, merged! Thanks for your contribution, @chanwit!

chanwit:
it's not as good as it should be, but thank you all for participation!

vieux:
@chanwit are you talking about `>=` and `<=` or something else ? Sorry to hear that.

chanwit:
@vieux  I meant to my code qualities compared to your last commit for this PR. Your code looks far better :)

For `>=` and `<=`, it's good to remove unused things before getting released.
I understand that, do not worry!!

PR #234, opened 2015-01-12
Add a simple key<->Container persistent store.

aluzzardi:

### DIFF
--- a/state/store.go
+++ b/state/store.go
@@ -1,4 +1,4 @@
-package cluster
+package state
 
 import (
 	"encoding/json"
@@ -9,8 +9,6 @@ import (
 	"path"
 	"path/filepath"
 	"sync"
-
-	"github.com/samalba/dockerclient"
 )
 
 var (
@@ -24,11 +22,6 @@ type Store struct {
 	values  map[string]*RequestedState
 
 	sync.RWMutex
-}
-
-type RequestedState struct {
-	Name   string
-	Config *dockerclient.ContainerConfig
 }
 
 func NewStore(rootdir string) *Store {
--- a/state/store.go
+++ b/state/store.go
@@ -9,11 +9,14 @@ import (
 	"path"
 	"path/filepath"
 	"sync"
+
+	log "github.com/Sirupsen/logrus"
 )
 
 var (
 	ErrNotFound      = errors.New("not found")
 	ErrAlreadyExists = errors.New("already exists")
+	ErrInvalidKey    = errors.New("invalid key")
 )
 
 // A simple key<->RequestedState store.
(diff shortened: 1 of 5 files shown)


state/store.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	if err := s.restore(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (s *Store) path(key string) string {
```
tnachen: I haven't seen where this is used yet, so ignore me if this is unrelevant.
Should the key be santizied (lowercased, trim, etc?)

aluzzardi:
Yeah, sorry about that.

It will be used by this PR: https://github.com/docker/swarm/pull/236

state/store.go
vieux: Key will be IDs so there is no need to cleanup. @aluzzardi can you confirm ?

aluzzardi: Indeed !

tnachen: Thanks for clarifying!

state/store.go
```diff
@@ -0,0 +1,177 @@
...
+
+		// Load the object back.
+		value, err := s.load(path.Join(s.RootDir, file))
+		if err != nil {
+			return err
+		}
+
+		// Extract the key.
+		key := file[0 : len(file)-len(extension)]
+		if len(key) == 0 {
```
tnachen: I know this is a very specific store for swarm, so trying to focus on swarm specific use cases.
But I think if we're going to be checking for the key length here, there is nothing stopping the Add() to pass a empty key. If somehow that happened, on restore we simply can't initialize anymore. I think it's worth fixing on the Add side.

state/store.go
```diff
@@ -0,0 +1,177 @@
...
+		file := fileinfo.Name()
+
+		// Verify the file extension.
+		extension := filepath.Ext(file)
+		if extension != ".json" {
+			return fmt.Errorf("invalid file extension for filename %s (%s)", file, extension)
+		}
+
+		// Load the object back.
+		value, err := s.load(path.Join(s.RootDir, file))
```
tnachen: Should we stop the restore if we can't load a single file?
I wonder if there are situations a crash before writing the whole file can cause half writes to the disk, and then cause us never able to restore. 
I would consider log error but still able to use the cache.

state/store.go
aluzzardi: Done. PTAL

tnachen:
LGTM :)

aluzzardi:
@vieux ?

vieux:
LGTM

PR #235, opened 2015-01-12
Container: Make Node public instead of exposing it through Node().

aluzzardi:


vieux:
LGTM

PR #236, opened 2015-01-12
Store requested state persistently.

aluzzardi:
Depends on https://github.com/docker/swarm/pull/234 and https://github.com/docker/swarm/pull/235

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -8,6 +8,8 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/swarm/discovery"
+	"github.com/docker/swarm/state"
+	"github.com/samalba/dockerclient"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


vieux:
@aluzzardi I rebase for you :D

flags.go
```diff
@@ -3,6 +3,11 @@ package main
 import "github.com/codegangsta/cli"
 
 var (
+	flStore = cli.StringFlag{
+		Name:  "store",
+		Value: "/var/lib/docker/swarm/store",
```
vieux: I think we should change the so you don't have to `sudo` or change `--store` every time.

`$HOME/.swarm/store` ?

cluster/cluster.go
```diff
@@ -17,20 +19,54 @@ var (
...
 
+// Deploy a container into a `specific` node on the cluster.
+func (c *Cluster) DeployContainer(node *Node, config *dockerclient.ContainerConfig, name string) (*Container, error) {
+	container, err := node.Create(config, name, true)
+	if err != nil {
+		return nil, err
+	}
+
+	// Commit the requested state.
+	st := &state.RequestedState{
```
tnachen: Btw, looking at this, I wonder if go is flexible enough to serialize fields that is missing when it reads json from disk?
The reason I'm asking is since we're serializing into json from a go struct, this RequestedState obj might change in schema and we need the store to handle this.

flags.go
aluzzardi: Not sure - I agree that it's bad to `sudo` just for that.

On the other hand, it's risky to use `$HOME` since changing user would mean losing the state (or having part of the state in `vieux` and other parts in `root`).

Thoughts?

vieux: Currently, the store is a nice to have, but isn't really mandatory: `/tmp` ?

cluster/cluster.go
vieux: If we add fields in `RequestedState` and load an old version, there will be no error, the new fields will be empty.

If we remove fields in `RequestedState` and load an old version, there will be no error either.

The only error could be if we change a string to a float or something like this.

Does this answer your question ?

tnachen: Yes it does, just wasn't sure how strict Go seriailizations are. Thanks!

Issue #149, opened 2014-12-05
More optimal host selection strategy


ykumar6:
How easy or hard would it be to add a strategy where

1) Selection favors data locality on host (image cache)
2) Selection favors least weighted host (each task .. building vs running, assigns different weights to a host)

Here is something we wrote internally.
https://github.com/CodeNow/mavis

We don't have any Go programmers but will try to contribute to this

aluzzardi:
Hey @ykumar6 

We addressed many of this issues since then in our new scheduler (balanced strategies, better filters, etc).

I guess this issue is very generic and it's going to be a constant focus for us to improve the scheduler.

If you want to suggest improvements, please open issues for specific areas.

Thanks!

PR #243, opened 2015-01-14
add affinity (start next to a container/image)

vieux:
There is no documentation and a couple TODOs, I'm simply opening the PR to start the discussion.

It adds the `affinity` prefix, with 2 available keys.

To start your frontend container next to your backend (why not? :D)
```
docker run -d --name backend php
docker run -d -e affinity:container=backend --name frontend nginx
```
To start your redis container in a node where it image is already pulled
```
docker run -d -e affinity:images=redis redis
```
Combining this PR with @chanwit #233 it would allow you to do
```
docker run -d --name redis-master redid
docker run -d affinity:container=!redis-master --name slave redis
```

### DIFF
--- a/cluster/node.go
+++ b/cluster/node.go
@@ -46,6 +46,7 @@ type Node struct {
 
 	ch              chan bool
 	containers      map[string]*Container
+	images          []*dockerclient.Image
 	client          dockerclient.Client
 	eventHandler    EventHandler
 	healthy         bool
@@ -81,6 +82,10 @@ func (n *Node) connectClient(client dockerclient.Client) error {
 
 	// Force a state update before returning.
 	if err := n.refreshContainers(); err != nil {
+		n.client = nil
+		return err
+	}
+	if err := n.refreshImages(); err != nil {
 		n.client = nil
 		return err
 	}
@@ -128,6 +133,18 @@ func (n *Node) updateSpecs() error {
 		kv := strings.SplitN(label, "=", 2)
 		n.Labels[kv[0]] = kv[1]
 	}
+	return nil
+}
+
+// Refresh the list of images on the node.
+func (n *Node) refreshImages() error {
+	images, err := n.client.ListImages()
+	if err != nil {
+		return err
+	}
+	n.Lock()
+	n.images = images
+	n.Unlock()
 	return nil
 }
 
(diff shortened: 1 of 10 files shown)


jfrazelle:
oooo this is cool

abronan:
I approve :+1: Should we close #149 in favor of this PR?

chanwit:
This is great!
Should I refactor #233's matching into `filter/utils.go`? So you can also use it in this PR.

vieux:
@abronan it's related, but I'm not sure it's completely fixing it.

aluzzardi:
:+1:

vieux:
I added doc, tests and image cache.

I decided to use a `[]*Image` instead of a `map[string]*image` because, unlike containers, we don't need to access one image via it's ID.
There is no point to complicate the code.

@aluzzardi PTAL

aluzzardi:
In the event handler (`handler`), we might need to refresh the images as well when something is pulled.

Not sure if `ev.Id` refers to the `container` or the image in that case.

cluster/node.go
```diff
@@ -374,6 +378,12 @@ func (n *Node) Containers() []*Container {
 	return containers
 }
 
+func (n *Node) Images() []*dockerclient.Image {
+	n.RLock()
```
aluzzardi: The lock is useless since we are returning a reference to the list.

How about copying the values into a new list (as we do for `Containers()`)?

scheduler/filter/README.md
```diff
@@ -79,6 +80,73 @@ Those tags are sourced from `docker info` and currently include:
...
+```
+$ docker run -d -p 80:80 --name front nginx
+ 87c4376856a8
+
+$ docker ps
+CONTAINER ID        IMAGE               COMMAND             CREATED                  STATUS              PORTS                           NODE        NAMES
+87c4376856a8        nginx:latest        "nginx"             Less than a second ago   running             192.168.0.42:80->80/tcp         node-1      front
+```
+
+Using `-e affinity:container=front` will schedule a container next to the container `front`.
```
aluzzardi: Note for later: We should update the example using regular expressions as it makes more sense for affinity.

e.g.

`--name front_1`
`--name front_2`
`-e affinity:container=front_*`

cluster/node.go
vieux: done

scheduler/filter/README.md
vieux: :+1:

vieux:
@aluzzardi et `ev.Id` is the image name but there is no way to refresh only one image, so I trigger a full refresh

vieux:
@aluzzardi PTAL I think it's good to go

aluzzardi:
LGTM

PR #246, opened 2015-01-14
add search and login

vieux:
`docker search` and `docker login` use the docker daemon as a proxy, they aren't doing anything meaningful on the daemon
(they could be implemented client side)

So, for those we simply pick a random node and proxy the request.


aluzzardi:
Maybe we should pick a random node within the healthy subset?

vieux:
right

PR #250, opened 2015-01-16
allow hostnames in join

### LINKED ISSUE #248: Allowing `addr` to be hostname
What would be the blockers (if any) that need to be resolved to allow `addr` to be a hostname (in `swarm join`)? 

When using SSL, forcing the advertised address (which I understand `swarm manage` will collect to) to be a hostname makes things a little bit more complicated than they should (i.e. you need a certificate with an IP SAN, and generating one with OpenSSL isn't exactly the most intuitive thing either).

Or maybe i'm missing something entirely?!

Happy to help if possible,

Thanks!

vieux:
Closes #248

### DIFF
--- a/join.go
+++ b/join.go
@@ -9,6 +9,11 @@ import (
 	"github.com/docker/swarm/discovery"
 )
 
+func checkAddrFormat(addr string) bool {
+	m, _ := regexp.MatchString("^[0-9a-zA-Z._-]+:[0-9]{1,5}$", addr)
+	return m
+}
+
 func join(c *cli.Context) {
 
 	if c.String("discovery") == "" {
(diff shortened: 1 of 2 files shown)


join.go
```diff
@@ -21,7 +26,8 @@ func join(c *cli.Context) {
 	}
 
 	addr := c.String("addr")
-	if m, _ := regexp.MatchString("^[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}.[0-9]{1,3}:[0-9]{1,5}$", addr); !m {
+
+	if !checkAddrFormat(addr) {
 		log.Fatal("--addr should be of the form ip:port")
```
tnachen: Seems like this should be "ip:port or hostname:port"

vieux: absolutly

vieux: updated

krallin:
When I register one my hosts using a hostname instead of an IP, I get the following error in the `swarm manage` logs:

`ERRO[32026] Get my.host.goes.here:2376/v1.15/info: unsupported protocol scheme "my.host.goes.here" `

Passing `tcp://my.host.goes.here:2376` as the `--addr` throws an error, of course. 

It seems `swarm manage` might be doing its own parsing somewhere and isn't able to accept a hostname?

aluzzardi:
LGTM

vieux:
@krallin you need the last version of dockerclient. can you please do a `go get -u ./...` before recompiling

krallin:
Ooh, that makes sense. Thanks, @vieux !

PR #252, opened 2015-01-16
Proposal: add `swarm config`

vieux:
I believe that when we will support `docker pull`, we are going to pull on every machines.
When looking at the current docker cli, I don't see any way for us to specify a node.

This PR adds a `swarm config` commands that returns you the `<ip:port>` of a node using it's name/ID.

It can be used like this:

```
docker `swarm  config token://XXXX ubuntu-2` pull redis
docker `swarm  config token://XXXX ubuntu-1` rmi nginx
```

I called the command `config` to be consistant with docker machine and
also because I don't know how to call `<ip:port>` (we can't call  the command `swarm ip`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,9 @@ $ docker -H tcp://<swarm_ip:swarm_port> ps
 $ docker -H tcp://<swarm_ip:swarm_port> logs ...
 ...
 
+# connect to a node directly
+$ docker `swarm config token://<cluster_id> <node_name>` pull ...
+
 # list nodes in your cluster
 $ swarm list token://<cluster_id>
 <node_ip:2375>
--- a/config.go
+++ b/config.go
@@ -1,0 +1,69 @@
+package main
+
+import (
+	"crypto/tls"
+	"fmt"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
+	"github.com/docker/swarm/cluster"
+	"github.com/docker/swarm/discovery"
+)
+
+func config(c *cli.Context) {
+	var (
+		tlsConfig *tls.Config = nil
+		err       error
+	)
+
+	// If either --tls or --tlsverify are specified, load the certificates.
+	if c.Bool("tls") || c.Bool("tlsverify") {
+		tlsConfig, err = loadTlsConfig(
+			c.String("tlscacert"),
+			c.String("tlscert"),
+			c.String("tlskey"),
+			c.Bool("tlsverify"))
+		if err != nil {
+			log.Fatal(err)
+		}
+	}
+
+	if !c.IsSet("discovery") {
+		log.Fatal("--discovery required to get config of a node")
+	}
+
+	if len(c.Args()) != 1 {
+		log.Fatal("an argument is required to get config of a node")
+	}
+	// get the list of nodes from the discovery service
+	d, err := discovery.New(c.String("discovery"), 0)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	nodes, err := d.Fetch()
+	if err != nil {
+		log.Fatal(err)
+
+	}
+
+	connectedNodes := make(chan *cluster.Node, len(nodes))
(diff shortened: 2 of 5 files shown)


vieux:
@ehazlett what do you think ?

ehazlett:
@vieux i like it (i'm not tied to the name `config` either if we wanted to use something different).  i think we should stay consistent though -- for example, in machine, we have `config` output all args not just the IP --  `docker $(machine config test-machine) ps` actually emits something like `docker --tls --tlscacert=/Users/user/.docker/machines/test-machine/ca.pem --tlscert=/Users/users/.docker/machines/test-machine/client.pem --tlskey=/Users/user/.docker/machines/test-machine/client-key.pem -H tcp://192.168.99.110:2376 ps`.  This is because currently we force TLS.  There is also https://github.com/docker/docker/pull/10151 that would allow a config file to be used -- swarm could use this as well.

vieux:
@ehazlett right, I'll add all those flags as well, good idea.

vieux:
@ehazlett after thinking about it more I'm not sure there is any advantage to add those flag:

```
# without
docker -H `swarm config --discovery token://XXXX ubuntu-2` --tls --tls-cert .... pull redis

#with
docker `swarm config --tls --tls-cert .... --discovery token://XXXX ubuntu-2` pull redis
```

ehazlett:
@vieux do you specify the `--tls ...` to swarm?  In machine we already have most of the metadata so we just use `machine config <name>`.  I was thinking something like `swarm config ubuntu-2` but you probably don't have the token so it might be different in your case.

vieux:
Yes you specify `--tls ...` to swarm. 
It's the exact same flags as docker.

ehazlett:
Ah ok. In machine it's the same but we already have the certs and keys so we can pre-populate the args.  I don't think it will work the same in swarm then.

vieux:
@ehazlett in any case, you are right, in my comment, from two days ago, the `without` section cannot work, because swarm needs the `--tls` flags to be able to connect to the node and get it's name.

vieux:
@ehazlett updated

ehazlett:
LGTM.  Machine could also help out with the TLS settings :)

vieux:
updated `README.md` and `discovery`

PTAL

ehazlett:
LGTM

aluzzardi:
I'm a little uncertain about this: since we don't provide tls configuration, the usefulness of `config` is limited as it can be achieved with a simple `swarm list | grep | cut`.

Thoughts?

vieux:
@aluzzardi not with `swarm list` and you don't get the `name` or `ID` in `swarm list`

Although, you're right,if we only care about `name`, we can use `docker info`

`docker info | grep <name> | cut -d : -f 2`

As long as we stay consistant in the docker info output, Im OK

ehazlett:
@aluzzardi Since Machine has more info (we have access to the tls config) it may not make sense for you guys to have the same.  idk :)

vieux:
Ok, I'm ok to close this one, @aluzzardi ?

aluzzardi:
Oh, I'm not sure if we should close it, I was just raising a point. It definitely is simpler to use then grepping around docker info and hoping the format won't change.

I was simply asking: Is it worth it?

vieux:
@aluzzardi what if `swarm config` was using the docker info API ? So we don't rely on the cli ?

```
docker `swarm config <name>` pull ubuntu
docker `swarm config -tls ... <name> pull ubuntu
```

Useless ?

vieux:
@aluzzardi what do you think of my last comment ?

aluzzardi:
In the end, I think that since we still need to pass a bunch of information to config (compared to machine), it might not be that useful.

Additionally, machine needs something handy to communicate with the nodes created, however with swarm you probably don't need to.

Regarding the pull example: we have to figure out to do it in a swarm-ish way.

`docker pull ubuntu@node-1` ?

vieux:
alright

Issue #253, opened 2015-01-19
Bugs in binpacking strategy


jhamrick:
Is the bin packing strategy supposed to choose the machine that has the most resources free, or the fewest resources free? Currently, it appears that it chooses the machine with the *fewest* free resources, I think due to the implementation of `Less`, which actually tests for greater: https://github.com/docker/swarm/blob/master/scheduler/strategy/binpacking.go#L81

If this is intentional, it might be good to add a comment clarifying that, or alternately just select the last element of the sorted list instead of the first (or use `sort.Reverse`).

Additionally, it seems that the `total` score is always zero. I think this is because dividing by 200 using integer division will always truncate to zero: https://github.com/docker/swarm/blob/master/scheduler/strategy/binpacking.go#L44 This line is equivalent to just dividing by 2; changing the code to do that seems to fix the problem.

rowoot:
> Is the bin packing strategy supposed to choose the machine that has the most resources free, or the fewest resources free? 

In an ideal world, the bin packing scheduler should be capable of packing the appropriate processes that result in maximum utilization of the node's resources (implicitly reducing wastage of resources).

> Currently, it appears that it chooses the machine with the fewest free resources, I think due to the implementation of Less, which actually tests for greater: https://github.com/docker/swarm/blob/master/scheduler/strategy/binpacking.go#L81

I think it is otherwise i.e it chooses machines with most free resources.  But, I agree that the working of the `Sort()` method here is a bit counter intuitive given that the expectation for `Sort()` is elements ordered by small to big but here it the opposite.

> Additionally, it seems that the total score is always zero. I think this is because dividing by 200 using integer division will always truncate to zero: https://github.com/docker/swarm/blob/master/scheduler/strategy/binpacking.go#L44 This line is equivalent to just dividing by 2; changing the code to do that seems to fix the problem.

I agree dividing by 2 makes more sense. That said dividing by 200 does not truncate to zero -- given that the score is multiplied by 100 first, so the value (unless 1) will always be greater than 200.

jhamrick:
> I think it is otherwise i.e it chooses machines with most free resources.

Well, at least from testing this out and adding in debug statements to verify, it currently chooses the machine with the fewest free resources (i.e. the one with the most containers already allocated to it).

> That said dividing by 200 does not truncate to zero -- given that the score is multiplied by 100 first, so the value (unless 1) will always be greater than 200.

[...]

PR #258, opened 2015-01-20
Add test that fails to execute the proper binpacking strategy

jhamrick:
This creates two nodes, each with 4GB of memory, and allocates first a 2GB container, then a 3GB container, then a 1GB container. According to the binpacking strategy, the 3GB and 1GB should end up on the same node, but they do not, due to the issue in #253.

### DIFF
--- a/scheduler/strategy/binpacking_test.go
+++ b/scheduler/strategy/binpacking_test.go
@@ -248,7 +248,7 @@ func TestComplexPlacement(t *testing.T) {
 	assert.NoError(t, node2.AddContainer(createContainer("c2", config)))
 
 	// check that they end up on separate nodes
-	assert.NotEqual(t, node1.ID, node2.ID, "")
+	assert.NotEqual(t, node1.ID, node2.ID)
 
 	// add one container 1G
 	config = createConfig(1, 0)
@@ -257,5 +257,5 @@ func TestComplexPlacement(t *testing.T) {
 	assert.NoError(t, node3.AddContainer(createContainer("c3", config)))
 
 	// check that it ends up on the same node as the 3G
-	assert.Equal(t, node2.ID, node3.ID, "")
+	assert.Equal(t, node2.ID, node3.ID)
 }
--- a/scheduler/strategy/binpacking_test.go
+++ b/scheduler/strategy/binpacking_test.go
@@ -226,3 +226,36 @@ func TestPlaceContainerDemo(t *testing.T) {
 	assert.Equal(t, node2.ID, node2bis.ID, "")
 	assert.Equal(t, len(node2.Containers()), len(node2bis.Containers()), "")
 }
+
+func TestComplexPlacement(t *testing.T) {
+	s := &BinPackingPlacementStrategy{}
+
+	nodes := []*cluster.Node{}
+	for i := 0; i < 2; i++ {
+		nodes = append(nodes, createNode(fmt.Sprintf("node-%d", i), 4, 4))
+	}
+
+	// add one container 2G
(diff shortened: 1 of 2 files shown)


jhamrick:
I needed to rebase because I didn't sign off, but you can see the test failure before I added the fix here: https://travis-ci.org/docker/swarm/jobs/47601356

phemmer:
Perhaps we should change the name of the variable as well. `( A + B ) / 2` isn't a total, it's an average :-)

vieux:
thank you, @jhamrick it's much more clear with a test :D 

Do we need the `/ 2` ? If it's just a score ?


```diff
diff --git a/scheduler/strategy/binpacking.go b/scheduler/strategy/binpacking.go
index 52203d6..46d7307 100644
--- a/scheduler/strategy/binpacking.go
+++ b/scheduler/strategy/binpacking.go
@@ -41,10 +41,8 @@ func (p *BinPackingPlacementStrategy) PlaceContainer(config *dockerclient.Contai
                if config.Memory > 0 {
                        memoryScore = (node.ReservedMemory() + config.Memory) * 100 / nodeMemory
                }
-               var total = (cpuScore + memoryScore) / 2
-
                if cpuScore <= 100 && memoryScore <= 100 {
-                       scores = append(scores, &score{node: node, score: total})
+                       scores = append(scores, &score{node: node, score: cpuScore + memoryScore})
                }
        }
```

ping @aluzzardi

jhamrick:
Yeah, dividing by 2 technically isn't even necessary because all that matters is the sort order, not the absolute score. I'll add the change from @vieux , thanks.

aluzzardi:
@crosbymichael Can you take a look into this one?

scheduler/strategy/binpacking_test.go
```diff
@@ -226,3 +226,36 @@ func TestPlaceContainerDemo(t *testing.T) {
...
+	assert.NoError(t, node1.AddContainer(createContainer("c1", config)))
+
+	// add one container 3G
+	config = createConfig(3, 0)
+	node2, err := s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.NoError(t, node2.AddContainer(createContainer("c2", config)))
+
+	// check that they end up on separate nodes
+	assert.NotEqual(t, node1.ID, node2.ID, "")
```
aluzzardi: The last argument of NotEqual can be be omitted (instead of being an empty string).

scheduler/strategy/binpacking_test.go
```diff
@@ -226,3 +226,36 @@ func TestPlaceContainerDemo(t *testing.T) {
...
+
+	// add one container 1G
+	config = createConfig(1, 0)
+	node3, err := s.PlaceContainer(config, nodes)
+	assert.NoError(t, err)
+	assert.NoError(t, node3.AddContainer(createContainer("c3", config)))
+
+	// check that it ends up on the same node as the 3G
+	assert.Equal(t, node2.ID, node3.ID, "")
```
aluzzardi: Same here.

aluzzardi:
Thank you so much @jhamrick for tracking this down and fixing it.

There a couple of nitpicks but overall looks good.

I'll let @crosbymichael take a look at it before merging.

crosbymichael:
LGTM

jhamrick:
Updated as per @aluzzardi 's comments.

Thanks!

vieux:
Thanks @jhamrick

aluzzardi:
Thanks @jhamrick!

PR #266, opened 2015-01-21
use cross-compile compatible home dir detection

MDrollette:
The use of `os/user` does not work when cross compiling this binary on Mac. It requires cgo and will fail at runtime with the message `user: Current not implemented on linux/amd64 `. This PR replaces it's use in a default flag value with a cross-compilation compatible version from https://github.com/mitchellh/go-homedir

### DIFF
--- a/flags.go
+++ b/flags.go
@@ -1,21 +1,17 @@
 package main
 
 import (
-	"os"
-	"os/user"
-	"path"
-
 	"github.com/codegangsta/cli"
-
-	log "github.com/Sirupsen/logrus"
+	"os"
+	"path/filepath"
+	"runtime"
 )
 
 func homepath(p string) string {
-	usr, err := user.Current()
-	if err != nil {
-		log.Fatal(err)
+	if runtime.GOOS == "windows" {
+		return filepath.Join(os.Getenv("USERPROFILE"), p)
 	}
-	return path.Join(usr.HomeDir, p)
+	return filepath.Join(os.Getenv("HOME"), p)
 }
 
 func getDiscovery(c *cli.Context) string {


vieux:
I'm pretty sure 


```
func getHomeDir() string {
		if runtime.GOOS == "windows" {
			return os.Getenv("USERPROFILE")
		}
		return os.Getenv("HOME")
}
```

is enough

aluzzardi:
Thanks @MDrollette!

I'm against introducing unnecessary dependencies, but mitchellh/go-homedir seems really simple.

@vieux ?

vieux:
Ok, why not. 

I don't like the name of the func being `homepath` and you have to use `~` in the argument.

Could you rename it `expandPath` ? Thanks

aluzzardi:
Looking at the code, it seems to be complex for what we need (it basically can resolve any user dir, e.g. ~foo, instead of just the current user), and @vieux's fix seems simple enough.

Thoughts?

phemmer:
Using `$HOME` isn't reliable. There are many cases where it can be stripped out of the environment. Though maybe this is acceptable, and people should manually specify the path in these cases.

MDrollette:
I'm fine with using @vieux's simplified version. It's only for a default value, and if a user is stripping the normal ENV vars then they'll just specify their preferred rootdir.

flags.go
```diff
@@ -1,21 +1,17 @@
...
+	"path/filepath"
+	"runtime"
 )
 
 func homepath(p string) string {
-	usr, err := user.Current()
-	if err != nil {
-		log.Fatal(err)
+	if runtime.GOOS == "windows" {
+		return os.Getenv("USERPROFILE")
```
aluzzardi: You need to `path.Join` on Windows as well.

vieux:
LGTM

vieux:
@MDrollette thanks

flags.go
```diff
@@ -1,21 +1,17 @@
...
+	"os"
+	"path/filepath"
+	"runtime"
 )
 
 func homepath(p string) string {
-	usr, err := user.Current()
-	if err != nil {
-		log.Fatal(err)
+	if runtime.GOOS == "windows" {
```
aluzzardi: nitpick:

Instead of repeating that `filepath.Join` twice, how about something like this?

```go
var home string
switch runtime.GOOS {
default:
    home = os.Getenv("HOME")
case "windows":
    home = os.Getenv("USERPROFILE")
}
return filepath.Join(home, p)
```

aluzzardi: Nevermind, this PR got already merged so I opened https://github.com/docker/swarm/pull/278

Thanks!

PR #269, opened 2015-01-21
add docker stats support

vieux:

### DIFF
--- a/api/utils.go
+++ b/api/utils.go
@@ -67,7 +67,7 @@ func proxy(tlsConfig *tls.Config, addr string, w http.ResponseWriter, r *http.Re
 
 	copyHeader(w.Header(), resp.Header)
 	w.WriteHeader(resp.StatusCode)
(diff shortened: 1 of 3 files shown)


vieux:
ping @crosbymichael @icecrime :dancer:

api/utils.go
```diff
@@ -67,7 +67,7 @@ func proxy(tlsConfig *tls.Config, addr string, w http.ResponseWriter, r *http.Re
 
 	copyHeader(w.Header(), resp.Header)
 	w.WriteHeader(resp.StatusCode)
-	io.Copy(w, resp.Body)
+	io.Copy(NewWriteFlusher(w), resp.Body)
```
aluzzardi: Do we need to use the write flusher for every proxy?

vieux: It doesn't hurt I believe, it's good improvement overall.

aluzzardi:
LGTM

PR #270, opened 2015-01-22
Replace --debug with --log-level

aluzzardi:


vieux:
I like this but can we keep `--debug` and the env var `DEBUG` to be consistent with docker ? I mean have both.

aluzzardi:
Updated. PTAL @vieux

vieux:
LGTM

PR #271, opened 2015-01-22
add support for history, inspect for images and improve inspect for containers

vieux:

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -163,36 +163,39 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 
 // GET /containers/{name:.*}/json
 func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
-	container := c.cluster.Container(mux.Vars(r)["name"])
-	if container != nil {
-		client, scheme := newClientAndScheme(c.tlsConfig)
-
-		resp, err := client.Get(scheme + "://" + container.Node.Addr + "/containers/" + container.Id + "/json")
-		if err != nil {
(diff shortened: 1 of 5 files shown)


api/api.go
```diff
@@ -292,6 +295,27 @@ func proxyContainer(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// Proxy a request to the right node
+func proxyImage(c *context, w http.ResponseWriter, r *http.Request) {
+	name := mux.Vars(r)["name"]
+
+	for _, node := range c.cluster.Nodes() {
+		for _, image := range node.Images() {
```
aluzzardi: In the same way we have `cluster.Containers()` and `cluster.Container(name)`, how about `node.Images()` and `node.Image(name)`?

In that case you could move the image matching code back there.

api/api.go
```diff
@@ -292,6 +295,27 @@ func proxyContainer(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+// Proxy a request to the right node
+func proxyImage(c *context, w http.ResponseWriter, r *http.Request) {
```
aluzzardi: Not sure how this would work - we only proxy to the first node containing the image.

I assume inspect would be slightly different depending on the node?

For instance, two nodes might have different images therefore the inspect would be different.

aluzzardi: Same for history

vieux: I agree, you can use the ID if you want to be exact

vieux:
@aluzzardi I added `node.Image(...)`

aluzzardi:
LGTM

PR #273, opened 2015-01-22
Proposal: port filter for the host mode

chanwit:
Currently, `swarm` supports port filtering for the `bridge` mode by default.
But behaviour of the port filter for the `host` mode is still undefined.

This PR proposes an improvement for the port filter to support `--net=host` mode.
It detects if `HostConfig.NetworkMode` is set to be `"host"` and finds available node through the value of `ContainerConfig.ExposedPorts`.

### DIFF
--- a/scheduler/filter/port.go
+++ b/scheduler/filter/port.go
@@ -20,6 +20,30 @@ func (p *PortFilter) Name() string {
 
 // Filter is exported
 func (p *PortFilter) Filter(config *dockerclient.ContainerConfig, nodes []*node.Node) ([]*node.Node, error) {
+	if config.HostConfig.NetworkMode == "host" {
+		return p.filterHost(config, nodes)
+	}
+
+	return p.filterBridge(config, nodes)
+}
+
+func (p *PortFilter) filterHost(config *dockerclient.ContainerConfig, nodes []*node.Node) ([]*node.Node, error) {
+	for port := range config.ExposedPorts {
+		candidates := []*node.Node{}
+		for _, node := range nodes {
+			if !p.portAlreadyExposed(node, port) {
+				candidates = append(candidates, node)
+			}
+		}
+		if len(candidates) == 0 {
+			return nil, fmt.Errorf("unable to find a node with port %s available in the Host mode", port)
+		}
+		nodes = candidates
+	}
+	return nodes, nil
+}
+
+func (p *PortFilter) filterBridge(config *dockerclient.ContainerConfig, nodes []*node.Node) ([]*node.Node, error) {
 	for _, port := range config.HostConfig.PortBindings {
 		for _, binding := range port {
 			candidates := []*node.Node{}
@@ -35,6 +59,19 @@ func (p *PortFilter) Filter(config *dockerclient.ContainerConfig, nodes []*node.
 		}
 	}
 	return nodes, nil
+}
+
+func (p *PortFilter) portAlreadyExposed(node *node.Node, requestedPort string) bool {
+	for _, c := range node.Containers {
+		if c.Info.HostConfig.NetworkMode == "host" {
+			for port := range c.Info.Config.ExposedPorts {
+				if port == requestedPort {
+					return true
+				}
+			}
+		}
+	}
+	return false
 }
 
 func (p *PortFilter) portAlreadyInUse(node *node.Node, requested dockerclient.PortBinding) bool {
--- a/scheduler/filter/README.md
+++ b/scheduler/filter/README.md
@@ -266,6 +266,48 @@ $ docker run -d -p 80:80 nginx
 2014/10/29 00:33:20 Error response from daemon: no resources available to schedule container
 ```
 
+### Port filter in Host Mode
+
+Docker in the host mode, running with `--net=host`, differs from the default `bridge` mode as the `host` mode does not perform any port binding. So, it requires to explicitly expose one or more port numbers (using `EXPOSE` in the Dockerfile or `--expose` on the command line). `Swarm` makes use of this information in conjunction with the `host` mode to choose an available node for a new container.
+
+For example, the following commands start `nginx` on 3-node cluster.
+```
+$ docker run -d --expose=80 --net=host nginx
+640297cb29a7
+$ docker run -d --expose=80 --net=host nginx
+7ecf562b1b3f
+$ docker run -d --expose=80 --net=host nginx
+09a92f582bc2
+```
+
+Port binding information will not be available through `ps` command because they are all started in the `host` mode.
+```
+$ docker ps
+CONTAINER ID        IMAGE               COMMAND                CREATED                  STATUS              PORTS               NAMES
+640297cb29a7        nginx:1             "nginx -g 'daemon of   Less than a second ago   Up 30 seconds                           box3/furious_heisenberg
+7ecf562b1b3f        nginx:1             "nginx -g 'daemon of   Less than a second ago   Up 28 seconds                           box2/ecstatic_meitner
+09a92f582bc2        nginx:1             "nginx -g 'daemon of   46 seconds ago           Up 27 seconds                           box1/mad_goldstine
+```
+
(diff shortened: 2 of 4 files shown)


aluzzardi:
Hey @chanwit, we are planning to cut the final `0.1.0` next week.

Do you want to include this by then or should it go into the next release?

Thanks

chanwit:
@aluzzardi I am not hurry with this one and want to test it a bit more. It's OK for me to go into the next release.

aluzzardi:
Awesome, thanks for all the contributions, we appreciate it!

chanwit:
rebased

chanwit:
@vieux @aluzzardi please take a look when you feel comfortable. I am not in hurry with this PR. Thank you.

aluzzardi:
Thanks @chanwit, awesome job.

We'll work on getting this merged once 0.1.0 is out as we are in a code freeze.

We still don't know how we are going to handle releases in the future - perhaps we should merge stuff into master anyway and keep a "release" branch where we can cherrypick fixes into (so we'd freeze "release" instead of "master").

Thoughts?

/cc @vieux

chanwit:
@aluzzardi I am a fan of trunk-based development. I like your branching model and cherry-picking to the "release" branch :-)

chanwit:
@aluzzardi is it possible to put this into 0.2.0? Thank you!

/cc @vieux

vieux:
@chanwit done, can you rebase so we can take another look ?

chanwit:
I've done the rebase, @vieux  could you please review it.

aluzzardi:
Looking into it - I need some more time to review it :)

chanwit:
@aluzzardi no problem. Thank you!

vieux:
@aluzzardi ping

vieux:
@abronan ping

aluzzardi:
@chanwit: Do you think you could write an integration test for this? It looks like a great candidate for integration.

chanwit:
@alzzardi yes. I'll do :)

abronan:
The behavior of `--net` on a single daemon is that stopped containers are not taken into account when checking for a bound port.

With the PR I can do:

```
$ docker run -d --expose=80 --net=host nginx
$ docker run -d --expose=80 --net=host nginx
$ docker run -d --expose=80 --net=host nginx

$ docker ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS                  PORTS               NAMES
d44bd614b9e0        nginx:latest        "nginx -g 'daemon of   3 seconds ago       Up Less than a second                       swarm-agent-0/boring_thompson
b0aa26ce1861        nginx:latest        "nginx -g 'daemon of   4 seconds ago       Up Less than a second                       swarm-agent-1/kickass_ritchie
7bf393922e99        nginx:latest        "nginx -g 'daemon of   5 seconds ago       Up Less than a second                       swarm-agent-2/modest_lumiere

$ docker stop d44bd614b9e0

$ docker run -d --expose=80 --net=host nginx
FATA[0000] Error response from daemon: unable to find a node with port 80/tcp available in the Host mode
```

Which is not the expected behavior. You should check for bound ports in host mode on *running* containers only.

aluzzardi:
Actually, we are doing the check on stopped containers as well on purpose in Swarm.

The reason is because of race conditions: `docker run` is a 2 step process, first we create then we start. If you run `docker run` in parallel (or fast enough), both containers end up on the same machine (since the first one wasn't started yet so we believe the port is available), then the first one is started and the second one fails to start.

abronan:
@aluzzardi Hmm I see. Shouldn't we wait to merge this until we change the scheduling part then? My take on it is that with scheduling decisions made fresh out of the store we could avoid these kind of race conditions by checking on: **running** containers + pending configs waiting to be started.

At least we will have a consistent behavior for `--host`. What do you think @chanwit? (sorry for delaying again on this one)

aluzzardi:
@abronan The thing is, containers can be directly created on the host bypassing swarm, thus we can't just rely on the store.

chanwit:
I have no problem waiting for this. I'll do an integration test for this PR in the mean time :) :)

chanwit:
rebased

chanwit:
@aluzzardi an integeration test is now in-place, PTAL.

abronan:
Thank you @chanwit!

@aluzzardi Should we really allow this though? I think that we should advise against bypassing Swarm in the doc for example. Going through Masters/Managers ensures that the system stays consistent. Swarm agents are part of a set, treating them individually by bypassing Swarm will only result in tricky situations to solve. If only local docker daemons were linked somehow to the Swarm storage we could ensure that requests bypassing Swarm are in sync with the Distributed state. But as of now this is not the case :(

chanwit:
@abronan, with @aluzzardi's explanation I think the point you've been raising is about the *port filter* mechanism as a whole, not specifically to this PR? If so, I think we should have another PR for discussing this :)

scheduler/filter/port.go
```diff
@@ -20,23 +20,51 @@ func (p *PortFilter) Name() string {
 
 // Filter is exported
 func (p *PortFilter) Filter(config *dockerclient.ContainerConfig, nodes []cluster.Node) ([]cluster.Node, error) {
-	for _, port := range config.HostConfig.PortBindings {
-		for _, binding := range port {
+	if config.HostConfig.NetworkMode == "host" {
```
aluzzardi: Minor nit: This function is a bit complex, could you break it down to two functions, `filterHost` and `filterBridge`, and in `Filter` do something like:

```
if config.HostConfig.NetworkMode == "host" {
  return filterHost(...)
} else {
  return filterBridge(...)
}
```

test/integration/host-mode-port-filter.bats
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bats
```
aluzzardi: Could we call this `port-filter.bats`? I think this would be the right home for other integration tests such as port filtering in bridge mode.

test/integration/host-mode-port-filter.bats
```diff
@@ -0,0 +1,37 @@
...
+
+@test "docker should filter port in host mode correctly" {
+	start_docker 3
+	start_manager
+
+	#
+	# Use busybox to save image pulling time for integration test.
+	# Running the first 3 containers, it should be fine.
+	#
+	run docker_swarm run -d --expose=80 --net=host busybox sleep 30
```
aluzzardi: For containers that should run forever, perhaps `sh` would be a better command than `sleep`

test/integration/host-mode-port-filter.bats
```diff
@@ -0,0 +1,37 @@
...
+
+load helpers
+
+function teardown() {
+	stop_manager
+	stop_docker
+}
+
+@test "docker should filter port in host mode correctly" {
+	start_docker 3
```
aluzzardi: I think starting `2` should prove that it runs correctly (the 3rd one will fail), maybe there's no need to start 3.

aluzzardi:
Agreed with @chanwit - the issue is in the port filter as a whole, for good reasons (at least until we get a store).

LGTM as is, with a few minor nits. @chanwit: Could you take a look?

ping @SvenDowideit for docs review

abronan:
Overall besides the small issue I outlined above, I think we should merge the PR. We could fix the behavior and discuss about it in another one or when we'll have the store.

LGTM (and thanks @chanwit, this PR is really useful!)

chanwit:
@abronan @aluzzardi PTAL. Thank you.

abronan:
Sorry @chanwit, it needs a rebase after merging a huge refactoring PR (#574).

chanwit:
@abronan thank you, rebase is done.
Could you please take a look.

abronan:
@chanwit I think you forgot to delete the old `host-mode-port-filter.bats` from the PR. Should be good to go after this :+1:

chanwit:
@abronan sorry my bad. The file has been deleted.

abronan:
Awesome! Thanks @chanwit!

Issue #272, opened 2015-01-22
Problem when specifying ports on docker run


technolo-g:
It seems like Swarm is treating an exposed port as a statically mapped port. If I have a cluster of 3 nodes and create 3 containers with this command: 
`docker run -p 80 -d -t nginx:latest` 

I get this error message upon creating the 4th: 
`2015/01/21 22:53:02 Error response from daemon: unable to find a node with port  available`. 

I would expect the above error if I were provisioning with the following command 4 times on a 3 node cluster:
`docker run -p 80:80 -d -t nginx:latest`

In addition, this does not happen when specifying `-P`.  
Here is the output of a transaction and the associated Swarm logs:

Shell:
```
[22:52:49 mbajor@mbajor-DraakBookPro:~/Dev/docker-swarm/master]
$ docker run -p 80 -d -t nginx:latest
3b56e54020eb78946b83d1b5a8b1b25eac0552028680c4d94db79f5bd9056bea
[22:52:55 mbajor@mbajor-DraakBookPro:~/Dev/docker-swarm/master]
$ docker run -p 80 -d -t nginx:latest
b1c5d2e872c136683b9e555ff10072f77db43e144b6779290013eeacac1beaa2
[22:52:56 mbajor@mbajor-DraakBookPro:~/Dev/docker-swarm/master]
$ docker run -p 80 -d -t nginx:latest
79b938ec3a1ea16563f0c5ecdcf86e59e4a08df43785ea4d8e5fa3329fab71be
[22:52:57 mbajor@mbajor-DraakBookPro:~/Dev/docker-swarm/master]
$ docker run -p 80 -d -t nginx:latest
2015/01/21 22:53:02 Error response from daemon: unable to find a node with port  available
[22:53:02 mbajor@mbajor-DraakBookPro:~/Dev/docker-swarm/master]
$ docker ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS                  PORTS                                   NAMES
79b938ec3a1e        nginx:latest        "nginx -g 'daemon of   16 seconds ago      Up Less than a second   443/tcp, 10.100.199.202:49156->80/tcp   dockerhost02/angry_pike
b1c5d2e872c1        nginx:latest        "nginx -g 'daemon of   17 seconds ago      Up Less than a second   443/tcp, 10.100.199.203:49164->80/tcp   dockerhost03/grave_leakey
[...]

PR #274, opened 2015-01-22
fix to not filter random ports

chanwit:
This PR is an attempt to fix issue #272. It allows a request with random port to go though the filter.

### DIFF
--- a/scheduler/filter/port.go
+++ b/scheduler/filter/port.go
@@ -35,7 +35,10 @@ func (p *PortFilter) portAlreadyInUse(node *cluster.Node, requested dockerclient
 	for _, c := range node.Containers() {
 		for _, port := range c.Info.HostConfig.PortBindings {
 			for _, binding := range port {
-				if binding.HostPort == requested.HostPort {
+				// Check if requested port is already mapped.
+				// But when the requested port is blank, we're requesting a random port.
+				// So, let it goes through the filter.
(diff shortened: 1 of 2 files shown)


scheduler/filter/port.go
```diff
@@ -35,7 +35,7 @@ func (p *PortFilter) portAlreadyInUse(node *cluster.Node, requested dockerclient
 	for _, c := range node.Containers() {
 		for _, port := range c.Info.HostConfig.PortBindings {
 			for _, binding := range port {
-				if binding.HostPort == requested.HostPort {
+				if binding.HostPort == requested.HostPort && binding.HostPort != "" && requested.HostPort != "" {
```
aluzzardi: nit:

Since we already asserted that `binding.HostPort` and `requested.HostPort` are the same, there is probably no need to check if both are `!= ""`.

Maybe only check `requested.HostPort`?

Otherwise, LGTM

chanwit: fixed! thank you, @aluzzardi

chanwit:
ping @aluzzardi 
/cc @technolo-g for testing

technolo-g:
LGTM

chanwit:
@technolo-g #310 would replace this PR. I'll close this after #310 get merged.

vieux:
#310 was merged

PR #277, opened 2015-01-22
add info in join

### LINKED ISSUE #275: Swarm config
I''m trying swarm and I have some doubts related with the configuration. It looks so simple but I cant do it work. I have two servers, both with docker `1.4.1` and `go1.3.3`. I configure docker service to listen by a port 

I can run the client with -H option to send commands between servers. Now come swarm part. I create de unique id from one of their and I try to join to first one from the other, but the command is frozen. Daemons doesn't listen anything 

Any debug test to debug it?

vieux:
Close #275

The current behaviour of `swarm join` isn't clear, some people think it's stuck because there is nothing printed.
I end the sentence by `...` so people know it will never end.

@aluzzardi suggestions are welcome.

### DIFF
--- a/join.go
+++ b/join.go
@@ -37,8 +37,9 @@ func join(c *cli.Context) {
 
 	hb := time.Duration(c.Int("heartbeat"))
 	for {
+		log.Infof("Registering %q on the discovery service %q every %d seconds...", addr, dflag, hb)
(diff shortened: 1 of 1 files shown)


aluzzardi:
LGTM

PR #282, opened 2015-01-23
Use actual binary name instead of hardcoding it

radeksimko:
See https://github.com/docker/machine/pull/292 for reasons.

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"fmt"
 	"os"
+	"path"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
@@ -18,7 +19,7 @@ import (
 
 func main() {
 	app := cli.NewApp()
-	app.Name = "swarm"
+	app.Name = path.Base(os.Args[0])
 	app.Usage = "a Docker-native clustering system"
 	app.Version = "0.1.0"
 	app.Author = ""
@@ -78,7 +79,7 @@ func main() {
 			Action: func(c *cli.Context) {
 				dflag := getDiscovery(c)
 				if dflag == "" {
-					log.Fatal("discovery required to list a cluster. See 'swarm list --help'.")
+					log.Fatalf("discovery required to list a cluster. See '%s list --help'.", c.App.Name)
 				}
 
 				d, err := discovery.New(dflag, 0)
--- a/help.go
+++ b/help.go
@@ -1,13 +1,18 @@
 package main
 
-import "github.com/codegangsta/cli"
+import (
+	"os"
+	"path"
+
+	"github.com/codegangsta/cli"
+)
 
 func init() {
-
+	// See https://github.com/codegangsta/cli/pull/171/files
 	cli.CommandHelpTemplate = `{{$DISCOVERY := or (eq .Name "manage") (eq .Name "join") (eq .Name "list")}}NAME:
    {{.Name}} - {{.Usage}}
 USAGE:
-   swarm {{.Name}}{{if .Flags}} [options]{{end}} {{if $DISCOVERY}}<discovery>{{end}}{{if .Description}}
+   ` + path.Base(os.Args[0]) + ` {{.Name}}{{if .Flags}} [options]{{end}} {{if $DISCOVERY}}<discovery>{{end}}{{if .Description}}
 DESCRIPTION:
    {{.Description}}{{end}}{{if $DISCOVERY}}
 ARGUMENTS:
--- a/join.go
+++ b/join.go
@@ -17,7 +17,7 @@ func checkAddrFormat(addr string) bool {
 func join(c *cli.Context) {
 	dflag := getDiscovery(c)
 	if dflag == "" {
-		log.Fatal("discovery required to join a cluster. See 'swarm join --help'.")
+		log.Fatalf("discovery required to join a cluster. See '%s join --help'.", c.App.Name)
 	}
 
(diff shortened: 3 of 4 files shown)


SvenDowideit:
yes please - this is very very useful when documenting things for users.

radeksimko:
@SvenDowideit 


I think swarm is relatively unique name, so one of the reasons mentioned in docker/machine may not apply here. Generally, built-in documentation (manual pages) can be generated from real name and the documentation here on github could stay as is and expect, that in most cases people are going to use it as `swarm`.

phemmer:
I have no strong feelings on either way (hard code, or programmatic), but I do have strong feelings that we should not use `os.Args[0]` directly. This will pull in the full path to the binary if the full path was used to start it. `path.Base(os.Args[0])` is much better.

radeksimko:
@phemmer Agreed, :+1: 

I've modified the PR.

vieux:
@radeksimko I think is we do this, we need to also update `CommandHelpTemplate` in `help.go`

radeksimko:
@vieux I agree, there's a few solutions for that:

1. Send a patch to the [`codegangsta/cli`](https://github.com/codegangsta/cli/blob/75ffa2d83fc35e7dc0520906679232bd56223959/help.go#L81-83), so we can get access to the parent context (parent `Command`) inside the template as this is currently the context we have available: `cli.Command{Name:"list", ShortName:"l", Usage:"list nodes in a cluster", Description:"", BashComplete:(func(*cli.Context))(nil), Before:(func(*cli.Context) error)(nil), Action:(func(*cli.Context))(0x5a90), Subcommands:[]cli.Command(nil), Flags:[]cli.Flag(nil), SkipFlagParsing:false, HideHelp:false}`
2. hardcode `command` - many other apps using that library do as they use [the default template](https://github.com/codegangsta/cli/blob/master/help.go#L36)
3. Hack around somehow

I think that it would be nice to solve it for everybody, so I'll take the 1st approach.

vieux:
@radeksimko I agree, (1) is the best.

Though, in the meantime we could simply do

```diff
diff --git a/help.go b/help.go
index 6361279..80fe465 100644
--- a/help.go
+++ b/help.go
@@ -1,13 +1,18 @@
 package main

-import "github.com/codegangsta/cli"
+import (
+       "os"
+       "path"
+
+       "github.com/codegangsta/cli"
+)

 func init() {

        cli.CommandHelpTemplate = `{{$DISCOVERY := or (eq .Name "manage") (eq .Name "join") (eq .Name "list")}}NAME:
    {{.Name}} - {{.Usage}}
 USAGE:
-   swarm {{.Name}}{{if .Flags}} [options]{{end}} {{if $DISCOVERY}}<discovery>{{end}}{{if .Description}}
+   ` + path.Base(os.Args[0]) + ` {{.Name}}{{if .Flags}} [options]{{end}} {{if $DISCOVERY}}<discovery>{{end}}{{if .Description}}
 DESCRIPTION:
    {{.Description}}{{end}}{{if $DISCOVERY}}
 ARGUMENTS:
```

vieux:
not much of a hack to me

radeksimko:
@vieux Cool, I can update this PR + already prepared relevant PR in upstream as mentioned earlier:
https://github.com/codegangsta/cli/pull/171

aluzzardi:
@SvenDowideit Yay or nay for this one?

radeksimko:
@vieux PR updated, I also changed the hardcoded `swarm` in the `list` subcommand error message.

main.go
```diff
@@ -78,7 +79,8 @@ func main() {
 			Action: func(c *cli.Context) {
 				dflag := getDiscovery(c)
 				if dflag == "" {
-					log.Fatal("discovery required to list a cluster. See 'swarm list --help'.")
+					log.Fatal("discovery required to list a cluster. " +
+						fmt.Sprintf("See '%v list --help'.", c.App.Name))
```
vieux: `join.go` and `manage.go` have similar messages.

vieux: could you update thoses as well ? Thanks

SvenDowideit:
the main reason its useful, is when developing, or making a custom version, its better to see the actual name of the application - as rebuilding a custom binary is how plugins seem to be working, this will mater then too

for example, I have a set of boot2docker-v1.X binaries, and some boot2docker-v1.X-vmware (or other local hacks) - and having them all report `boot2docker` isn't helpful when I'm then copy and pasting that info into documentation for others to consume.

netroby:
In most case, you may using alias or link -s to point differeent binary files
ln -s /path/to/real/swarm  /usr/bin/swarm-0.1 or what ever
hard code swarm is ok for me, because it save the path resolve.

main.go
radeksimko: Updated, I used `log.Fatalf` instead to keep the code a bit cleaner.

vieux:
@radeksimko you commited the `docker-swarm` binary

radeksimko:
@vieux Noticed, fixed, you're faster then I expected :)

vieux:
@radeksimko why did you cut the `log.Fatalf(` in 2 lines ? It's not formatted properly

radeksimko:
> nit: Why %v when Name is a string?

Fair point, sorry, changed to `%s`.

> why did you cut the log.Fatalf( in 2 lines ? It's not formatted properly

To keep it under 80 chars per line, I put it back on one line now.

vieux:
@radeksimko thanks!

vieux:
LGTM

PR #283, opened 2015-01-23
Cleanup: implement full path creation for zookeeper

chanwit:
This PR is a cleanup to support full path creation for ZooKeeper. The reason is that `ZooKeeper` is more rigid than `Etcd` and others, as we need to firstly add the top-level path before adding the next level path.

Currently, if a user is trying to use path like `/path/subpath` an unexpected error will be occurred.

### DIFF
--- a/discovery/zookeeper/zookeeper.go
+++ b/discovery/zookeeper/zookeeper.go
@@ -13,12 +13,30 @@ import (
 
 type ZkDiscoveryService struct {
 	conn      *zk.Conn
-	path      string
+	path      []string
 	heartbeat int
 }
 
 func init() {
 	discovery.Register("zk", &ZkDiscoveryService{})
+}
+
+func (s *ZkDiscoveryService) fullpath() string {
+	return "/" + strings.Join(s.path, "/")
+}
+
+func (s *ZkDiscoveryService) createFullpath() error {
+	for i := 1; i <= len(s.path); i++ {
+		newpath := "/" + strings.Join(s.path[:i], "/")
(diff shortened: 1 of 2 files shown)


aluzzardi:
@chanwit Should we instead implement the full path creation instead of forbidding using subpaths?

vieux:
I agree with @aluzzardi it would be nice to implement the full path creation

chanwit:
@aluzzardi @vieux OK, I'll go with full path :)

aluzzardi:
Awesome, thanks!

aluzzardi:
LGTM

discovery/zookeeper/zookeeper.go
```diff
@@ -21,6 +21,24 @@ func init() {
...
 }
 
+func (s *ZkDiscoveryService) fullpath() string {
+	return "/" + strings.Join(s.path, "/")
+}
+
+func (s *ZkDiscoveryService) createFullpath() error {
+	for i := 1; i <= len(s.path); i++ {
+		newpath := "/" + strings.Join(s.path[:i], "/")
+		_, err := s.conn.Create(newpath, []byte{1}, 0, zk.WorldACL(zk.PermAll))
```
chenchun: @aluzzardi can you share the thoughts of `[]byte{1}` here? 
I doubt of the fix in https://github.com/docker/swarm/issues/974

abronan: @chenchun good catch, I'm also intrigued now..

chenchun: Yes, quit weird. :smiley:

PR #294, opened 2015-01-24
Add missing --discovery option in Example usage

ankushagarwal:
Signed-off-by: Ankush Agarwal <ankushagarwal11@gmail.com>


vieux:
@ankushagarwal we removed the `--discovery` flag although the rc-1 still uses it, the final beta won't.
You can use master if you want to try it.

Thanks!

PR #297, opened 2015-01-25
fix tests that were not compiling or panicking

mota:
Further to the client.StartMonitorEvents prototype change tests where either not compiling or panicking (mock got more arguments than expected).


vieux:
LGTM

Issue #298, opened 2015-01-25
Proposal: Affinity values as regexp patterns


mota:
Currently, *affinity* values can be considered as either a regexp pattern if bounded by slashes, or a globbing pattern.
In the latter, any wildcard encountered is replaced by the regexp equivalent (".*") for a greedy match. Apart from this replacement, anything else in a value can be matched as a valid regexp special character, making the whole value a mere regexp pattern.

Furthermore, in the globbing case, the expression is encompassed between the start/end of line characters, making it difficult to match simply a container or an image name as specified in the README.md document of the `swarm.scheduler.filter` package.

To obtain the convenience of partially specify an affinity value name, one is forced to put at least a wildcard character, which must be escaped when using cli.

I propose to simply remove the distinction between regexp and globbing patterns and make the affinity value plain simple regexp patterns, not bounded by start and end characters. 

If not possible to drop the support of the globbing pattern, at least remove start and end characters of this processing.

mota:
After some research, there's a mismatch between `swarm.scheduler.filter.constraint_test` and the README.

The tests clearly reject a constraint in the form `constraint:region==us` to match the set {us-west, us-east}.
However, in the README's examples, an affinity with the form `affinity:image==redis` is supposed, I presume, match the **redis:lastest** image.

Secondly, the code currently used to perform globbing is, as I previously said, a regexp matching with some replacement to emulate the wildcard special character.

Golangs `path.filepath` [provides](http://golang.org/pkg/path/filepath/#Match) a `Match` function that matches a pattern against a string value. Since both `regexp.MatchString` and `path.filepath.Match` share the same interface, patching the code wouldn't be a problem.

Remains the problem of the greediness of the match. The way I see it, there's three solutions:
- We keep the current behavior even for affinities
- We go for a flexible match that assumes a wildcard character at the end of the pattern
- A compromise of both that allows to pass a parameter to the `expr.Match` function that tells it to opt for a greedy behavior or not

vieux:
I opened a PR to fix the `redis:lastest` issue.

But I don't get the rest of your argument.
I like to do `region==us*` (personally I'll never use regexp) I don't want to do `region==^us.*$`

chanwit:
Removal slashes causes will make things more confused.
It's ugly to control exact matches, e.g., you have to always specify `name==^node1$` for an exact match. And `name==node1` will become partially matching, say `abcnode1abc` will be also matched.

mota:
[...]

PR #299, opened 2015-01-25
Consider affinity value as a regexp patterns

mota:
See docker/swarm#298


mota:
Build failed because of tests, fixed in #297

PR #300, opened 2015-01-25
match only repo in affinity image

### LINKED ISSUE #298: Proposal: Affinity values as regexp patterns

vieux:
Close #298

### DIFF
--- a/scheduler/filter/affinity.go
+++ b/scheduler/filter/affinity.go
@@ -2,6 +2,7 @@ package filter
 
 import (
 	"fmt"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/swarm/cluster"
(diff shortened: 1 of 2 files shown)


vieux:
ping @mota, I agree this is the expected behaviour, we should test `repo` and `repo:tag`

aluzzardi:
LGTM

PR #301, opened 2015-01-26
Add error message when discovery service returns non 200

ankushagarwal:
status code. Update doc for token.go/Fetch

Signed-off-by: Ankush Agarwal <ankushagarwal11@gmail.com>

### DIFF
--- a/discovery/token/token.go
+++ b/discovery/token/token.go
@@ -36,8 +36,9 @@ func (s *TokenDiscoveryService) Initialize(urltoken string, heartbeat int) error
 	return nil
 }
 
-// FetchNodes returns the node for the discovery service at the specified endpoint
+// Fetch returns the list of nodes for the discovery service at the specified endpoint
 func (s *TokenDiscoveryService) Fetch() ([]*discovery.Node, error) {
+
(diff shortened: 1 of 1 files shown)


ankushagarwal:
Takes care of this (and whenever Docker Hub discovery service has a problem) :
image

discovery/token/token.go
```diff
@@ -52,6 +54,10 @@ func (s *TokenDiscoveryService) Fetch() ([]*discovery.Node, error) {
 		if err := json.NewDecoder(resp.Body).Decode(&addrs); err != nil {
 			return nil, err
 		}
+	} else {
+		err = errors.New(fmt.Sprintf("Failed to fetch nodes from Discovery service. "+
```
chanwit: there is `fmt.Errorf` for creating an error with formatted message.

discovery/token/token.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"errors"
```
chanwit: This will be unnecessary if you fix below.

PR #303, opened 2015-01-26
Improve container state refresh.

aluzzardi:
Currently, container inspection is performed only on creation (or during
exec by the API).

The problem is that certain informations such as NetworkSettings are not
available during creation, therefore we must inspect the containers
during other events.

This change refactors a bit the API so that RefreshContainer() and
RefreshContainers() now accept a `full` flag to force a deep refresh.

The node event handler in turn uses that flag whenever a container
starts or dies.

### DIFF
--- a/cluster/node.go
+++ b/cluster/node.go
@@ -85,10 +85,11 @@ func (n *Node) connectClient(client dockerclient.Client) error {
 	}
 
 	// Force a state update before returning.
-	if err := n.refreshContainers(); err != nil {
+	if err := n.RefreshContainers(true); err != nil {
 		n.client = nil
 		return err
 	}
+
 	if err := n.refreshImages(); err != nil {
 		n.client = nil
 		return err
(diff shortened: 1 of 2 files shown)


aluzzardi:
ping @vieux

cluster/node.go
```diff
@@ -153,32 +154,33 @@ func (n *Node) refreshImages() error {
...
 		return err
 	}
 
-	n.Lock()
-	defer n.Unlock()
-
 	merged := make(map[string]*Container)
 	for _, c := range containers {
-		merged, err = n.updateContainer(c, merged)
+		merged, err = n.updateContainer(c, merged, full)
```
vieux: `n.updateContainer` is using `n.containers` I think the lock is required before, no ?

cluster/node.go
```diff
@@ -186,66 +188,59 @@ func (n *Node) refreshContainer(ID string) error {
...
-		// This is a brand new container.
-		if err := n.inspectContainer(c, containers, false); err != nil {
-			return nil, err
-		}
-	}
 	return containers, nil
 }
 
-func (n *Node) refreshContainersAsync() {
+func (n *Node) RefreshContainersAsync() {
```
vieux: I don't see where it's called outside on the pkg

cluster/node.go
```diff
@@ -440,10 +435,18 @@ func (n *Node) String() string {
...
-	if ev.Status == "pull" || ev.Status == "untag" || ev.Status == "delete" {
+	switch ev.Status {
+	case "pull", "untag", "delete":
+		// These events refer to images so there's no need to update
+		// containers.
 		n.refreshImages()
-	} else {
-		n.refreshContainer(ev.Id)
+	case "start", "die":
+		// If the container is started or stopped, we have to do an inspect in
```
vieux: do we want to use `stop` here instead of `die` ?

vieux: or in addition ?

aluzzardi: `stop` is always preceded by `die` as far as I can tell, so this is to avoid doing inspect twice when stopping.

cluster/node.go
aluzzardi: It's just to be consistent - since `RefreshContainer()` is now "public", I've made `RefreshContainers()` and `RefreshContainersAsync` public as well.

cluster/node.go
vieux: :+1: alright

cluster/node.go
vieux: ok

cluster/node.go
aluzzardi: You're right - PTAL

vieux:
LGTM

PR #304, opened 2015-01-26
Add node's addr as a constraint.

mota:
Currently, there's 2 ways to put a constraint on a node, that is node's ID and node's name.

It would also be useful when using swarm within a somehow fixed network environment to be able to select a node by its domain name or ip address, including or excluding the listening port.

Let suppose we have a node started like this:

```$ swarm join --addr 'node1.private:2375' token://```

Then this PR allows to put a constraint on this node this ways:

```$ docker run -it -e constraint:node=node1.private:2375 busybox```

or even

```$ docker run -it -e constraint:node=node1.private busybox```

### DIFF
--- a/scheduler/filter/constraint.go
+++ b/scheduler/filter/constraint.go
@@ -2,6 +2,7 @@ package filter
 
 import (
 	"fmt"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/swarm/cluster"
@@ -26,7 +27,7 @@ func (f *ConstraintFilter) Filter(config *dockerclient.ContainerConfig, nodes []
 			switch constraint.key {
 			case "node":
 				// "node" label is a special case pinning a container to a specific node.
-				if constraint.Match(node.ID, node.Name) {
+				if constraint.Match(node.ID, node.Name, node.Addr, strings.Split(node.Addr, ":")[0]) {
 					candidates = append(candidates, node)
 				}
 			default:
--- a/scheduler/filter/constraint_test.go
+++ b/scheduler/filter/constraint_test.go
@@ -40,7 +40,7 @@ func testFixtures() (nodes []*cluster.Node) {
 	return
 }
 
-func TestConstrainteFilter(t *testing.T) {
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: scheduler/filter/constraint.go [hunk 1]
```diff
@@ -26,7 +27,7 @@ func (f *ConstraintFilter) Filter(config *dockerclient.ContainerConfig, nodes []
 			switch constraint.key {
 			case "node":
 				// "node" label is a special case pinning a container to a specific node.
-				if constraint.Match(node.ID, node.Name) {
+				if constraint.Match(node.ID, node.Name, node.Addr, strings.Split(node.Addr, ":")[0]) {
```
chanwit: if `node.Addr` does not contain `:`, this would break.
could you please do some checking before using it?

mota: No it doesn't.

vieux:
LGTM

vieux:
@aluzzardi what do you think about this one ?

aluzzardi:
I'm afraid this could mislead people.

If they set it on providers such as EC2, as soon as the machine reboots the container won't be able to schedule anymore.

mota:
Indeed but I was more considering a static environment.

Not that is what swarm is intended for but this could help to have a global docker dispatcher when working on an internal production domain. As an example I could totally use such a feature for work.

Although, I agree to the fact it can mislead people for the reason you exposed.

bacongobbler:
I think I'm missing the value proposed here. Wouldn't this behave the same way as using a node name constraint? In a more static environment, the node ID and name would always be the same.

vieux:
@bacongobbler 

ID looks like `RHIF:GSSW:XWFW:WKED:XWKS:ZN6T:VUDZ:4JXR:5NWH:WYNI:JTWR:TBDJ`
Name is hostname, so it can look like `box-017`
Addr can look like `ec2-xx-xx-xx-xx.us-west-2.compute.amazonaws.com:4242`

So it could be useful, but I agree with @aluzzardi, it could be problematic when the addr changes.

PR #308, opened 2015-01-27
Proposal: require ps -a to show swarm containers

vieux:
If you do the `swarm join` in a container, you will end up with one line per node in docker ps `on swarm join` per node.

In this PR I propose to hide the swarm containers in `docker ps`.
They will be visible with `docker ps --all`

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -136,6 +136,10 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		if !strings.Contains(tmp.Status, "Up") && !all {
 			continue
 		}
+		// Skip swarm containers unless -a was specified.
+		if strings.Split(tmp.Image, ":")[0] == "swarm" && !all {
+			continue
+		}
 		if !container.Node.IsHealthy() {
 			tmp.Status = "Pending"
 		}
--- a/api/README.md
+++ b/api/README.md
@@ -49,6 +49,9 @@ DELETE "/images/{name:.*}"
 ```
 * `GET "/containers/{name:.*}/json"`: `HostIP` replaced by the the actual Node's IP if `HostIP` is `0.0.0.0`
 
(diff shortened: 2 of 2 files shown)


vieux:
ping @ehazlett and @aluzzardi  what do you think ?

ehazlett:
+1. I think it's confusing if you see it from a normal "ps". However, being able to restart, check logs etc is absolutely necessary.

tnachen:
Is it possible to log that it's explicitly hiding all swarm images, and tell the user if they want to see it please use -a?

vieux:
@tnachen on the daemon you mean ?

aluzzardi:
+1

aluzzardi:
I think this needs to wait until we get a proper official image so we can filter that one exactly (instead of everything named "somethingsomething/swarm".

ehazlett:
@aluzzardi +1 for official image :)

vieux:
ok I updated the code, this should match only the official image. let's wait

api/api.go
```diff
@@ -136,6 +136,10 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		if !strings.Contains(tmp.Status, "Up") && !all {
 			continue
 		}
+		// Skip swarm containers unless -a was specified.
+		if string.Split(strings.TrimPrefix(tmp.Image, "library/"), ":")[0] == "swarm" && !all {
```
aluzzardi: Can't we just check `== library/swarm`?

Also, it should be `strings`.

vieux: I'm not sur, I believe the `library/` is stripped by docker already. Let's wait for the official image

vieux:
@aluzzardi it's good to merge

aluzzardi:
LGTM

PR #312, opened 2015-01-27
Cleanup: fix fatal error when registering node

chanwit:
The extra leading `/` causes fatal error when registering nodes. This PR fixes it.


vieux:
LGTM

PR #302, opened 2015-01-26
Remove the container's state from the store when destroyed

mota:
Signed-off-by: Pierre Wacrenier <pierre.wacrenier@gmail.com>

PR #314, opened 2015-01-27
Do not fail when destroying a container if it's not found in the store.

aluzzardi:
Related to #302 


vieux:
LGTM

PR #316, opened 2015-01-27
Fix various cases in matching with constraints and affinities

vieux:
`-e affinity:container!=redis` wasn't matching nodes with no container
`-e affinity:images!=redis` wasn't matching nodes with no image
`-e constraint:region!=us-west` wasn't matching containers with no label
`-e affinity:container==redis` wasn't matching containers names redis, because their name is actually `/redis`

### DIFF
--- a/scheduler/filter/expr.go
+++ b/scheduler/filter/expr.go
@@ -72,6 +72,10 @@ func parseExprs(key string, env []string) ([]expr, error) {
 	return exprs, nil
 }
 
+func (e *expr) MatchEmpty() bool {
+	return e.operator == NOTEQ
+}
+
 func (e *expr) Match(whats ...string) bool {
 	var (
 		pattern string
--- a/scheduler/filter/affinity.go
+++ b/scheduler/filter/affinity.go
@@ -26,24 +26,31 @@ func (f *AffinityFilter) Filter(config *dockerclient.ContainerConfig, nodes []*c
 		for _, node := range nodes {
 			switch affinity.key {
 			case "container":
+				if len(node.Containers()) == 0 && affinity.MatchEmpty() {
+					candidates = append(candidates, node)
+					break
+				}
 				for _, container := range node.Containers() {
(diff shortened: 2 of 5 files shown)


scheduler/filter/expr.go
```diff
@@ -72,6 +72,10 @@ func parseExprs(key string, env []string) ([]expr, error) {
 	return exprs, nil
 }
 
+func (e *expr) MatchEmpty(whats ...string) bool {
```
mota: Isn't the `whats` parameter unnecessary here?

vieux: no, I'll update

vieux:
I updated the description

aluzzardi:
I believe this is still broken.

Let's say you have two containers on node1, "A" and "B".

You run with `affinity:container!=B`.

When evaluating "A", `affinity.Match` will return `true` and will immediately accept node1 and break, even though it has "B".

We have to rethink `NOTEQ` expressions. I think we should somehow give `Match` the entire list of candidates we want to match and let it handle the not equal correctly (including returning true if the operator is `!=` and the list is empty).

vieux:
@aluzzardi you are probably right, we need to a list with all the containers and call Match only once with the list, like we do in images.

aluzzardi:
Yes, however I think we are doing it wrong in images as well: we call `Match` in a loop while going through the list of images, the `whats` should be constructed in the loop and `Match` should be called only once outside of the loop.

In that case, it doesn't make sense anymore for Match() to take varargs.

vieux:
yes, and we don't need matchempty anymore, we can just check if the match parameter is empty :P

mota:
I noticed this behavior yesterday, I made a commit based on @vieux's branch to not interfere with this PR. mota@19bba42681a86c408add05fd60850768391343e2

I also fixed tests.

vieux:
replaced by #327

PR #318, opened 2015-01-27
add missing status codes

### LINKED ISSUE #313: Incorrect http status in non-proxy cases
Re #191, it doesn't look like all cases are fixed since not everything goes through ```proxyContainer```.

In particular I see:

    POST /containers/create
    DELETE /containers/create

do not return the proper success status, which is not ```200```.

vieux:
Closes #313


vieux:
LGTM

PR #322, opened 2015-01-28
travis: go vet and go test -race.

aluzzardi:


vieux:
This is nice, we need to add `go get code.google.com/p/go.tools/cmd/vet`

PR #327, opened 2015-01-28
Fix affinity matching.

aluzzardi:


vieux:
LGTM, one comment though,

if I'm not mistaken, we don't need the special case for empty label, Match will handle it correctly.

`Match("") will return true for !=`

```diff
diff --git a/scheduler/filter/constraint.go b/scheduler/filter/constraint.go
index ecc26a5..5051d2b 100644
--- a/scheduler/filter/constraint.go
+++ b/scheduler/filter/constraint.go
@@ -30,17 +30,8 @@ func (f *ConstraintFilter) Filter(config *dockerclient.ContainerConfig, nodes []
                                        candidates = append(candidates, node)
                                }
                        default:
-                               if label, ok := node.Labels[constraint.key]; ok {
-                                       if constraint.Match(label) {
-                                               candidates = append(candidates, node)
-                                       }
-                               } else {
-                                       // The node doesn't have this particular label.
-                                       if constraint.operator == NOTEQ {
-                                               // Special case: If the operator is != and the node doesn't
-                                               // have the label at all, consider it as a candidate.
-                                               candidates = append(candidates, node)
-                                       }
+                               if constraint.Match(node.Labels[constraint.key]) {
+                                       candidates = append(candidates, node)
                                }
                        }
                }
```

If I'm wrong, we need to add one more test, because tests are passing with this patch.

aluzzardi:
Done. PTAL

vieux:
LGTM

PR #328, opened 2015-01-28
Fix a panic that occurs in the log handler when pulling images.

aluzzardi:


vieux:
LGTM

PR #330, opened 2015-01-29
Tweak Dockerfile for official image.

psftw:
Replaced ADD with COPY and ENTRYPOINT to CMD per best practices article.
Yes, this means that you'd need to type an extra "swarm" on the CLI:
``docker run swarm swarm <args>``.  Removed unnecessary Godeps install,
and added a .dockerignore to shave off ~23M.

Signed-off-by: Peter Salvatore <peter@psftw.com>


vieux:
I don't like the change `ENTRYPOINT` to `CMD`

psftw:
Commented on the reasoning behind it over in the official-images PR.  Happy to update the PR with whatever gets decided over there to be consistent.

Dockerfile
```diff
@@ -1,10 +1,10 @@
...
 WORKDIR /go/src/github.com/docker/swarm
 
-RUN go get github.com/tools/godep
-RUN godep go install
+ENV GOPATH $GOPATH:/go/src/github.com/docker/swarm/Godeps/_workspace
+RUN go install -v
 
 EXPOSE 2375
-ENTRYPOINT ["swarm"]
+CMD ["swarm"]
```
jfrazelle: I'm just going to give my 2 cents here, because isn't that what open source is all about ;)

I think it should be `ENTRYPOINT` because we should use the best practices for our Dockerfiles. So:
```console
$ docker run -d swarm swarm --version
```
is kinda weird when it should be:
```console
$ docker run -d swarm --flag my --cluser is --awesomer than --everyone elses
```

Also if you are concerned with people not knowing the difference between `ENTRYPOINT` and `CMD` this would be a great example to show if you did something like:
```
ENTRYPOINT ["swarm"]
CMD ["--help"]
```

crosbymichael: This is the right way

```
ENTRYPOINT ["swarm"]
CMD ["--help"]
```

tianon: @jfrazelle "because we should use the best practices for our Dockerfiles"

https://docs.docker.com/articles/dockerfile_best-practices/#entrypoint :wink:

psftw:
OKOK, the Docker Gods have spoken.  I've updated the PR to use ENTRYPOINT/CMD.  :cat:

psftw:
This PR isn't for official image changes any longer, since that is shaping up to be an independent thing, so feel free to merge or close and pick out whatever you like. ;-)

vieux:
@psftw thanks!

PR #340, opened 2015-01-31
hack code to not error on go vet

vieux:


vieux:
@aluzzardi what do you think ?

aluzzardi:
LGTM

Issue #342, opened 2015-02-01
Leaking file descriptors


jhamrick:
It seems as though swarm is leaking file descriptors. Earlier today, I started seeing a bunch of errors like this in the swarm log:

```
2015/02/01 01:20:33 http: Accept error: accept tcp 127.0.0.1:2735: too many open files; retrying in 5ms
2015/02/01 01:20:33 http: Accept error: accept tcp 127.0.0.1:2735: too many open files; retrying in 10ms
2015/02/01 01:20:33 http: Accept error: accept tcp 127.0.0.1:2735: too many open files; retrying in 20ms
2015/02/01 01:20:33 http: Accept error: accept tcp 127.0.0.1:2735: too many open files; retrying in 40ms
2015/02/01 01:20:33 http: Accept error: accept tcp 127.0.0.1:2735: too many open files; retrying in 80ms
2015/02/01 01:20:33 http: Accept error: accept tcp 127.0.0.1:2735: too many open files; retrying in 160ms
2015/02/01 01:20:33 http: Accept error: accept tcp 127.0.0.1:2735: too many open files; retrying in 320ms
```

I only had about 80 containers running, but checking `lsof` revealed that the swarm process had 1035 file descriptors open (which is above the current limit of 1024). It would appear that whenever I stop or start a container, swarm opens a new fd, and then never closes it.

Edit: Also, to clarify, restarting swarm fixed the problem. However, I imagine it will appear again next time swarm reaches >1024 file descriptors.

jhamrick:
I'm not sure if this is fixed, unfortunately, as descriptors still seem to be counting up when containers are run and stopped:

```
root@compmodels-hub-test:~# lsof -a -p $(pidof swarm) | wc -l
14
root@compmodels-hub-test:~# docker --tls -H 127.0.0.1:2376 run hello-world
Hello from Docker.
This message shows that your installation appears to be working correctly.
...
root@compmodels-hub-test:~# lsof -a -p $(pidof swarm) | wc -l
19
root@compmodels-hub-test:~# docker --tls -H 127.0.0.1:2376 run hello-world
Hello from Docker.
This message shows that your installation appears to be working correctly.
...
root@compmodels-hub-test:~# lsof -a -p $(pidof swarm) | wc -l
22
```

The full output of `lsof` shows that the connections are still in the `ESTABLISHED` state:

```
root@compmodels-hub-test:~# lsof -a -p $(pidof swarm)
[...]

PR #343, opened 2015-02-01
Force connection close with token discovery

mota:
Probably not related to #342 but this PR fixes a leak of FDs with the token discovery.

Since http response bodies were not closed, connections stayed in CLOSE_WAIT state.


vieux:
LGTM

Issue #214, opened 2014-12-31
Integrate Mesos into Scheduler API


tnachen:

abronan:
Do we integrate with Mesos directly or through Marathon which already manages containers on top of Mesos? Seems to me that we could easily map Swarm label/constraints with Marathon constraints. Same about your comment on #197 with health checks, tracking, etc. Swarm can access the Marathon API or be notified with the scheduler API thus letting the underlying scheduler deal with health checking mechanisms.

I'm not too familiar with Mesos but it seems that we still need to create a lightweight framework to deal with constraints so that we accept or decline offers from the Mesos Master if those constraints are met or not.

tnachen:
It depends on what swarm is trying to cover and the information that it needs.

Ideally using Marathon will be the simplest approach although swarm will needs another dependency service.

Building a framework in swarm will be the most flexible approach, with the cost of building from scratch, which might not be too bad.

I am trying to avoid having two frameworks (marathon and another) that swarm runs on. It becomes a lot easier to fail and harder to reason since there are two states that needs to be in sync.

abronan:
> "Building a framework in swarm will be the most flexible approach"

> I am trying to avoid having two frameworks (marathon and another) that swarm runs on. It becomes a lot easier to fail and harder to reason since there are two states that needs to be in sync.

:+1: Agree

@vieux Is it possible to extract, adapt and merge core parts of VoltFramework into Swarm then? At least we could get some inspiration from this project after some work with the pluggable Scheduler API :-)

tnachen:
I don't think voltframework integrates with swarm?

vieux:
I also agree that swarm being a framework itself would be the most flexible approach.

No volt doesn't integrate with swarm, but it's a very simple framework written in pure go so it might look like the "swarm framework" in the end.

aluzzardi:
Agreed as well.

davelester:
Since multiple Mesos schedulers (including Marathon, Aurora, and Singularity) can now schedule Docker containers, would it be possible for Swarm to offer a simple integration point with those instead creating a new framework? It's unclear to me why you'd want to create a new framework for this when there are already several options.

vieux:
Is there a way something generic ? To my knowledge, it's not possible, as they all have a different way to communicate (proprietary API / CLI)

Our plan was to do a Scheduler API + implement simple Swarm framework, after, is the Aurora team wants to build and support the Swarm's "aurora backend" that would be great.

[...]

Issue #213, opened 2014-12-31
Pluggable scheduler API


tnachen:
Like to start the discussion around pluggable scheduler API, to make sure it doesn't become impossible to integrate with other schedulers.

I'm coming from the Mesos side, but I'm sure there are others too.

bfirsh:
:+1:

The sooner we define an API the better – both so we don't become tied to our scheduler and so other providers (Kubernetes, Fleet, etc) can write their own drivers.

tnachen:
@vieux @aluzzardi how do we move forward with this? I think we should start implementing the default one under a interface at the very least.
Also it seems like the current scheduler class is only about the placement of a container, but there is also discovery and managing the state of a cluster.
I think all of this should be pluggable under interfaces.

vieux:
@tnachen you're right, right now, scheduler and discovery are separated. Discovery already has it's own interface (consul, etcd, etc...), but not Scheduler.

You could look at this https://github.com/docker/swarm/pull/178 for an attempt at really basic version of the Scheduler interface.

Regarding mesos I feel like Scheduler and Discovery needs to be blended because swarm won't get a list of nodes, but a list of mesos master.


My dream would be `swarm manage --discovery mesos://<zookeeper_url_with_the_mesos_masters>`

vieux:
Actually, after thinking about this more, I think we don't have to mix Scheduler and Discovery.

This will probably the main use case:
`swarm manage --scheduler mesos --discovery zk://<zookeeper_url_with_the_mesos_masters>/path`

But you should be able to use other discovery mechanism
`swarm manage --scheduler mesos --discovery file://list_mesos_master`

@tnachen what do you think ?

tnachen:
I think I need to understand more if discovery info is going to be used by some other commands or components of swarm.

Seems like discovery is simply listing all the swarm agents, does it also plan to list all the containers or find a particular container in the cluster?

But yes, I think discovery can stay as-is.
I'm thinking that the swarm framework can launch custom executors which launches the swarm agent that will be running on each host, and we can let the custom executor talk to the discovery backend out of band from Mesos if we want to utilize swarm discovery features in the future.

btw @vieux, if a swarm manager dies, how does it bootstrap again to get to know all the running containers?

vieux:
discovery gives you a list of ip/port you need to talk to:
with the default scheduler, it's a list of docker daemon, with mesos I think it'll be a list of mesos master.

[...]

PR #347, opened 2015-02-02
Proposal: Swarm plugin system

chanwit:
**Work in-progress**

This PR proposes yet another plugin system for Swarm.
It would be, at least, a step stone for the following issues and PRs.

#144 Plugins
#149 More optimal host selection strategy
#213 Pluggable schedular API
#214 Integrate Mesos into Schedular API
#227 add `balanced` scheduler strategy

Learnt from above, we can infer that everyone would like to see Swarm to be extensible in some ways.
This PR has been heavily developed for a couple of weeks, however there's still a lot of rooms for improvement. So feedback is very welcome :-)

As Go does not support shared libraries (yet), the approach proposed by this PR is to have a plugin running as a separate process, talking to `swarm` via unix sockets or windows pipes.

Swarm will be responsible for:
- Scanning executable files next to it.
- Register plugins. If files contain plugin prefixes, e.g. "swarm-strategy-" for strategy plugins, Swarm will register them as plugins. For example, "swarm-strategy-first" will be a strategy named "first" in the strategy's map. For discovery plugins, the prifix is "swarm-discovery-".
- Start a plugin executable, if required. Communication will be the Go default RPC format (encoding/gob).

The plugin framework will be responsible for:
- Provide developers an easy, Swarm-compatible, way to write plugins.
- An easy way to test plugins

This approach has many advantages as follows:

1. Changes to the Swarm's core is minimized.
2. People'd like to implement different kinds of plugins. For example, some want them to be remote APIs, some don't. This approach obviously frees them to write whatever they want.
3. A local plugin can be simply written, in case it's just a computational.
4. Also, a remote plugin can be written by letting a local plugin call remote services.
5. Potential plugins can later ship with Swarm without modification of the core.
6. This approach of plugin scales well when a number of nodes becomes large, because data is passed locally. It's the resonsibility of a plugin itself to digest input data, and make remote calls. Test has been simulated for a strategy plugin at 30,000 calls, mashalling data of 1,000 nodes. Each call is around 1.2 ms on a 5-year-old machine. See benchmark below.
7. Plugins will be auto-registered and auto-removed upon placing or removing the plugin executables. No extra parameter is required to pass to `swarm`. Just to keep things simple. Plugin reloading at runtime has not been implemented yet. Current, `swarm` is required a restart.
8. Follows the "batteries included, but removable" principle. We can still use Swarm standalone.
9. A plugin can override the existing behaviour using the same name. For example, One can improve the Binpacking and override the current implementation with `swarm-strategy-binpacking`.

**Benchmark**
```sh
$ swarm-plugin-tester strategy:first 30000
Success = 100%
total = 38567 ms
each  = 1.285567 ms
```

An early implementation of the plugin framework can be found here (this PR):

http://github.com/chanwit/swarm/tree/plugin-system

Changes not yet made to Swarm
- Refactor cluster.Node to common.Node so mashalling performance can be improved.

Writing a plugin is simple. Example of plugins can be found here:

http://github.com/chanwit/swarm-strategy-first
http://github.com/chanwit/swarm-strategy-balance
http://github.com/chanwit/swarm-strategy-revbinpack
[TODO] http://github.com/chanwit/swarm-discovery-zk

The plugin tester can be found here:

http://github.com/chanwit/swarm-plugin-tester

Open Questions:
1. Where's a good place for plugins? Current setting is next to swarm binary. Should we introduce plugindir?
2. What kinds of plugin should we support in Swarm, strategy, discovery, filter? You name it.


chanwit:
This PR requires a proper functional tests (#376) to make it happens.

chanwit:
This PR will not be valid any more. I shall close it.

PR #348, opened 2015-02-02
update swarm manage requirements

### LINKED ISSUE #346: [Question] How do I use or configure `swarm manage` in multi-region?
According to the README.md, if I were right, `swarm manage` should be able to be used in other networks which is not the same as the network of `swarm join`.

    # on each of your nodes, start the swarm agent
    # <node_ip> doesn't have to be public (eg. 192.168.0.X),
    #  as long as the other nodes can reach it, it is fine.
    $ swarm join --addr=<node_ip:2375> token://<cluster_id>

    # start the manager on any machine or your laptop
    $ swarm manage -H tcp://<swarm_ip:swarm_port> token://<cluster_id>

But, after some testing, I realized that `swarm manage`(the manager) needs to be able to reach it. If I want to join via private networks(192.168.0.X) and also I want to multi region(us-east-1,us-wast-2 ...) too, how do I use and configure `swarm manage`(the manager)?

Thanks for reading. :-)

vieux:
Close #346

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ $ swarm create
 
 # on each of your nodes, start the swarm agent
 #  <node_ip> doesn't have to be public (eg. 192.168.0.X),
-#  as long as the other nodes can reach it, it is fine.
+#  as long as the manager and the docker cli can reach it, it is fine.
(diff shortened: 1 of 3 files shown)


aluzzardi:
LGTM

PR #337, opened 2015-01-30
use docker pull <node>...<image>

vieux:
I'm opening this PR even though it will likely never get merged :D

@aluzzardi I can't find the PR you suggested using docker pull node@image, but I thougt about that also, and it's not possible, because the docker cli checks that the repository should be valid.

As an example in this PR I used `...`

It's working great :D

PR #349, opened 2015-02-02
Add support for pull on all hosts (no change in docker run)

vieux:
Right now, we have an issue because when we do a docker run, the scheduler will select a node.
Then, if this node doesn't have the image already pulled, swarm will pull the image automatically.
We do that because there is no way to tell to the client to pull the missing image on a particular node.
One issue is: it doesn't work with private images, because swarm doesn't have the client credential.

With this PR, you are able to do a `docker pull` manually to pull an image on all the nodes, this works for private images.
This PR also add a basic implementation of `docker rmi` and `docker inspect` for images

There is no change in the way docker run works.

```bash
docker pull rethinkdb
ubuntu-2: Pulling downloaded... 
fedora-1: Pulling downloaded... 
ubuntu-1: Pulling downloaded... 
```
then 
```bash
docker pull rethinkdb
ubuntu-2: Pulling downloaded...
fedora-1: Pulling downloaded...
ubuntu-1: Pulling downloaded... : rethinkdb:latest
```
then 
```bash
docker pull rethinkdb
ubuntu-2: Pulling downloaded... : rethinkdb:latest
fedora-1: Pulling downloaded... : rethinkdb:latest
ubuntu-1: Pulling downloaded... : rethinkdb:latest
```

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -11,5 +11,7 @@ type Cluster interface {
 	Containers() []*Container
 	Container(IdOrName string) *Container
 
+	Pull(name string, begin, end func(string))
+
 	Info() [][2]string
 }
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -3,15 +3,31 @@ package cluster
 import "github.com/samalba/dockerclient"
 
 type Cluster interface {
+	// Create a container
 	CreateContainer(config *dockerclient.ContainerConfig, name string) (*Container, error)
+
+	// Remove a container
 	RemoveContainer(container *Container, force bool) error
 
+	// Return all images
 	Images() []*Image
+
+	// Return one image matching `IdOrName`
 	Image(IdOrName string) *Image
+
+	// Return all containers
 	Containers() []*Container
+
+	// Return container the matching `IdOrName`
 	Container(IdOrName string) *Container
 
-	Pull(name string, begin, end func(string))
+	// Pull images
+	// `callback` can be called multiple time
+	//  `what` is what is being pulled
+	//  `status` is the current status, like "", "in progress" or "downloaded
+	Pull(name string, callback func(what, status string))
 
+	// Return some info about the cluster, like nb or containers / images
+	// It is pretty open, so the implementation decides what to return.
 	Info() [][2]string
 }
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -186,14 +186,18 @@ func (c *Cluster) Image(IdOrName string) *cluster.Image {
 	return nil
 }
 
-func (c *Cluster) Pull(name string, begin, end func(string)) {
+func (c *Cluster) Pull(name string, callback func(what, status string)) {
 	size := len(c.nodes)
 	done := make(chan bool, size)
 	for _, n := range c.nodes {
 		go func(nn *node) {
-			begin(nn.Name())
-			nn.Pull(name)
-			end(nn.Name())
+			if callback != nil {
+				callback(nn.Name(), "")
+			}
+			nn.pull(name)
+			if callback != nil {
+				callback(nn.Name(), "downloaded")
+			}
 			done <- true
 		}(n)
 	}
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -187,10 +187,18 @@ func (c *Cluster) Image(IdOrName string) *cluster.Image {
 }
 
 func (c *Cluster) Pull(name string, begin, end func(string)) {
-	for _, node := range c.nodes {
-		begin(node.Name())
-		node.Pull(name)
-		end(node.Name())
+	size := len(c.nodes)
+	done := make(chan bool, size)
+	for _, n := range c.nodes {
+		go func(nn *node) {
+			begin(nn.Name())
+			nn.Pull(name)
+			end(nn.Name())
+			done <- true
+		}(n)
+	}
+	for i := 0; i < size; i++ {
+		<-done
 	}
 }
 
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -186,6 +186,14 @@ func (c *Cluster) Image(IdOrName string) *cluster.Image {
 	return nil
 }
 
+func (c *Cluster) Pull(name string, begin, end func(string)) {
+	for _, node := range c.nodes {
+		begin(node.Name())
+		node.Pull(name)
+		end(node.Name())
+	}
+}
+
 // Containers returns all the containers in the cluster.
 func (c *Cluster) Containers() []*cluster.Container {
 	c.RLock()
--- a/api/README.md
+++ b/api/README.md
@@ -18,7 +18,7 @@ GET "/containers/{name:.*}/attach/ws"
 
 POST "/commit"
 POST "/build"
-POST "/images/create"
+POST "/images/create" (pull implemented)
 POST "/images/load"
 POST "/images/{name:.*}/push"
 POST "/images/{name:.*}/tag"
--- a/api/api.go
+++ b/api/api.go
@@ -263,13 +263,14 @@ func postImagesCreate(c *context, w http.ResponseWriter, r *http.Request) {
 		if tag := r.Form.Get("tag"); tag != "" {
 			image += ":" + tag
 		}
-		begin := func(name string) {
-			fmt.Fprintf(wf, "{%q:%q,%q:\"Pulling %s...\",%q:{}}", "id", name, "status", image, "progressDetail")
+		callback := func(what, status string) {
+			if status == "" {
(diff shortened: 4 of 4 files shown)


chanwit:
I think its the best solution for the time being.

aluzzardi:
A couple of points:

- Given that `DeployContainer` doesn't trigger a pull automatically anymore, it means that re-scheduling will fail. Perhaps the pull flag should be exposed there and set to `false` by the API (and we'll let the scheduler set it to `true` when re-scheduling).

- What happens when the cluster grows? Let's say you did a `docker run redis`. It will pull `redis` on every hosts. Now a new host comes up and we do a `docker run redis` again, which happens to end up on this new machine. The CLI will issue a `pull`, resulting in the update of the `redis` image on every single host, which may not be something the user expected?

chanwit:
> What happens at scale? 
Should we intercept `docker run` and explicitly check and let `docker` pull missing images?
With this, it may allow pulling only for the node selected to run.

aluzzardi:
Unfortunately, `docker run` is entirely on the CLI side and issues a `create` (gets a 404 since the image doesn't exist) / `pull` / `create` / `start`

vieux:
* I agree about the flag, it should set by the API.

* What happens when the cluster grows?
Here is an idea that might be crazy, instead of "remembering the pulls":
When a new node joins the cluster, the master could have a rule: it scans every node it already new and for each image that is on every node, pull it a new one.

* What happens at scale?
I don't think it would be an issue, as you said it would speed up the rebalancing.

bfirsh:
+1 we need this! yay!

My concerns with the implementation are similar to @aluzzardi's. Also – what if you commit an image? How does that image get propagated to other nodes in the cluster?

Perhaps instead of each node having all of the images, a swarm could be attached to a registry which represented the images in a swarm. When a container is created, the assigned node could pull the image from the registry. When you commit, the image gets pushed to the registry.

Also – another crazy idea: what if Engines could push and pull to each other? Perhaps the images in a swarm could be an aggregate of the images on each of the nodes. When a node in the cluster needed an image that was on another node, it could pull from that node. /cc @stevvooe @dmp42 who have had similar ideas

stevvooe:
@bfirsh We can break this down using the following rules:

1. The "swarm" has a global image store.
2. The current location of any image should never matter to the operator.
3. Push, commit and pull client commands from swarm operate on this store.

From this, it follows that using the run command _may_ result in an "internal pull". If an image cannot be resolved in the local cluster (read "swarm registry"), the run command would fail. I'm not sure if runtime image resolution exists currently, but a pluggable registry for internal clusters would make this possible.

As a side note, pulling all images to all machines is a huge problem for large deployments. Imagine the read amplification for a small deployment on a large cluster, where the application is deployment to a small percentage of nodes.

allisonvoll:
> As a side note, pulling all images to all machines is a huge problem for large deployments. Imagine the read amplification for a small deployment on a large cluster, where the application is deployment to a small percentage of nodes.

maybe the option to apply the filter in `docker pull` (as is possible scheduler `docker run`)

vieux:
The thing is, we are limited by the docker cli, and right now the cli
doesn't accept any flag.

We also tried `docker pull redis` pulls on every machine and `docker pull
redis@node1` pulls only on node1, but we cannot use symbols

See https://github.com/docker/swarm/pull/337 for reference.

On Fri, Feb 20, 2015 at 11:35 AM, allisonvoll <notifications@github.com>
wrote:

> As a side note, pulling all images to all machines is a huge problem for
> large deployments. Imagine the read amplification for a small deployment on
> a large cluster, where the application is deployment to a small percentage
> of nodes.
>
> maybe the option to apply the filter in docker pull (as is possible
> scheduler docker run)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/docker/swarm/pull/349#issuecomment-75304604>.
>



-- 
Victor Vieux
Docker Core Engineer
vieux@docker.com

kgoudeaux:
What about maintaining a mapping of images to filters? On a create call that requests image X with filter constraint=ssd store the X->[filter1, filter2] relationship. Any subsequent pull for X (e.g. from the client that just got a 404 on create) will trigger the pull on all hosts that match the cached filters for image X. If there are no associated filters, pull everywhere. For new nodes, walk the data in reverse to find images that match the node.

bacongobbler:
+1 for the ability to pre-fetch images onto every swarm host!

> What happens at scale? We have our target of ~100 nodes. Making docker pull pull on every node makes sense, the user explicitly asked that. However, pulling a hundred times on every docker run might not be something we want. Or not. Perhaps this is great as we can re-schedule containers on the fly without having to pull images, resulting in a smaller downtime. Thoughts?

I think this is actually a *great* thing that the global pull occurs on a `docker run`. For our use case, we would like a speedy recovery in the instance that a host dies rather than having to wait however long it takes to pull the image onto the new host.

> Here is an idea that might be crazy, instead of "remembering the pulls":
> When a new node joins the cluster, the master could have a rule: it scans every node it already new and for each image that is on every node, pull it a new one.

:+1: for "smart" operations on node joins. I can see that we could add "smart" node re-balancing scheduler strategies with this workflow in the future, too :)

smothiki:
Victor has announced that there is a plan for Swarm manager to be  HA  and leader election through Raft . If there is a way for swarm manager to store or cache the images that are being pulled on a node and while creating a container on a different node other than the one that has the image , then the node can simply pull from the swarm manager --> just requires to edit the image or json template for the pull request from manager to node .

stevvooe:
@vieux I don't have a solution for the CLI aspect, other than tying image fetch to a resource offer framework.

I would respectfully request that you avoid the `docker pull redis@node1` since it will collide with the proposal in docker/docker#10740.

vieux:
@stevvooe thanks for pointing that out. In any case, we are avoiding it, because it is not possible with today's CLI.

@gabrtv @aluzzardi I updated the PR to remove the pull on docker run. To use private images, you'll have to docker pull (on every node) manually.

vieux:
I updated the title and description to make things more accurate.

/cc @aluzzardi

cluster/swarm/cluster.go
```diff
@@ -186,6 +186,14 @@ func (c *Cluster) Image(IdOrName string) *cluster.Image {
 	return nil
 }
 
+func (c *Cluster) Pull(name string, begin, end func(string)) {
+	for _, node := range c.nodes {
```
aluzzardi: This should probably happen in parallel

cluster/cluster.go
```diff
@@ -11,5 +11,7 @@ type Cluster interface {
 	Containers() []*Container
 	Container(IdOrName string) *Container
 
+	Pull(name string, begin, end func(string))
```
aluzzardi: Isn't begin/end leaking the fact that we have `Node`s?

cluster/swarm/cluster.go
vieux: I'll try to find a way, last time I checked, it was messing the docker cli UI completly

cluster/cluster.go
vieux: I didn't get it.

Btw I don't like those begin/end, if you have a better idea I'd love to hear it.
(I think giving the `io.Writer` to pull is a but too much, don't you think ?)

cluster/swarm/cluster.go
vieux: Ah since, I changed, the output, it works!

vieux:
@aluzzardi please take another look I updated with your suggestions

cluster/swarm/cluster.go
```diff
@@ -186,6 +186,22 @@ func (c *Cluster) Image(IdOrName string) *cluster.Image {
 	return nil
 }
 
+func (c *Cluster) Pull(name string, callback func(string, string)) {
+	size := len(c.nodes)
+	done := make(chan bool, size)
+	for _, n := range c.nodes {
+		go func(nn *node) {
+			callback(nn.Name(), "")
```
aluzzardi: Could you ensure the callback is not null before invoking it?

cluster/swarm/cluster.go
```diff
@@ -186,6 +186,22 @@ func (c *Cluster) Image(IdOrName string) *cluster.Image {
 	return nil
 }
 
+func (c *Cluster) Pull(name string, callback func(string, string)) {
```
aluzzardi: Could you add a comment (either here or in the interface) and explain the callback (what are the arguments, it may be called multiple times, etc)?

vieux: I will

cluster/swarm/cluster.go
vieux: I will even though I don't think is is necessary. The function isn't called by the cluster driver implem, but by the swarm api. So we won't call it with nil.

vieux:
@aluzzardi PTAL

huslage:
Wouldn't adding support for `docker login` on each node solve the credentials issue and allow `docker run` to pull images again?

jippi:
:+1:

PR #353, opened 2015-02-03
add gitcommit to version

vieux:


aluzzardi:
Fixes #331

vieux:
With the current state of the PR, the version will be empty when you `go build`

vieux:
@aluzzardi should we default to `master` as you suggested ?

aluzzardi:
LGTM

PR #364, opened 2015-02-04
replace discovery.Node by discovery.Entry + small refactor

vieux:
Today I find very confusing that we have 2 structs names Node, one in the `discovery` (host/port) and one in `cluster` (actual docker node).

Today those two structs are kind of related because a `discovery.Node` is converted to a `cluster.Node`
This will become even more confusing in the future, because the `discovery.Node` could be a Mesos master or a Swarm manager instead of a docker node.

That's why I propose to rename `discovery.Node` to `discovery.Machine`

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -99,11 +99,11 @@ func (c *Cluster) AddNode(n *Node) error {
 	return n.Events(c)
 }
 
-func (c *Cluster) UpdateNodes(nodes []*discovery.Node) {
-	for _, addr := range nodes {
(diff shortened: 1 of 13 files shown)


aluzzardi:
LGTM - Not sure regarding `discovery.Machine` vs `discovery.Entry`, I'll let you decide on that, either way is fine for me.

vieux:
@aluzzardi I used `Entry`, thanks for the suggestion.

I also refactored the code a bit: `+91  −123`

aluzzardi:
LGTM

PR #368, opened 2015-02-04
do not replace 0.0.0.0 by 127.X

vieux:
@metcalfc this should fix your issue


metcalfc:
Tested this. Its close. But instead of `0.0.0.0` we need it to resolve to the IP of the hostname.

vieux:
@metcalfc we are resolving the hostname to get the IP here: https://github.com/docker/swarm/blob/master/cluster/node.go#L64

With your setup, the IP resolved is 127.0.0.1, which is worst than 0.0.0.0 so we don't replace it.

PR #369, opened 2015-02-05
Fix race condition on exec

mota:
Since swarm returns proxified http response before refreshing the target container, the docker client tends to request the start of the newly created exec before swarm starts to update its internal states.
This behavior results in a unfound execID upon start, though the exec exists.

This PR introduces a asynchronous api proxy that allows to seek for a passed callback before feeding the http response to the client. In the meantime, it fixes the `exec` behavior.

### DIFF
--- a/api/utils.go
+++ b/api/utils.go
@@ -50,7 +50,7 @@ func copyHeader(dst, src http.Header) {
 	}
 }
 
-func proxy(tlsConfig *tls.Config, addr string, w http.ResponseWriter, r *http.Request) error {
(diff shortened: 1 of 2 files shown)


api/utils.go
```diff
@@ -50,7 +50,7 @@ func copyHeader(dst, src http.Header) {
 	}
 }
 
-func proxy(tlsConfig *tls.Config, addr string, w http.ResponseWriter, r *http.Request) error {
+func proxy_async(tlsConfig *tls.Config, addr string, w http.ResponseWriter, r *http.Request, callback func(*http.Response)) error {
```
vieux: can you use `proxyAsync`

mota: Sure thing, damn Python habits...

vieux:
LGTM

PR #375, opened 2015-02-06
Add cross-compile back to Travis and Go 1.4

tianon:
I discussed this with @vieux in `#docker-dev`, and he mentioned that I should open a PR so we can test the speed. :+1:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,13 @@
 language: go
 
 go:
-  - 1.3
-#  - 1.4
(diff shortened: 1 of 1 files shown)


vieux:
@tianon tests fail

tianon:
Whoo DCO! :smile:

tianon:
Looks like it needs to exclude `go vet` on not host platforms too :innocent:

tianon:
So I guess the benefits are getting smaller and smaller now, reintroducing complexity in `.travis.yml`... :disappointed:

PR #381, opened 2015-02-10
closeIdleConnections to prevent leaks with https

### LINKED ISSUE #342: Leaking file descriptors

vieux:
Fix #342 

This is a solution that works for me and doesn't requite a change in `dockerclient`

If it's still not enough, I have a step2: instead of creating a client for each proxy, we use the one from `dockerclient`

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -180,6 +180,11 @@ func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		httpError(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	// cleanup
+	defer resp.Body.Close()
+	defer closeIdleConnections(client)
+
 	data, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		httpError(w, err.Error(), http.StatusInternalServerError)
--- a/api/utils.go
+++ b/api/utils.go
@@ -50,6 +50,13 @@ func copyHeader(dst, src http.Header) {
 	}
 }
 
+// prevents leak with https
+func closeIdleConnections(client *http.Client) {
+	if tr, ok := client.Transport.(*http.Transport); ok {
+		tr.CloseIdleConnections()
+	}
+}
+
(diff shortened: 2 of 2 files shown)


vieux:
ping @aluzzardi @jhamrick please test this one! Thanks

jhamrick:
Thanks! I am away this week so may not be able to test for a few days but I will definitely test this as soon as I get back, if not sooner.

aluzzardi:
LGTM

vieux:
@jhamrick we merged it to have it in the rc4, but you are more than welcome to test, we will reopen #342 if it's not completely fixed for you.

PR #386, opened 2015-02-10
Fixes #372 Consul discovery backend should support multiple ip

### LINKED ISSUE #372: [ENH] consul discovery backend, add support for multiple consul nodes
With a consul "cluster" we do not know at join time which of the consul nodes are healthy. Should it be possible to specify more than one consul ip address so when joining it take the first available and healthy node ?

david-guenault:
This is my first try with go and i'm not sure i'm doing it the right way.
According to documentation consul discovery backend should support multiple addresses in discovery uri. Now it work. Just added a loop to iterate each address until we found one available. Since consul api does not provide a way to test if a node is available nor the hability to give multiple host in one connection (like etcd does), i use net/http and io/ioutil to achieve the same goal.

Signed-off-by: David GUENAULT <david.guenault@gmail.com>

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -99,11 +99,11 @@ func (c *Cluster) AddNode(n *Node) error {
 	return n.Events(c)
 }
 
-func (c *Cluster) UpdateNodes(nodes []*discovery.Node) {
-	for _, addr := range nodes {
-		go func(node *discovery.Node) {
-			if c.Node(node.String()) == nil {
-				n := NewNode(node.String(), c.overcommitRatio)
+func (c *Cluster) UpdateNodes(entries []*discovery.Entry) {
+	for _, entry := range entries {
+		go func(m *discovery.Entry) {
+			if c.Node(m.String()) == nil {
+				n := NewNode(m.String(), c.overcommitRatio)
 				if err := n.Connect(c.tlsConfig); err != nil {
 					log.Error(err)
 					return
@@ -113,7 +113,7 @@ func (c *Cluster) UpdateNodes(nodes []*discovery.Node) {
 					return
 				}
 			}
-		}(addr)
+		}(entry)
 	}
 }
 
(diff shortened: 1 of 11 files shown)


david-guenault:
Sorry for the multiple travis fails. Got to learn more about vet and coding style. Now everything seems to be working fine.

aluzzardi:
Can you melt those commits together?

You can do that by doing:
```
git rebase -i `git merge-base master 372-Add_support_to_multiple_consul_nodes`
# change "pick" to "fixup" in every commit except the first one
git push -f
```

That way, this PR will appear as a single commit and that commit will include the content of all the 4 commits (so we keep a clean history, and every commit is buildable and passes tests)

vieux:
@JeffChien can you review ? Thanks!

vieux:
@david-guenault I think you should try to do like etcd: instead of picking a healthy machine at the beginning, you should keep all the machine in memory, and when you get an error, try another machine.

See https://github.com/coreos/go-etcd/blob/master/etcd/cluster.go for inspiration, on each failure() a new machine is picked.

david-guenault:
@aluzzardi thx for the tip :-)

@vieux thx for the help :-) gona check this.

JeffChien:
agree with vieux. I prefer use the go api instead of doing http request, maybe you could maintain a client pool, if the old connection fail, pick a new one from the pool.

david-guenault:
@vieux @JeffChien  I think i have to drop this PR since there are a lot of changes in my code. But here is what i think is good

Initialize
=> We give a list of consul nodes (at least 2 but could be more)
=> Discovery module try each given nodes until we find one available
=> Get nodes list from this consul available node (this is the real pool)

Fetch/Register/WaitForChange
=> If error Pick a new node from the pool and return with error (the next time it will work)

What i'm not sure is when we need to ask consul cluster to update pool list (since we can add/remove consul nodes at running time).

What do you think ?

(sorry beeing so long but i'm learning go ...)

PR #388, opened 2015-02-10
Dependency Filter: co-schedule dependent containers on the same node.

### LINKED ISSUE #251: How to deal with --link --volume_from network container etc.
In a word, most of the cases in real world requires relationship between Docker containers like --link --volume-from --net=container, that means, at least, swarm should be able to schedule those containers to one host.

But I did not see this kind of strategy in swarm currently, is it under developing or swarm has other approach?

aluzzardi:
Supported dependencies:
- Shared volumes: `--volumes-from=dependency`
- Links: `--link=dependency:alias`
- Shared network stack: `--net=container:dependency`

Fix #251

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"crypto/tls"
 	"errors"
-	"strings"
 	"sync"
 
 	log "github.com/Sirupsen/logrus"
@@ -119,8 +118,8 @@ func (c *Cluster) UpdateNodes(entries []*discovery.Entry) {
 
 // Containers returns all the containers in the cluster.
 func (c *Cluster) Containers() []*Container {
-	c.Lock()
-	defer c.Unlock()
+	c.RLock()
+	defer c.RUnlock()
 
 	out := []*Container{}
 	for _, n := range c.nodes {
@@ -139,17 +138,13 @@ func (c *Cluster) Container(IdOrName string) *Container {
 	if len(IdOrName) == 0 {
 		return nil
 	}
-	for _, container := range c.Containers() {
-		// Match ID prefix.
-		if strings.HasPrefix(container.Id, IdOrName) {
-			return container
-		}
 
-		// Match name, /name or engine/name.
-		for _, name := range container.Names {
-			if name == IdOrName || name == "/"+IdOrName || container.Node.ID+name == IdOrName || container.Node.Name+name == IdOrName {
-				return container
-			}
+	c.RLock()
+	defer c.RUnlock()
+
+	for _, n := range c.nodes {
+		if container := n.Container(IdOrName); container != nil {
+			return container
 		}
 	}
 
--- a/cluster/node.go
+++ b/cluster/node.go
@@ -394,6 +394,7 @@ func (n *Node) Events(h EventHandler) error {
 	return nil
 }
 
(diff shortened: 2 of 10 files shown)


vieux:
LGTM

aanand:
Amazing. Will try this out tomorrow.

vieux:
@aanand thanks!

bacongobbler:
big :+1: from me! Definitely something required to add swarm as a supported scheduler for Deis :)

aanand:
:+1: This works great.

vieux:
@aanand thanks for the feedback!

anentropic:
Do I understand correctly that the current recommended workaround is to use this pattern: https://docs.docker.com/engine/articles/ambassador_pattern_linking/

...to break the `link` dependency between containers in my docker-compose.yml so that they can be scheduled to different nodes?

abronan:
@anentropic The current way to link containers between different nodes is to use the `overlay` networking feature of docker `1.9`/`1.10` and use the `docker network` command.

Please refer to the [documentation](https://docs.docker.com/engine/userguide/networking/get-started-overlay/) if you want more details on how to setup multi-host networking and get it working with Docker Compose.

After that you'll be able to use a single Compose file to describe your stack and all the containers are created on the same overlay across different hosts, thus could see and discover each other without using `--link` or the `ambassador pattern`.

anentropic:
Thanks!

Shouldn't this be the default though?  How am I supposed to know I need a specialist feature "multi-host networking overlay" in order for things to work as expected?

It's confusing because there's a bunch of places like https://github.com/docker/compose/blob/master/SWARM.md which still say cross-node links are not possible.

anentropic:
for anyone using docker-compose, you don't need to manually create an overlay network as described in the docs linked above

instead upgrade to compose 1.6 and see the new docs here, where network stuff is configured from the compose file (...this is starting to make sense!):
https://github.com/docker/compose/blob/1.6.0-rc1/docs/networking.md
(and here https://github.com/docker/compose/blob/1.6.0-rc1/docs/compose-file.md)

bfirsh:
@anentropic Recommended solution is, as @abronan says, to use the new Compose file and not specify any links. All Compose containers will be scheduled onto the same overlay network which spans multiple Swarm nodes.

@abronan Docker 1.10 raises an interesting new issue now though – it is possible to use links in an overlay network, so containers probably shouldn't be co-scheduled if they're using links and on an overlay network. Shall I raise a new issue for this?

aanand:
> if they're using links and on an overlay network

@mavenugo what's the precise logic for whether `--link` means an alias or an old-style link? Is it an old-style link if the container is connected *solely* to the `bridge` network?

mavenugo:
@aanand 

> what's the precise logic for whether --link means an alias or an old-style link? Is it an old-style link if the container is connected solely to the bridge network?

That's correct. To be more specific. It is an old-style link only  if the container is connected to the *default* bridge network (`docker0`)

aanand:
What if it's connected to both that network and a custom network?

mavenugo:
Both `--link` and `--alias` are network scoped. (with `--link` further scoped for a container).
If a container is connected to both the default and custom network, then the `--link` will work on each of these networks just as intended. The only exception to the rule is that, if the user prefers the `ENV` variables, then the container must be started with default bridge network in `--net`.

PR #395, opened 2015-02-12
Correct IRC channel in the readme

stervo4ka:
I joined #swarm and it was empty and I was like ":((((" and then I had to google the right one.


stervo4ka:
Just ignore it.

vieux:
Why you just need to sign it, are you creating a new PR ?

PR #403, opened 2015-02-18
Fix potential null dereference in token discovery

alexzorin:
Noticed that this caused a panic during Docker node shutdown/reboot process.


alexzorin:
duplicated by #407 I guess.

vieux:
@alexzorin sorry I didn't saw your PR

PR #430, opened 2015-02-27
Add change log

bfirsh:
For future use!

Probably also worth updating the release page with the latest tag: https://github.com/docker/swarm/releases

For Machine I removed all of the descriptions from the older releases to make the main release stand out: https://github.com/docker/machine/releases

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,0 +1,17 @@
+Change log
+==========
+
+0.1.0 (2015-02-26)
+------------------
+
+Initial beta release.
+
+* Small static go binary with no dependencies.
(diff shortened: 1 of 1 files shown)


aluzzardi:
We chose to delete the releases instead - we are not shipping binaries anymore (docker image instead).

Regarding the ChangeLog - I'm not sure about that. We are working on having the roadmap / milestones in the wiki, that information will be redundant.

@vieux ?

vieux:
@aluzzardi I agree I would close it in favor of the  roadmap / milestones in the wiki

PR #439, opened 2015-03-02
cleanup: SwarmCluster -> Cluster && Node to node

vieux:


vieux:
ping @tnachen what do you think about the new method in the interface ?

tnachen:
Hey there, what is the method for? I can't tell from your change

vieux:
It is used only in one specific place for now, on `docker exec`

Same as `docker run`, `docker exec` in in fact `start + attach` so, after the `start` we need the refresh the container before the client does the attach. That's why we explicitly call `RefreshContainer` in `api.go`

aluzzardi:
I'm actually concerned about this. Nodes shouldn't leak to the API at all.

They are an internal thing of a cluster and only Swarm and Mesos with a Swarm framework will ever use them (for instance, if we have a marathon backend, they shouldn't deal with the nodes at all)

vieux:
@aluzzardi You are right, thanks to #441 `api.go` doesn't use node anymore, I'll update this PR right away

vieux:
@aluzzardi PTAL

aluzzardi:
:+1:

PR #441, opened 2015-03-03
remove proxyContainerAndForceRefresh

vieux:
Right now we have a "special" proxy for docker exec, because we need to do a force refresh in the before returning the result, do get the execID.

This PRs remove this `proxyContainerAndForceRefresh` to get the execID from return of the `/exec`.

Then we insert the execID to the container. Even though this is not saved in the store, it's fine, will be reflected in the store on the next refresh (a few seconds later).

Here are all the advantages of this PR:
* Removes a cast to `swarm.Node` because we were breaking the Node Interface.
* Exec much faster.
* Better error message.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -15,7 +15,6 @@ import (
 	log "github.com/Sirupsen/logrus"
 	dockerfilters "github.com/docker/docker/pkg/parsers/filters"
 	"github.com/docker/swarm/cluster"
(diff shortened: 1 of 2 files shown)


tnachen:
Can't speak much of the code, but I like the new approach :)

aluzzardi:
LGTM

PR #442, opened 2015-03-03
Cleanup: Make swarm Node and Cluster functions private

vieux:
ping @aluzzardi


aluzzardi:
nit: Could the commit message be something like "Cleanup: Make swarm Node and Cluster functions private"? Those messages are everything we'll have in the changelog.

vieux:
@aluzzardi done

PR #400, opened 2015-02-15
Proposal: Functional test POC for version and gitcommit

chanwit:
As @aluzzardi suggested in #376 that Swarm requires functional tests, this PR serves as a POC of doing it automatically using a VirtualBox-based framework from http://github.com/chanwit/marine.

To enable functional testing, follow steps below:

```
$ marine prepare -i=files/ubuntu-14.10-server-amd64.ova -o=files/ubuntu-0.ova -f
INFO[0000] Removed "ubuntu"
INFO[0019] Imported "ubuntu"
INFO[0020] Modified nic2 for "ubuntu"
INFO[0021] Started "ubuntu"
INFO[0041] VM "ubuntu" ready to connect
INFO[0041] Sudo: "wget -qO- --no-check-certificate https://get.docker.com/ | bash"
INFO[0245] Sudo: "service docker start"
INFO[0246] Sudo: "docker pull golang:1.3"
INFO[0365] Stopping "ubuntu"
INFO[0380] VM "ubuntu" is now poweroff
INFO[0380] Snapshot "ubuntu/origin" taken
INFO[0500] Exported "ubuntu" to files/ubuntu-0.ova
INFO[0502] Removed "ubuntu"

$ export SWARM_FUNC_TEST_IMAGE=files/ubuntu-0.ova

$ SWARM_FUNC_TEST=true go test -v github.com/docker/swarm/functest
=== RUN TestVersionCommitId
INFO[0046] Imported "ubuntu"
INFO[0047] Modified nic2 for "ubuntu"
INFO[0047] Snapshot "ubuntu/origin" taken
INFO[0048] Modified nic2 for "box001"
INFO[0048] Clone: Machine has been successfully cloned as "box001"
INFO[0048] Started "box001"
INFO[0068] VM "box001" ready to connect
INFO[0068] Sudo: "git clone --depth 1 http://github.com/docker/swarm"
INFO[0084] Sudo: "cd swarm* && docker build -t swarm:build ."
INFO[0094] Sudo: "docker run --rm -t swarm:build --version"
INFO[0094] Stopping "box001"
INFO[0107] VM "box001" is now poweroff
INFO[0108] Removed "box001"
--- PASS: TestVersionCommitId (108.02 seconds)
PASS
ok      github.com/docker/swarm/functest        108.027s
```

Anyway, codes of functional tests do not affect Swarm itself directly, so it can totally be optional.
/cc @vieux

Issue #376, opened 2015-02-06
Swarm needs integration tests


aluzzardi:
Right now, we only rely on unit tests to ensure everything is working properly but that's not enough.

For instance, TLS was recently broken (https://github.com/docker/swarm/issues/341) and we didn't notice until a user reported the issue.

Integrations test should set up a real cluster of engines and daemons and test things such as:
- TLS/PKI (`swarm<->node` and `cli<->swarm`)
- End-to-end features (affinity, constraints, ...)
- API endpoints (especially the proxied one)
- Run against different versions of the engine and make sure we are compliant (both older versions and new versions).

This would also help validate that upcoming Docker releases are Swarm compatible.

chanwit:
> Run against different versions of the engine and make sure we are compliant (both older versions and new versions).

Boot2docker does not seem to be a choice then.

chanwit:
@aluzzardi here's a quick and dirty framework to test cluster on my own. 

https://github.com/chanwit/marine

Just in case if you're interested.

chanwit:
As framework in #448 is working great now, scheduling strategies need an addition set of integration tests too.
  * Spread in #458 
  * Binpack tweaks in #451

aluzzardi:
Replaced by specific integration test issues (such as #492).

PR #443, opened 2015-03-03
Proposal: PoC for Bats integration tests

chanwit:
Swarm requires integration tests (#376) and @aluzzardi suggested in #400 that Swarm should use `Bats` to run this kind of tests. This PR follows Docker Machine's way of testing (docker/machine#615). It proposes a PoC to perform integration tests for Swarm.

There are 2 tests under `test/integration` directory:
  1. A test to match version and gitcommit.
  2. A test that uses `timeout` to check if Swarm fail to connect to a non-existed node.

To run these tests:
  1. Compile swarm and make sure that there's the binary in `$GOBIN`.
  2. Run `bats test/integration`

/cc @vieux 

### DIFF
--- a/test/integration/manage-non-available-node.bats
+++ b/test/integration/manage-non-available-node.bats
@@ -1,0 +1,15 @@
+#!/usr/bin/env bats
+
+load vars
+
+@test "managing non-available node should be timed out" {
+
+	#
+	# timeout does not accept a bash function, so hard-coded path is required
+	#
+	run timeout 15s $GOBIN/swarm manage -H 127.0.0.1:2375 nodes://8.8.8.8:2375
+
+	[ "$status" -ne 0 ]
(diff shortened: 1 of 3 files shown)


bacongobbler:
should the directory be renamed to `integration-cli`(similar to docker) or `test/integration`? That seems to be the common directory structure, though really it's a matter of whether the dress is white and gold or blue and black.

chanwit:
@bacongobbler Thank you for your comments. I just follow how Docker Machine does for its testing directory. I am fine to change it to any better name.

aluzzardi:
Thanks @chanwit, this is really awesome!

I would probably do `test/integration`, I imagine eventually we'd have `test/stress` and so on.

chanwit:
done renaming to `test/integration`.

aluzzardi:
Thanks!

PR #445, opened 2015-03-04
dealing with agent disconnection

hoonmin:
Hi.
Swarm seems to have two heart-beater: discovery watcher and node refresher.
- The node refresher has a role to refresh internal storage and modify the node's health.
- And the discovery watcher watches discovery's directory then add new nodes to the swarm cluster.

I think the discovery watcher should inform us of the agent disconnections and check docker connections right away.
- We can check the node's status before the stateRefreshPeriod(30s).
- Operators need to know about swarm-agent's status at least in form of a log.

This PR includes very basic features:
- figures out removed nodes from the discovery watch entries.
- logs them.
- refresh the node's storage to check the healthiness only if the node was healthy.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -114,6 +114,27 @@ func (c *Cluster) RemoveContainer(container *cluster.Container, force bool) erro
 
 // Entries are Docker Nodes
 func (c *Cluster) newEntries(entries []*discovery.Entry) {
+	var removed []*node
+SKIP:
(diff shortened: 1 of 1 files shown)


vieux:
@aluzzardi can you take a look ?

aluzzardi:
Thanks, @hoonmin. However, I don't quite understand why we'd trigger a refresh for nodes that disappeared. Could you please explain further?

aluzzardi:
The underlying problem has been fixed by #811

PR #448, opened 2015-03-06
Integration tests: Initial framework

aluzzardi:
This is a continuation of the integration tests, related to #376


chanwit:
This is fantastic!

@aluzzardi is it possible to have a Markdown describing Swarm's testing plan? Like what you did for the roadmap.

aluzzardi:
@chanwit Yes! I'll be working some more on the framework so it will be possible to write more advanced tests (resources, scheduler policy, constraints, ...) and then I'll put a testing plan somewhere (hopefully sometime next week).

I'm planning to use these for regression tests as well. Right now, the version of docker in declared in DOCKER_VERSION (only 1.5 is available atm). Eventually, I'd like to have all versions of docker available in the registry and run the integration tests against all the supported ones (starting at 1.4 and onwards) and perhaps even include master (so we are sure that swarm works all the time with the latest docker before it's even released).

Hopefully I can get this running on docker's jenkins to run nightly. @jfrazelle?

chanwit:
@aluzzardi this is great!
Waiting for this PR to be merged, I'll crafting some tests for basic commands I can think of in the mean time.

ehazlett:
minions

:+1:

test/integration/helpers.bash
```diff
@@ -0,0 +1,76 @@
...
+
+# Use a random base port (for engines) between 5000 and 6000.
+BASE_PORT=$(( ( RANDOM % 1000 )  + 5000 ))
+
+# Join an array with a given separator.
+function join { local IFS="$1"; shift; echo "$*"; }
+
+# Run the swarm binary.
+function swarm() {
+	${SWARM_ROOT}/swarm $@
```
kojiromike: Is the bare wordsplitting intended, or should `$@` be in double quotes?

test/integration/helpers.bash
```diff
@@ -0,0 +1,76 @@
...
+
+# Start N docker engines.
+function start_docker() {
+	local current=${#DOCKER_CONTAINERS[@]}
+	local instances="$1"
+	shift
+	local args="$@"
+
+	# Start the engines.
+	for i in `seq $current $((instances - 1))`; do
```
kojiromike: Should `i` be marked `local`?

kojiromike: For that matter, would `for ((i=current; i < instances; i++)); do` be an improvement (avoiding the subshell and call to `seq`?

test/integration/helpers.bash
```diff
@@ -0,0 +1,76 @@
+#!/bin/bash
+
+# Root directory of the repository.
+SWARM_ROOT=${BATS_TEST_DIRNAME}/../..
+
+# Docker image and version to use for integration tests.
+DOCKER_IMAGE=${DOCKER_IMAGE:-aluzzardi/docker}
```
vieux: Where is this used ?

aluzzardi:
Thanks @kojiromike, I've made the changes.

PTAL

test/integration/helpers.bash
```diff
@@ -0,0 +1,77 @@
...
+	local instances="$1"
+	shift
+	local args="$@"
+	local i
+
+	# Start the engines.
+	for ((i=current; i < instances; i++)); do
+		local port=$(($BASE_PORT + $i))
+		HOSTS[$i]=127.0.0.1:$port
+		DOCKER_CONTAINERS[$i]=$(docker run -d --name node-$i -h node-$i --privileged -p 127.0.0.1:$port:$port -it ${DOCKER_IMAGE}:${DOCKER_VERSION} docker -d -H 0.0.0.0:$port $args)
```
kojiromike: Sorry I didn't catch this the first time around. Why "$args" instead of just "$@" at the end of the `docker` command? Right now you lose the splitting of arguments passed to `start_docker`.

test/integration/helpers.bash
```diff
@@ -0,0 +1,77 @@
...
+DOCKER_VERSION=${DOCKER_VERSION:-1.5}
+
+# Host on which the manager will listen to (random port between 6000 and 7000).
+SWARM_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 6000 ))
+
+# Use a random base port (for engines) between 5000 and 6000.
+BASE_PORT=$(( ( RANDOM % 1000 )  + 5000 ))
+
+# Join an array with a given separator.
+function join { local IFS="$1"; shift; echo "$*"; }
```
kojiromike: This comment is just about consistent style. Should `join` have parentheses after it to be consistent with other function definitions? FWIW if you have the parens, you don't need the `function` keyword.

test/integration/helpers.bash
```diff
@@ -0,0 +1,77 @@
...
+	local instances="$1"
+	shift
+	local args="$@"
+	local i
+
+	# Start the engines.
+	for ((i=current; i < (current + instances); i++)); do
+		local port=$(($BASE_PORT + $i))
+		HOSTS[$i]=127.0.0.1:$port
+		DOCKER_CONTAINERS[$i]=$(docker run -d --name node-$i -h node-$i --privileged -p 127.0.0.1:$port:$port -it ${DOCKER_IMAGE}:${DOCKER_VERSION} docker -d -H 0.0.0.0:$port $args)
```
kojiromike: Sorry I didn't catch this the first time around. Why $args instead of just "$@" at the end of the docker command? Right now you lose the splitting of arguments passed to start_docker.

test/integration/helpers.bash
aluzzardi: It was just to make it obvious that args don't container the first one (since it's right after shift).

It doesn't really though, I can use "$@" instead

aluzzardi:
@kojiromike PTAL

vieux:
LGTM

@kojiromike please open another PR if you can to fix stuff.

kojiromike:
Happy to, but I think it's all good now. :)

PR #451, opened 2015-03-07
use number of containers in binpack + fix some tests

vieux:
When using binpacking, if multiple nodes get the same weight, swarm will now look at the number of containers to choose the node with the higher number of container.

### DIFF
--- a/scheduler/strategy/binpack.go
+++ b/scheduler/strategy/binpack.go
@@ -23,5 +23,15 @@ func (p *BinpackPlacementStrategy) PlaceContainer(config *dockerclient.Container
 	// sort by highest weight
 	sort.Sort(sort.Reverse(weightedNodes))
 
-	return weightedNodes[0].Node, nil
+	topNode := weightedNodes[0]
+	for _, node := range weightedNodes {
+		if node.Weight != topNode.Weight {
+			break
+		}
+		if len(node.Node.Containers()) > len(topNode.Node.Containers()) {
+			topNode = node
+		}
+	}
+
+	return topNode.Node, nil
 }
--- a/scheduler/strategy/binpack_test.go
+++ b/scheduler/strategy/binpack_test.go
@@ -28,6 +28,38 @@ func createContainer(ID string, config *dockerclient.ContainerConfig) *cluster.C
 	return &cluster.Container{Container: dockerclient.Container{Id: ID}, Info: dockerclient.ContainerInfo{Config: config}}
 }
 
+func TestPlaceEqualWeight(t *testing.T) {
+	s := &BinpackPlacementStrategy{}
+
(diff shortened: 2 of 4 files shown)


scheduler/strategy/binpacking.go
```diff
@@ -54,5 +55,16 @@ func (p *BinPackingPlacementStrategy) PlaceContainer(config *dockerclient.Contai
 	// sort by highest weight
 	sort.Sort(sort.Reverse(weightedNodes))
 
-	return weightedNodes[0].Node, nil
+	topWeight := weightedNodes[0].Weight
+	for _, node := range weightedNodes {
```
aluzzardi: I think this could be written as:

```go
	for _, node := range weightedNodes; node.Weight == topWeight {
		node.Weight = int64(len(node.Node.Containers()))
		topNodes = append(topNodes, node)
	}
```

It will only iterate on top nodes (so if there are 100 nodes but only the first one has a better score it will stop immediately) and as added bonus it makes the code simpler.

WDYT?

aluzzardi: And actually to simplify further how about:

```go
        topNode := weightedNodes[0]
        for _, node := range weightedNodes; node.Weight == topNode.Weight {
                if len(node.Node.Containers()) > len(topNode.Node.Containers()) {
                        topNode = node
                 } 
        }
```

So you don't need to keep another array and sort again

vieux: `range weightedNodes; node.Weight == topNode.Weight {` does compile ?

vieux:
@aluzzardi PTAL

aluzzardi:
Please rebase

aluzzardi:
Also, could you do the same for spread?

vieux:
@aluzzardi done, PTAL

aluzzardi:
/cc @gabrtv

What do you think of this? Basically, it goes one step further of bin packing - all things being equal (as in, multiple machines have the exact same memory/cpu usage), the binpack strategy will pick the machine with most containers running on it.

The downside is: if you schedule containers with no resources, they'll all end up on the same host.

gabrtv:
@aluzzardi makes good sense to me.  

Is it true that `spread` strategy is the way to achieve even distribution when _not_ specifying resources?  If so I'd suggest we make `spread` the default scheduling strategy, as most first-time swarm consumers are *not* specifying resources and achieving even spread seems the desirable default.

vieux:
@gabrtv `spread` is already the default strategy today. And yes, with this patch, as we use the # of containers, it would give you an even distribution when you don't specify resources.

gabrtv:
:+1:

aluzzardi:
LGTM

PR #465, opened 2015-03-09
integration: add a test to verify creating cluster on hub

chanwit:
This is a test for `swarm create` and `swarm c`. These command should return a proper cluster token, hexadecimal string of size 32.

### DIFF
--- a/test/integration/helpers.bash
+++ b/test/integration/helpers.bash
@@ -88,24 +88,7 @@ function swarm_join() {
 		SWARM_JOIN_PID[$i]=$!
 		((++i))
 	done
-	wait_until_swarm_joined $i
-}
-
-# Wait until a swarm instance joins the cluster.
-# Parameter $1 is number of nodes to check.
-function wait_until_swarm_joined {
-	local attempts=0
-	local max_attempts=10
-
-	until [ $attempts -ge $max_attempts ]; do
-		run docker -H $SWARM_HOST info
-		if [[ "${lines[3]}" == *"Nodes: $1"* ]]; then
-			break
-		fi 
-		echo "Checking if joined successfully for the $((++attempts)) time" >&2
-		sleep 1
-	done
-	[[ $attempts -lt $max_attempts ]]
+	retry 30 1 [ `docker_swarm info | grep -q "Nodes: $i"` -eq 0 ]
 }
 
 # Stops the manager.
--- a/test/integration/token-discovery.bats
+++ b/test/integration/token-discovery.bats
@@ -1,0 +1,29 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function token_cleanup() {
+	curl -X DELETE https://discovery-stage.hub.docker.com/v1/clusters/$1
+}
+
+function teardown() {
+	swarm_join_cleanup
+	swarm_manage_cleanup
+	stop_docker
+}
+
+@test "token discovery should be working properly" {
+	start_docker 2
+
+	TOKEN=$(swarm create)
+	[ "$status" -eq 0 ]
+	[[ ${TOKEN} =~ ^[0-9a-f]{32}$ ]]
+
+	swarm_manage token://$TOKEN
+	swarm_join   token://$TOKEN
+
+	run docker_swarm info
+	[[ "${lines[3]}" == *"Nodes: 2"* ]]
+
+	token_cleanup $TOKEN
+}


bacongobbler:
I'm liking this new integration test library :+1:

chanwit:
Me too! it's a cool stuff done by @aluzzardi

aluzzardi:
I believe once we get #582 merged we could update this to do a full test of the token discovery.

chanwit:
Relates to #644

chanwit:
@aluzzardi I think I need your advise on how should we continue the `token` tests?

aluzzardi:
@vieux: Would there be a way to destroy clusters? Worried that this test will create thousands of clusters overtime.

abronan:
@aluzzardi There is a `DELETE` method on the Hub discovery service to destroy clusters. We can use it to delete the created cluster at the end of the test (or in `teardown()`).

abronan:
Well for now you have to use:

```
curl -X DELETE https://discovery-stage.hub.docker.com/v1/clusters/<cluster-id>
```

As there is no `swarm destroy` command..

chanwit:
@aluzzardi could you please check this command:

`godep go run main.go create | echo $?`
On my machine, it returns `2`.

While `$GOBIN/swarm create | echo $?` returns `0`.

With this, we cannot properly check `$status` via `godep go run`. Any idea?

chanwit:
@aluzzardi never mind my ping. I think it's caused by the Hub does not response sometime. `godep go run main.go create` now returns `0`.

test/integration/create-cluster-on-hub.bats
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function token_cleanup() {
+	curl -X DELETE https://discovery-stage.hub.docker.com/v1/clusters/$1
```
abronan: We should probably wait for a `swarm destroy` command so you can actually test the creation and then the destruction of a cluster.

chanwit: @abronan OK. I'll wait for it.

aluzzardi: I'm not sure if we're going to get the `swarm destroy` soon enough and we need this integration test.

For me it's an LGTM

@abronan @vieux Thoughts?

abronan: @aluzzardi Well, we can change it if we introduce something like a `swarm destroy` command in the future, so this is not an issue. Let's merge it if we need it now :+1:

chanwit:
Please do not merge yet. I still can't do 100% regression for all tests.

chanwit:
done. could you please review again.

test/integration/helpers.bash
```diff
@@ -88,7 +88,7 @@ function swarm_join() {
 		SWARM_JOIN_PID[$i]=$!
 		((++i))
 	done
-	wait_until_swarm_joined $i
+	wait_until_swarm_joined $i "$@"
```
aluzzardi: I'm wondering if `wait_until_swarm_joined` could use the new `retry` function (just like `wait_until_reachable`)

Maybe something like ```retry 10 1 [ `swarm list "$@" | wc -l` -eq $i ]```

test/integration/create-cluster-on-hub.bats
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bats
```
aluzzardi: Also, regarding the name, I'd call it `token-discovery.bats` to be consistent with the other integration tests you wrote (`file-discovery.bats`, etc)

test/integration/helpers.bash
chanwit: @aluzzardi I haven't looked at your new `retry` yet. Will be trying it out.

test/integration/create-cluster-on-hub.bats
chanwit: sure :)

test/integration/helpers.bash
```diff
@@ -88,24 +88,7 @@ function swarm_join() {
...
-	until [ $attempts -ge $max_attempts ]; do
-		run docker -H $SWARM_HOST info
-		if [[ "${lines[3]}" == *"Nodes: $1"* ]]; then
-			break
-		fi 
-		echo "Checking if joined successfully for the $((++attempts)) time" >&2
-		sleep 1
-	done
-	[[ $attempts -lt $max_attempts ]]
+	retry 30 1 [ `swarm list "$@" | wc -l` -eq $i ]
```
chanwit: @aluzzardi I set this large to let the test pass as my machine is quite far from Hub. Maybe `30` would be reduced when you've got integration tests run on CI.

test/integration/token-discovery.bats
```diff
@@ -0,0 +1,28 @@
...
+	start_docker 2
+
+	TOKEN=$(swarm create)
+	[ "$status" -eq 0 ]
+	[[ ${TOKEN} =~ ^[0-9a-f]{32}$ ]]
+
+	swarm_manage token://$TOKEN
+	swarm_join   token://$TOKEN
+
+	retry 10 1 docker_swarm info && [[ "${lines[3]}" == *"Nodes: 2"* ]]
```
chanwit: @aluzzardi This too to avoid network time-out problem. Not sure if you're OK with this.

aluzzardi: Oh, I see the problem - basically `docker list` returns 2 items faster than what swarm sees, right?

In that case, what about changing the test in swarm_join from:
`[ `swarm list "$@" | wc -l` -eq $i ]`
to something that runs docker_swarm info. Maybe:
[ -n `docker_swarm info | grep "Nodes: $i" ]

test/integration/helpers.bash
```diff
@@ -88,24 +88,7 @@ function swarm_join() {
...
-	until [ $attempts -ge $max_attempts ]; do
-		run docker -H $SWARM_HOST info
-		if [[ "${lines[3]}" == *"Nodes: $1"* ]]; then
-			break
-		fi 
-		echo "Checking if joined successfully for the $((++attempts)) time" >&2
-		sleep 1
-	done
-	[[ $attempts -lt $max_attempts ]]
+	retry 30 1 [ `docker_swarm info | grep "Nodes: $i"` -eq 0 ]
```
chanwit: @aluzzardi should this do?

aluzzardi: Yup, I think so! Not sure what `-eq 0` is doing though? `[ -n ``docker_swarm ... ``]` should work too I guess (`-n` = the length of STRING is nonzero)

chanwit: it's testing the exit code of the expression in backticks.
`if [[ `echo 2 | grep -q 2` -eq 0 ]]; then echo yes; else echo no; fi` shows yes and 
`if [[ `echo 2 | grep -q 2` -eq 1 ]]; then echo yes; else echo no; fi` shows no for me.

Maybe I should add `-q` to it.

chanwit:
regression goes well. PTAL.
@aluzzardi @vieux @abronan

aluzzardi:
Awesome!

As you may have noticed, we run those automatically now (`janky` in the list of GH checks)

LGTM

chanwit:
Wow, that `janky` is super awesome! I've check its console output and all discovery tests are doing well there. Super!

aluzzardi:
@chanwit yep! If any integration test failed, this PR would be marked as failed (just like when travis fails because of broken tests)

aluzzardi:
Related to #465

chanwit:
@aluzzardi I think you just did a circular reference here.

aluzzardi:
Woops.

#644

PR #468, opened 2015-03-10
add support for docker ps -l / -n

### LINKED ISSUE #426: `docker ps -l` shows last *running* container
**What I do:** `docker run -d busybox sleep 42; docker run busybox true; sleep 1; docker ps -l`

**What I expect to see:** the `true` container

**What I see instead:** the `sleep` container

Note: I understand that there might be (at least) two ways to implement this correctly:
- display the last container started across the cluster;
- display the last container started *through this Swarm master*.

The latter version, while not 100% correct technically, *might* be easier to implement, and would still be pretty useful (and could display nothing if no container was started through this Swarm master).

```
$ docker version
Client version: 1.4.1
Client API version: 1.16
Go version (client): go1.3.3
Git commit (client): 5bc2ff8
OS/Arch (client): linux/amd64
Server version: swarm/0.1.0
Server API version: 1.16
Go version (server): go1.3.3
Git commit (server): 2acbea1
```

Note: actually, I wonder if `docker ps -l` is implemented at all; in which case, feel free to rename this issue to "`docker ps -l` is not implemented" :cry:

vieux:
Fix #426

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"runtime"
 	"sort"
+	"strconv"
 	"strings"
 
 	log "github.com/Sirupsen/logrus"
@@ -117,12 +118,13 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	all := r.Form.Get("all") == "1"
+	limit, _ := strconv.Atoi(r.Form.Get("limit"))
 
 	out := []*dockerclient.Container{}
 	for _, container := range c.cluster.Containers() {
 		tmp := (*container).Container
 		// Skip stopped containers unless -a was specified.
-		if !strings.Contains(tmp.Status, "Up") && !all {
+		if !strings.Contains(tmp.Status, "Up") && !all && limit <= 0 {
 			continue
 		}
 		// Skip swarm containers unless -a was specified.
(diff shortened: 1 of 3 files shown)


vieux:
ping @aluzzardi, can you also take a look at the tests.

Thanks

api/api.go
```diff
@@ -117,12 +118,13 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	all := r.Form.Get("all") == "1"
+	limit, _ := strconv.Atoi(r.Form.Get("limit"))
```
aluzzardi: What happens when no limit is provided? Does it default to 0?

api/api.go
```diff
@@ -117,12 +118,13 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
...
 
 	out := []*dockerclient.Container{}
 	for _, container := range c.cluster.Containers() {
 		tmp := (*container).Container
 		// Skip stopped containers unless -a was specified.
-		if !strings.Contains(tmp.Status, "Up") && !all {
+		if !strings.Contains(tmp.Status, "Up") && !all && limit <= 0 {
```
aluzzardi: nit: Maybe the limit could be checked outside the for loop since it doesn't make sense to iterate over all containers if limit is 0.

Then again, I think it would be a useless optimization: limit of 0 is a strange limit and it doesn't make much sense to optimize for this.

In short: up to you :) feel free to discard this

aluzzardi:
Minor comments (probably useless), otherwise LGTM

api/api.go
vieux: yes

api/api.go
vieux: yes it does, if limit is 0 it means no limit

PR #471, opened 2015-03-13
add an environment variable to expose the node's ip address

hoonmin:
Hi.
What about providing swarm node's IP address inside it's containers?
As swarm-agent starts with an reachable/exposed address, it should be useful for some cases.
(e.g. a linked container writing discovery informations like <nodes' IP>:<mapped_port>)

This PR adds an environment variable `SWARM_NODE_TCP_ADDR` before creating a container.

### DIFF
--- a/cluster/swarm/node.go
+++ b/cluster/swarm/node.go
@@ -20,6 +20,9 @@ const (
 
 	// Timeout for requests sent out to the node.
 	requestTimeout = 10 * time.Second
+
+	// Environment variable to expose the node's ip address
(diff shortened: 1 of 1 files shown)


chanwit:
In an orchestration, it's the responsibility of Docker Machine, not Swarm, to expose the node's IP address, AFAIK.

vieux:
Also, when the container get's rescheduled, the IP could change. I think it's safer to not add the IP to the env.

aluzzardi:
Agreed - this should be a feature of Docker, not Swarm.

PR #476, opened 2015-03-16
introduce engine.go

vieux:
While @tnachen and I were working on the Mesos integration, we realised Mesos needed some sort of node as well.

I suggest the following refactor:

- a new  **class** `cluster/engine.go` (common code to swarm and mesos)
- `cluster/swarm/node.go` inherit from `cluster/engine.go` and implements it's own create function using the engine directly
and for mesos
- `cluster/mesos/node.go` inherits from `cluster/engine.go` + has a list of offers and implements it's own create function using the offers

I'm not entirely sure about this new design so please give your feedback.

Thanks

EDIT: If go had proper inheritance, `engine` would be an abstract class, and `swarm/node` and `cluster/node` would simply implement it.


vieux:
ping @abronan @aluzzardi @tnachen

bacongobbler:
@vieux how about this?

 - a new **interface** of type `Engine` in `cluster/engine.go` (common interface to both swarm and mesos nodes)
 - `cluster/swarm/node.go` has everything from `cluster/engine.go` + implements it's own create function using the engine directly and for mesos
 - `cluster/mesos/node.go` embeds `cluster/swarm/node.go` + has a list of offers and has its own create function using the offers

That way if someone were to implement another cluster module (e.g. kubernetes, fleet) we have a standardized interface to work from rather than everyone embedding `engine.Engine` into their cluster modules and overriding it's behaviour (becomes optional rather than mandatory).

vieux:
@bacongobbler we already have a standard interface for other cluster modules: `cluster/node.go` we don't want to add a 3rd one `Cluster` and `Node` seems enough.

I did the `engine` class because I didn't want the `mesos` pkg to import the `swarm` pkg (and I'm not sure it's possible because of dependencies) it's just common code, it could be named `BaseNode`

aluzzardi:
I'm worried about this - we are adding more and more abstractions and types and exposing internals (e.g. `RefreshContainer` becomes public).

I'm afraid it will become complex and hard to maintain.

We also have to imagine that eventually (6 months or so), once we get plugins (as in, dynamically loaded) all the Mesos code will live outside of Swarm and won't have access to any types/interfaces: it will only communicate through some sort of (undefined yet) JSON API and I have no idea how we are going to do that.

This is kind of an edge case since we are building a Swarm framework for Mesos and therefore need to embed our own scheduler - but it's the only occurrence we will ever need to do this.

I don't have a solution for this, I just wanted to raise awareness and get your opinion.

Also, I understand the reasoning that we don't want to import `swarm` into `mesos`. However, what about this: Imagine the `mesos` pkg was actually living in a separate repository. Would you do things differently? I guess `swarm` wouldn't be a pkg anymore (it will probably be the way it was, into the `cluster` pkg directly), and `mesos` would just import `cluster`. Again, I don't know where I am going with this, you have a better idea of the whole picture, I'm just trying to kind of prepare for the future where this code won't live into this repository.

vieux:
@aluzzardi I understand your concerns, I have the same.
Just to clarify, `RefreshContainer` is public because go has no concept of protected.

tnachen:
@aluzzardi it sounds like a valid concern, but I think given the language constraint and the requirements we know I'm not sure what's a better interface.
IMO this seems like a good small step, as swarm I imagine should also provide good base framework code to help different backends to plugin to the strategies and commands swarm supports.
That said, I think what's good for the swarm code base is to start another scheduler backend design within a month or so the code don't become too opinionated towards certain backend and help design the interface that is simple but yet flexible, and help compose better.
I'm happy to help design the interface with another interested scheduler backend and we should be able to make a better design and iterate.

cluster/engine.go
```diff
@@ -0,0 +1,508 @@
+package cluster
```
aluzzardi: I believe this should be either `git mv`ed or `git cp`ed as it looks like git thinks it's a brand new file

vieux: done, I'm not sure why it didn't work with for the test, but at least it worked for the code

cluster/engine.go
```diff
@@ -0,0 +1,527 @@
...
+}
+
+// Refresh the list of images on the node.
+func (n *Engine) refreshImages() error {
+	images, err := n.Client.ListImages()
+	if err != nil {
+		return err
+	}
+	n.Lock()
+	n.images = nil
```
noxiouz: Seems if `n.images` is not used anywhere at that time it's better to reuse allocated memory:
```
n.images = n.images[:0]
```

aluzzardi:
Taking a step back:

What if we revert `Node` the way it was (as in, not an interface)? Maybe renaming it to `Engine` would make sense since it represents an Engine and that's true in both cases (swarm and mesos clusters).

That way, Mesos can leverage all the code in Engine and it should be exactly the same: after all, an engine is still an engine in both cases.

Now, for the scheduler, what we could do is have a different structure (and maybe this one can be called `Node`) that contains all the relevant informations for scheduling (Images, CPU, ...). This structure would be filled by the `Cluster` implementation: In the case of swarm, it would simply be a copy of the information provided by `Engine`, in the case of mesos, it would be filled by a combination of `Engine` and whatever mesos-specific stuff we can get.

This would also help in cases such as https://github.com/docker/swarm/issues/568: the `Node` structure could have scheduler specific data such as keeping track of whether the node has already been tried or not.

@vieux @tnachen: Thoughts?

vieux:
@aluzzardi I feel like there will be a few quirks but it might work.  I'll play with the idea and comment back here.

vieux:
replaced by #574

PR #481, opened 2015-03-17
mitigate race in pull while we wait for a proper store

vieux:

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -67,8 +67,11 @@ func (c *Cluster) CreateContainer(config *dockerclient.ContainerConfig, name str
 
 	c.RLock()
 	defer c.RUnlock()
-
 retry:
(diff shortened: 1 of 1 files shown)


abronan:
LGTM

abronan:
ping @aluzzardi

chanwit:
Wow, we've got a new maintainer @abronan on board?
I am so glad to have more man-powers for code reviews, so my PRs could get merged faster :)

aluzzardi:
Could you add a //FIXME or //TODO?

This is hack-ish but I know there's no other way around for now.

Other than that, LGTM

aluzzardi:
Ohhh - actually, silly request: I see the code coverage decreased with this PR.

I'd like eventually to force all PRs to never decrease the coverage: do you think you could be able to do that as a test with this one?

You could possibly rely on mocks and return different data during the different calls.

It may be overkill though - it's up to you :)

Issue #488, opened 2015-03-21
Minor typos in "Filter" documentation


Nickology:
Though the content appears to be identical, there are two small typos on [docs.docker.com](http://docs.docker.com/swarm/scheduler/filter/) and one in the [README.md](https://github.com/docker/swarm/blob/master/scheduler/filter/README.md).

On [docs.docker.com](http://docs.docker.com/swarm/scheduler/filter/):
* _You can schedule a container only on nodes where the images **are** already pulled._
* _So 'NoDe1' or 'NODE1' will also **match**._

In [swarm/scheduler/filter/README.md](https://github.com/docker/swarm/blob/master/scheduler/filter/README.md):
* _So 'NoDe1' or 'NODE1' will also **match**._

abronan:
Hey @Nickology, thank you for reporting the typos! Would you like to submit a Pull Request for this? We would be happy to merge :)

Nickology:
Hi @abronan. I'll submit one right away. :)

PR #489, opened 2015-03-21
Corrected two typos

Nickology:
Corrected two typos: one described in #488, one new one. This does not fix typos located on [docs.docker.com](http://docs.docker.com/swarm/scheduler/filter/)

Signed-off-by: Nicolas Jouannem <nick@nickology.com>


abronan:
LGTM

vieux:
Thanks @Nickology, docs.docker.com will be updated at the next release.

PR #512, opened 2015-03-25
Move note on top of Nodes config to avoid confusion

abronan:
Moving the Note at the bottom of the Nodes Configuration part for better visibility. Saw a lot of confusion on IRC and a few Issues on the setup part (and the fact that you need to launch the docker daemon listening on TCP).

ping @SvenDowideit

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -43,6 +43,11 @@ Each swarm node will run a swarm node agent which will register the referenced
 Docker daemon, and will then monitor it, updating the discovery backend to its
 status.
 
+> **Note**: In order for the Swarm manager to be able to communicate with the node 
+agent on each node, they must listen to a common network interface. This can be 
+achieved by starting docker with the `-H` flag (e.g. `docker -H tcp://0.0.0.0:2375 -d` 
+on each swarm node).
+
 The following example uses the Docker Hub based `token` discovery service:
 
 ```bash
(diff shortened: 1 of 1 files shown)


vieux:
@abronan can you please sign your commit :P

abronan:
@vieux Errr just changed my git aliases and forgot the -s :ghost:

aluzzardi:
LGTM

@SvenDowideit ?

abronan:
ping @moxiegirl

moxiegirl:
@abronan When you have a document with this many notes in odd places, it can indicate structural problems.  I think you would get a better performance on the page by clarifying what is required to run Swarm in a prereq section. Then, making the Setup procedure look like one.  So, I recommend this:

https://gist.github.com/moxiegirl/5874c99da108bb630763

abronan:
Thank you for the feedback! I definitely agree, will update with the proper structure.

abronan:
@moxiegirl Updated thanks to your feedback and added a few details.

moxiegirl:
@abronan You lost the alignment on the indented procedure i. and ii. --- Also, just as an fyi it isn't good practice to preface with at `$` if you are only showing command syntax.  That is why I removed them.  People expect to see a real command line example and you are actually illustrating the command line syntax/format.

vieux:
@moxiegirl can you please give me one last LGTM ? Thanks

moxiegirl:
LGTM

abronan:
@moxiegirl Thanks for the thorough review :+1:

PR #513, opened 2015-03-25
Closes #501: Error if unknown args are used in create

### LINKED ISSUE #501: Add error checking on swarm create if unknown arguments are used.
**Problem**

swarm create command must fail if an unknown argument is used.

**Solution**

Check if argument length is > 0 in the code that handles the create command in `main.go` `main()` function using : if len(c.Args()) != 0 and log a fatal message

technolo-g:
This implements the suggested method to check for additional arguments passed to the `create` command.

### DIFF
--- a/main.go
+++ b/main.go
@@ -67,6 +67,9 @@ func main() {
 				discovery := &token.TokenDiscoveryService{}
 				discovery.Initialize("", 0)
 				token, err := discovery.CreateCluster()
+				if len(c.Args()) != 0 {
+					log.Fatalf("the `create` command takes no arguments. See '%s create --help'.", c.App.Name)
+				}
 				if err != nil {
 					log.Fatal(err)
 				}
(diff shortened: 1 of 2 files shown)


technolo-g:
Should I add tests for this? If so, could someone point me in the right direction?

vieux:
can you add `See '%s create --help'.", c.App.Name)` like the error message below ?
Are you able to run the tests in https://github.com/docker/swarm/tree/master/test/integration with bats ? 
If so, adding a test should be pretty trivial.

abronan:
LGTM

vieux:
LGTM

vieux:
big thanks @technolo-g @endzyme @mattknox @hasuf @cap10morgan Fed & Tom!

cap10morgan:
Happy #dockerbday!

technolo-g:
We'd also like to thank the Academy and @mattk42! Huzzah!

abronan:
`run swarm create derpderpderp` :+1:  :joy:

PR #518, opened 2015-03-26
Fix 404s that manifest in GitHub but not in the Docs site. Links should ...

spanktar:
...work in both.

Spoke with Mary (Moxiegirl) about this issue and it occurs because
when Docs is built, it pulls files from various places. This causes
the GitHub view to have broken links (and images) which isn't ideal
since people may come to the docs via either approach.

Signed-off-by: Tom Kapanka <tom@kapanka.com>

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -94,15 +94,15 @@ the certificates.
 
 ## Discovery services
 
-See the [Discovery service](discovery.md) document for more information.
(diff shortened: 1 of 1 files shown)


vieux:
LGTM

vieux:
@SvenDowideit ?

moxiegirl:
LGTM

PR #526, opened 2015-03-27
Make Swarm CpuShares compatible with Docker API

smothiki:
fixes https://github.com/docker/swarm/issues/475 
Memory is a very large value and applying overcommit ratio increases the total memory significantly  but the increase in total cpus after applying overcommit ratio is negligible. Hence I removed the overcommit ratio for Cpus
```
$ docker -H 172.17.8.101:2395 run --name some1-mysql -e MYSQL_ROOT_PASSWORD=mysecretpassword -c 512 -d mysql
e147f9aa38a35e3ba789e0f3171d50af206eb5aca906a56bc102ec238d4f8bf9
```
To show used Cpus in docker info , the cpu shares are normalized to total number of Cpus .

```
$docker -H 172.17.8.101:2395 info
Containers: 5
Nodes: 2
 deis-02: 172.17.8.101:2375
  └ Containers: 3
  └ Reserved CPUs: 0.500 / 1
  └ Reserved Memory: 0 B / 2.057 GiB
 deis-01: 172.17.8.100:2375
  └ Containers: 2
  └ Reserved CPUs: 1.000 / 1
  └ Reserved Memory: 0 B / 2.057 GiB
```

### DIFF
--- a/api/events_test.go
+++ b/api/events_test.go
@@ -28,7 +28,7 @@ func (fn *FakeNode) Image(_ string) *cluster.Image         { return nil }
 func (fn *FakeNode) Containers() []*cluster.Container      { return nil }
 func (fn *FakeNode) Container(_ string) *cluster.Container { return nil }
 func (fn *FakeNode) TotalCpus() int64                      { return 0 }
-func (fn *FakeNode) UsedCpus() int64                       { return 0 }
+func (fn *FakeNode) UsedCpus() float64                     { return 0 }
 func (fn *FakeNode) TotalMemory() int64                    { return 0 }
 func (fn *FakeNode) UsedMemory() int64                     { return 0 }
 func (fn *FakeNode) Labels() map[string]string             { return nil }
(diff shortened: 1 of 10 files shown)


vieux:
@jhamrick does it look good to you ?

@aluzzardi @abronan even though it's a big, breaking change, I think we merge it.
If we are not comfortable with it, we might add a `--cpu-mode` to swarm to say we want to use `-c 1` or `-c 1024`

smothiki:
+1 for having --cpu-mode

aluzzardi:
NOT LGTM

Will add comments to #475 directly

vieux:
@smothiki @aluzzardi now that cpu quotas are merged in docker https://github.com/docker/docker/pull/10736 we should use those instead of `cpushares`

smothiki:
@vieux Please close this PR. If possible I will submit a new PR soon using cpu quotas if no one is working on it

aluzzardi:
Thanks @smothiki 

Awesome, looking forward for a CPU quotas PR

PR #528, opened 2015-03-28
Validate value by checking for whitespace in constraint expression

snrism:
When using `operatingsystem` standard constraint, it fails when we use the value shown in `docker info`. For example, if I try to create the following for ubuntu OS, 
docker run -d -e constraint:operatingsystem=="Ubuntu 14.04.1 LTS" --name snort glanf/snort 
swarm returns
FATA[0000] Error response from daemon: Value 'Ubuntu 14.04.1 LTS' is invalid

however, if I use 
docker run -d -e constraint:operatingsystem=="Ubuntu*" --name snort glanf/snort 
it identifies the host and creates the container.

This is because while checking for the expression in expr.go, the matchstring does not check for whitespaces. This PR fixes this issue.

### DIFF
--- a/scheduler/filter/README.md
+++ b/scheduler/filter/README.md
@@ -291,7 +291,7 @@ This filter will prevent scheduling containers on unhealthy nodes.
 
 ## Docker Swarm documentation index
 
-- [User guide](./../index.md)
-- [Discovery options](./../discovery.md)
-- [Sheduler strategies](./strategy.md)
-- [Swarm API](./../API.md)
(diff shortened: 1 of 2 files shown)


vieux:
@snrism can you add a test ? thanks

vieux:
Can you remove the last commit (merge commit) and do a proper rebase ? Thanks!

snrism:
Oops..lemme fix that

snrism:
@vieux Sorry..let me open a new pull request with the changes. I think this is messed up..

PR #479, opened 2015-03-17
Add windows CI through Appveyor

weswigham:
The project will likely need to be activated by maintainers at https://ci.appveyor.com/projects

Also, it would be nice to have a status badge from https://ci.appveyor.com/project/docker/swarm/settings/badges added to the readme as well.

PR #537, opened 2015-03-30
Add windows CI through Appveyor

weswigham:
#479 Was closed from some CI issues, reopening for comment.

The project will likely need to be activated by maintainers at https://ci.appveyor.com/projects

Also, it would be nice to have a status badge from https://ci.appveyor.com/project/docker/swarm/settings/badges added to the readme as well.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,30 @@
+version: '{build}'
+build_script:
+- ps: >-
+    $env:GOPATH="c:/gocode";
+
+    mkdir $env:GOPATH | Out-Null
+
+    $env:GOBIN="$env:GOPATH/bin";
+
+    $env:PATH="$env:GOBIN;$env:PATH";
+
+    go version
+
+    go env
+
+
+    #Get deps
+
+    go get golang.org/x/tools/cmd/vet
+
+    go get github.com/tools/godep
+
+    godep restore
+
+    go get .
+test_script:
+- ps: >-
+    go vet ./...
+
+    go test -v -cover -race ./...


aluzzardi:
Thanks @weswigham !

So we are actually thinking of using Docker's Jenkins infrastructure at https://jenkins.dockerproject.com/

There are already runners for both Linux and Windows therefore we would use the same CI configuration for both platforms.

Thoughts?

ahmetalpbalkan:
Depending on the tests ran in the CI, something like AppVeyor could just suffice. e.g. Distribution project right now uses CircleCI because it's simply bunch of unit tests with "go test ./..." and various other checks. Docker itself on the other hand requires far more complex environment to run the tests amd they take long. Core Maintainers can explain better why they switched from Drone.io to hosted Jenkins.

aluzzardi:
@ahmetalpbalkan We are actually already using Jenkins (https://jenkins.dockerproject.com/job/Swarm%20Master/) to run our integration test suite.

It launches docker engines to try out swarm with a real engine/CLI, just like `integration-cli` of Docker.

ahmetalpbalkan:
@aluzzardi I see now. Several questions to evaluate before testing on Windows:

- Why are we testing Swarm on Windows? 
- Will swarm cli (e.g. join or manage) be supported on windows (can't see any reason why it should be until windows docker engine lights up)
- What sort of tests are valuable on windows (eg in docker core we only run integration-cli on windows and many parts solely testing the engine in that suite is skipped in windows b/c there's no value —windows right now is a cli-only platform like Mac).

weswigham:
You should probably run tests on Windows simply because `swarm manage` functions on Windows pretty well already. It's probably not worth risking regressing what functionality you already have.


_Upper left: swarm manage running on a windows host, upper right: docker daemon running on a linux host in the cluster, bottom: local windows machine running docker cli_

I'm not sure why docker engine would actually be required. As far as usefulness goes, I'm pretty sure that compose and machine both run in Windows already, and there's certainly a use case, as a developer, to want to spawn up a bunch of containers in VMs using machine and/or compose and then use swarm to mask them all to my local (windows) box.

ahmetalpbalkan:
@weswigham I guess it really comes down to if maintainers want to have Windows as a supported platform to host Swarm Manager. If that's the case, I believe tests for Swarm include creating a cluster: which is forming a cluster of actual docker engines in the same machine (which don't work on Windows).

For testing Docker core on Windows, we have Windows CI boxes on Jenkins, running the integ-cli tests and they talk to a pre-configured Docker engine living on some other machine, exclusively dedicated to that Jenkins box. So for each Windows Jenkins machine we have a Docker engine sitting on some other machine independently. Problem with that is, somebody has to refresh engine binaries periodically because they don't get built along with the PR submitted itself.

ahmetalpbalkan:
My opinion is, right now, there's not much point in making Swarm Manager supported on Windows (as it'd be weird if people host their slaves on Linux and manage them on a Windows Server). Theoretically, this will definitely be useful when the Docker Engine for Windows comes up.

I believe there hasn't been any specific work or fixes taken to support Swarm on Windows. It just works simply because it doesn't depend on the operating system and is a cross-compiling Go program. Therefore it's less likely to regress or will be easier to fix when it does actually regress.

If you just want to make sure the compilation does not break on Windows, yeah it makes sense to have a CI with AppVeyor, however without running the integration test suite on windows, I don't think we actually gain a lot from that CI.

aluzzardi:
@ahmetalpbalkan for the "why", I have no idea :) I was just questioning the usage of Appveyor versus our existing Jenkins. The usefulness of running those tests is still questionable.

Unrelated to this PR: For the integration-cli stuff and stale binaries, we have kind of a solution for that. Basically, we run our integration tests against a Docker engine running as a container within a Docker.

We have docker-in-docker images for every release of docker since 1.0.0 (https://hub.docker.com/u/dockerswarm/dind) and also docker-in-docker images for the master branch of docker that gets rebuilt at every merge (https://hub.docker.com/u/dockerswarm/dind-master).

However those are not built along the PR, but it's close enough.

ahmetalpbalkan:
@aluzzardi Yes I'm aware of the library/docker effort. It might help a bit but the main problem with running the integ-cli tests for Windows CLI is most tests were relying on the fact that engine was running in the same machine, I had to send around 50 pull requests to clean that up and make most of it work against an engine running on a different host.

One solution to that might be running Boot2Docker on the Jenkins Windows machine, compiling docker (from PR) in there and making Docker CLI talk to Docker engine built from the same source code on the same machine (living in a b2d VM). But then another problem comes up: Jenkins machines are actually VMs on some cloud provider and VMs can't run VMs. :)

aluzzardi:
@ahmetalpbalkan How about building the Docker image as part of the Windows testing process and then running the integration-cli against a docker-in-docker?

ahmetalpbalkan:
@aluzzardi makes sense. In this case, the test runner must coordinate delivering this image to the engine running on a remote machine. And yes, dind help in this case. cc: @jfrazelle would be interested in this.

aluzzardi:
@ahmetalpbalkan If the cli supports build, then we could leverage that.

```
DOCKER_HOST=linuxbox
docker build -t integration-cli:$(UUID or something) -f Dockerfile.integration .
docker run integration-cli:$(THAT SAME UID) -p <doceker port>:<some random port> docker -d [...]
DOCKER_HOST=linuxbox:<that random port>
[run cli commands]
```

ahmetalpbalkan:
Yes windows CLI can build docker's Dockerfile.

weswigham:
As an aside, assuming that integration tests (and not just unit tests as this PR would have brought) can start getting run under windows, would the swarm integration tests need to be remade in a crossplatform manner, much like docker core's integration-cli?

dnephin:
just as an fyi, we're using appveyor to run compose unit tests on windows, and to build the windows binary: https://ci.appveyor.com/project/docker/compose

It has worked well for us so far.

PR #539, opened 2015-03-31
Fix 404 Error in Filter and Strategy README files

snrism:
Update README files under filter and scheduler with correct documentation links


vieux:
LGTM

PR #543, opened 2015-03-31
digital ocean discovery

xcrezd:
Signed-off-by: dmitrik <dmitrik@ironsrc.com>

Adds digital ocean discovery package

All the droplets will be fetched periodically from DO API and added to swarm

optional env variables

SWARM_DISCOVERY_DIGITALOCEAN_PORT
docker port in digital ocean droplet

SWARM_DISCOVERY_DIGITALOCEAN_NETWORK
network type (public/private)


vieux:
@xcrezd something is wrong with your PR can you rebase on the last master and to a force push ?

aluzzardi:
I wouldn't want to go down the rabbit hole - if we add discovery for one provider then we'd need to add it to a bunch of different providers.

The same today can be achieved by using `docker-machine`: It has drivers for digital ocean and will set up swarm with discovery automatically for you.

/cc @vieux

xcrezd:
@aluzzardi 
Can you please explain how can i achieve this now by docker-machine.

xcrezd:
@vieux 

$ git ch master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
$ git pull
Already up-to-date.
$ git ch digital_ocean_discovery
Switched to branch 'digital_ocean_discovery'
$ git rebase master
Current branch digital_ocean_discovery is up to date.

shimont:
+1 great commit
i really love it

vieux:
@xcrezd don't you see your PR has more than 7k new lines and 51 lines changed ? I think there is an issue.

vieux:
@xcrezd my bad, it's godep :D

aluzzardi:
@xcrezd: Basically, docker-machine would create the nodes for you and then automatically set them up for discovery

abronan:
Thank you for the contribution @xcrezd, although as @aluzzardi mentioned, merging this would mean accepting PRs to include discovery mechanisms from many other providers, which will be hard for us to maintain in the long term as the code base increases. DiscoveryThis change would probably make more sense in a repository like `docker-machine` where the discovery mechanism would be directly handled by the tool when using the *Digital Ocean* driver, without relying on the Hub discovery service.

I'm closing this for now but feel free to comment below.

PR #545, opened 2015-03-31
Make container creation and destruction synchronous.

### LINKED ISSUE #427: Race condition when pulling+running named containers
**What I do:** 

On a clean cluster (no image present):

`docker run --name lemon -d busybox sleep 42 & docker  run --name lemon -d busybox sleep 42`

**What I expect to see:**

The ID of one container, and an error message about the name `lemon` being already in use.

**What I see instead:**

Two containers are started, confirmed by:

```
$ docker ps
CONTAINER ID        IMAGE                       COMMAND             CREATED             STATUS                  PORTS               NAMES
3f3f9c68f531        busybox:buildroot-2014.02   "sleep 42"          6 seconds ago       Up Less than a second                       52.11.76.162/lemon   
ca85e952ee6a        busybox:buildroot-2014.02   "sleep 42"          15 seconds ago      Up Less than a second                       52.11.85.52/lemon    
```

I believe Swarm checks the uniqueness of the name at the wrong time (i.e. before carrying on with the pull).

Ideas:
- having a global "name lock manager" in Swarm, so that before starting the pull, Swarm locks the name (maybe for a given time, so that if the machine doing the pull burns in flames, the name doesn't stay locked forever...)
- when doing the run, and when receiving events, use the lock manager, to report things like "WARNING: a container with an existing duplicate name was started!"

```
$ docker version
Client version: 1.4.1
Client API version: 1.16
Go version (client): go1.3.3
Git commit (client): 5bc2ff8
OS/Arch (client): linux/amd64
Server version: swarm/0.1.0
Server API version: 1.16
Go version (server): go1.3.3
Git commit (server): 2acbea1
```

aluzzardi:
This is in order to fix race conditions. Currently, container creation
can happen in parallel which means that the scheduler operates on stale
data.

Fixes #427

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -66,35 +66,16 @@ func (c *Cluster) Handle(e *cluster.Event) error {
 
 // CreateContainer aka schedule a brand new container into the cluster.
 func (c *Cluster) CreateContainer(config *dockerclient.ContainerConfig, name string) (*cluster.Container, error) {
+	c.scheduler.Lock()
+	defer c.scheduler.Unlock()
 
-	c.RLock()
-	defer c.RUnlock()
-retry:
-	// FIXME: to prevent a race, we check again after the pull if the node can still handle
-	// the container. We should store the state in the store before pulling and use this to check
-	// all the other container create, but, as we don't have a proper store yet, this temporary solution
-	// was chosen.
 	n, err := c.scheduler.SelectNodeForContainer(c.listNodes(), config)
 	if err != nil {
 		return nil, err
 	}
 
 	if nn, ok := n.(*node); ok {
-		container, err := nn.create(config, name, false)
-		if err == dockerclient.ErrNotFound {
-			// image not on the node, try to pull
-			if err = nn.pull(config.Image); err != nil {
-				return nil, err
-			}
-
-			// check if the container can still fit on this node
-			if _, err = c.scheduler.SelectNodeForContainer([]cluster.Node{n}, config); err != nil {
-				// if not, try to find another node
-				log.Debugf("Node %s not available anymore, selecting another one", n.Name())
-				goto retry
-			}
-			container, err = nn.create(config, name, false)
-		}
+		container, err := nn.create(config, name, true)
 		if err != nil {
 			return nil, err
 		}
@@ -113,8 +94,8 @@ retry:
 // RemoveContainer aka Remove a container from the cluster. Containers should
 // always be destroyed through the scheduler to guarantee atomicity.
 func (c *Cluster) RemoveContainer(container *cluster.Container, force bool) error {
-	c.Lock()
-	defer c.Unlock()
+	c.scheduler.Lock()
+	defer c.scheduler.Unlock()
 
 	if n, ok := container.Node.(*node); ok {
 		if err := n.destroy(container, force); err != nil {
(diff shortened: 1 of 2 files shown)


aluzzardi:
ping @vieux @abronan

vieux:
@smothiki can you also take a look ?

vieux:
Even though this mean we can lock the whole scheduler during a pull, this is the safest for now. LGTM

abronan:
@abronan LGTM to avoid races until we take scheduling decisions directly out of the store and hopefully schedule tasks in parallel.

cluster/swarm/cluster.go
```diff
@@ -66,35 +66,16 @@ func (c *Cluster) Handle(e *cluster.Event) error {
...
-	// the container. We should store the state in the store before pulling and use this to check
-	// all the other container create, but, as we don't have a proper store yet, this temporary solution
-	// was chosen.
 	n, err := c.scheduler.SelectNodeForContainer(c.listNodes(), config)
 	if err != nil {
 		return nil, err
 	}
 
 	if nn, ok := n.(*node); ok {
-		container, err := nn.create(config, name, false)
```
smothiki: I'm okay if this is a temporary fix . But don't you think that locking create container method will increase the overall container launch time ? . If there are 10 create container calls involving 10 different images simultaneously this is a worst case scenario . A better fix would be maintaining a queue of scheduled containers in node struct 
```
lock()
n, err := c.scheduler.SelectNodeForContainer(c.listNodes(), config)
 	if err != nil {
 		return nil, err
 	}
 n.addtoqueue(config)
unlock()
container, err := nn.create(config, name, false)

node.create method {
defer n.removefromqueue(config)
...........
.........
}
```
while applying filters or placing the container according to the scheduling strategy  Add a check for scheduleQueue
This would decrease the overall create container time significantly because we are not waiting for pul Image time

abronan: @smothiki It is a temporary fix. This will ensure that we can do heavy plumbing on the storage part and take scheduling decisions directly out of the store. I proposed a queue in another issue but basically between the `SelectNodeForContainer` and `addToQueue` part, we'll need to add a lot of checks (for name uniqueness, etc.). As this part is going to be temporary anyway, better to hold-on to the safest solution to work on the storage part in the meantime.

smothiki: I have been working for a couple of days for maintaining a scheduled container queue and I'm not close at all. I'm okay with the temporary Fix   .

aluzzardi: Thanks @smothiki, great idea.

This is a temporary fix to get 0.2.0 out of the door since being race-condition free is a top priority, and being fast comes way after. I definitely agree we need to optimize this, but for now it does the work since only `docker run` and `docker rm` would be locked, all the other commands would remain lock free.

Maybe this is exactly what you just described, but I think a simple implementation for your solution would be to have `addtoqueue` (for instance) add the config to the internal `containers` map of Node with status `pending` or something.

That way, there is no need to explicitly handle it in the scheduler: `node.Containers()` will return both actual running containers and scheduled ones.

Maybe we should open an issue for this?

vieux:
LGTM

smothiki:
@aluzzardi  Yeah we should have an issue for open discussion on the implementation of atomic operations . I'll open an Issue

cluster/swarm/cluster.go
smothiki: container struct has only types containerInfo and container . Where as at the time of create container we have only container config and name . To add containerConfig to container struct we have to modify containers interface . It's better to have a PR for open discussion of Implementation details of optimizing atomic operations

PR #546, opened 2015-04-01
use 1024 as total cpushare for a machine

vieux:

### DIFF
--- a/cluster/swarm/node.go
+++ b/cluster/swarm/node.go
@@ -265,7 +265,7 @@ func (n *node) updateContainer(c dockerclient.Container, containers map[string]*
 		}
 		container.Info = *info
 		// real CpuShares -> nb of CPUs
-		container.Info.Config.CpuShares = container.Info.Config.CpuShares / 100.0 * n.Cpus
+		container.Info.Config.CpuShares = container.Info.Config.CpuShares * 1024.0 / n.Cpus
 	}
 
 	return containers, nil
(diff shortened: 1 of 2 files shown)


cluster/swarm/node.go
```diff
@@ -265,7 +265,7 @@ func (n *node) updateContainer(c dockerclient.Container, containers map[string]*
 		}
 		container.Info = *info
 		// real CpuShares -> nb of CPUs
-		container.Info.Config.CpuShares = container.Info.Config.CpuShares / 100.0 * n.Cpus
+		container.Info.Config.CpuShares = container.Info.Config.CpuShares / 1024.0 * n.Cpus
```
smothiki: container.Info.Config.CpuShares results in zero if ```container.Info.Config.CpuShares < 1024```

vieux: @smothiki right, I thought using 1024.0 was enough but as `container.Info.Config.CpuShares` is an int it doesn't work, I'll change thanks.

vieux: @smothiki it should be `container.Info.Config.CpuShares = container.Info.Config.CpuShares * 1024.0 / n.Cpus`

vieux:
@smothiki @abronan @aluzzardi PTAL

smothiki:
Looks good .But only issue is every data type ``int64`` So its very hard to get a fine grained CPU usage from the system . what happens if the system has 1 core ? or some one wants to run a batch process Job that should take very less CPU ? 

Also refresh time is 30 Seconds and what if a scheduler tries to launch a container in between .

aluzzardi:
LGTM

aluzzardi:
@smothiki That, unfortunately, is a limitation of the Docker API and there's nothing we can do about that.

PR #549, opened 2015-04-01
Update token and filter readme

snrism:
Updated the token README to point to the discovery-stage URL. Additionally, fixed link issues to use <token> and not token. 

Do you think it is worth to mention the use of publically reachable IP for POST/DELETE commands in token?


vieux:
LGTM

PR #551, opened 2015-04-01
[eventsHandler] Use Lock/Unlock to sync a write access

noxiouz:
Handle method modifies internal maps, so we have to use Lock/Unlock

### DIFF
--- a/api/events.go
+++ b/api/events.go
@@ -49,11 +49,12 @@ func (eh *eventsHandler) Handle(e *cluster.Event) error {
 		"time", e.Time,
 		"node", cluster.SerializeNode(e.Node))
 
+	var failed []string
+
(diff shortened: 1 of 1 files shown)


aluzzardi:
Not sure this is the right way to go: I'm worried it would cause lots of lock contention since the handler is writing to the network and it might take a while.

Any better idea?

/cc @vieux @abronan

noxiouz:
the modification section was moved out of the reading section

vieux:
LGTM

noxiouz:
Sorry, I accidentally removed the branch. It's restored now

aluzzardi:
LGTM

@vieux ?

vieux:
LGTM

PR #556, opened 2015-04-02
Cleanup: inline index and reuse size vars in random strategy

chanwit:


vieux:
LGTM

Issue #564, opened 2015-04-03
Docker Swarm:  Race condition in etcd discovery


mgoelzer:
In the etcd discovery provider, when you use a short heartbeat interval, a race occurs between the TTL and the heartbeat update causing nodes to briefly disappear from the cluster (due to etcd key expiration) and then reappear.

How to reproduce:

1.  Start swarm join with a short heartbeat interval, e.g., 1 second:

```
swarm join --addr=10.0.1.101:2375 --hb 1 etcd://10.0.1.101:4001/myswarm
```

2.  Note that this will cause ttl to be 1 because of this line:  ```s.ttl = uint64(heartbeat * 3 / 2)``` (in discovery/etcd.go)

3.  Now start an etcdctl watcher:

```etcdctl watch /myswarm --recursive --forever```

About 1/1000 of the time you'll see a blank line because ttl won the race and node disappeared.

Suggested fix:  guarantee TTL will always be substantially longer than heartbeat interval, e.g., s.ttl = max(uint64(heartbeat * 3 / 2), 30).  I can open a PR if maintainers think this is appropriate.

abronan:
@mgoelzer Good catch!

Of course feel free to submit a PR, we'll be there to test and review :)

aluzzardi:
Fixed by https://github.com/docker/swarm/pull/567

PR #567, opened 2015-04-03
Make TTL > heartbeat interval so that keys do not disappear

mgoelzer:
Fixes issue #564 by ensuring that TTL is always greater than heartbeat, even for short heartbeat intervals

The race condition described in #564 is won't have any real world effect as long as etcd is only being used for discovery, since the brief disappearance of a key doesn't harm an already discovered node.  However, I don't see any harm in applying this fix.

Signed-off-by: Mike Goelzer <mike@goelzer.com>


vieux:
LGTM

aluzzardi:
LGTM

PR #574, opened 2015-04-04
Transform `node interface` to `engine struct`

vieux:
Here is the work in progress state about moving the `cluster.Node` interface to a `cluster.Engine` struct.


vieux:
@aluzzardi @abronan This PR is now ready to review.

It's way better this way, I rebase the mesos POC we started with @tnachen on this branch, and as you can see the mesos slave is now much more smaller as it simply extends the engine class https://github.com/vieux/swarm/blob/mesos_test_engine/cluster/mesos/slave.go

abronan:
LGTM, we should merge this as soon as possible (to avoid conflict with future PRs)

cluster/engine.go
```diff
@@ -0,0 +1,556 @@
...
+	"strings"
+	"sync"
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/samalba/dockerclient"
+)
+
+const (
+	// Force-refresh the state of the node this oftee.
```
aluzzardi: of the **engine** this **often**

cluster/engine.go
```diff
@@ -0,0 +1,556 @@
...
+	stateRefreshPeriod = 30 * time.Second
+
+	// Timeout for requests sent out to the node.
```
aluzzardi: s/node/engine/

cluster/engine.go
```diff
@@ -0,0 +1,556 @@
...
+	// Remove the container from the state. Eventually, the state refresh loop
+	// will rewrite this.
+	e.Lock()
+	defer e.Unlock()
+	delete(e.containers, container.Id)
+
+	return nil
+}
+
+// Pull an image on the node
```
aluzzardi: s/node/engine/

cluster/engine_sorter.go
```diff
@@ -0,0 +1,21 @@
+package cluster
+
+// EngineSorter implements the Sort interface to sort Cluster.Node.
```
aluzzardi: cluster.Engine

cluster/engine_sorter.go
```diff
@@ -0,0 +1,21 @@
...
+// It is not guaranteed to be a stable sort.
+type EngineSorter []*Engine
+
+// Len returns the number of nodes to be sorted.
```
aluzzardi: s/nodes/engines/

cluster/engine_sorter.go
```diff
@@ -0,0 +1,21 @@
...
+func (s EngineSorter) Len() int {
+	return len(s)
+}
+
+// Swap exchanges the node elements with indices i and j.
```
aluzzardi: s/node/engine/

cluster/engine_sorter.go
```diff
@@ -0,0 +1,21 @@
...
+func (s EngineSorter) Swap(i, j int) {
+	s[i], s[j] = s[j], s[i]
+}
+
+// Less reports whether the node with index i should sort before the node with index j.
```
aluzzardi: s/node/engine/g

cluster/engine_sorter.go
```diff
@@ -0,0 +1,21 @@
...
+// Nodes are sorted chronologically by name.
```
aluzzardi: s/Nodes/Engines/

cluster/engine_sorter_test.go
```diff
@@ -0,0 +1,18 @@
+package cluster
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNodeSorter(t *testing.T) {
```
aluzzardi: TestEngineSorter

cluster/engine_sorter_test.go
```diff
@@ -0,0 +1,18 @@
...
+	nodes := []*Engine{{Name: "name1"}, {Name: "name3"}, {Name: "name2"}}
```
aluzzardi: s/nodes/engines/

cluster/engine_sorter_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	sort.Sort(EngineSorter(nodes))
```
aluzzardi: s/nodes/engines/

cluster/engine_sorter_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+	assert.Equal(t, nodes[0].Name, "name1")
```
aluzzardi: same

cluster/swarm/cluster.go
```diff
@@ -254,35 +250,47 @@ func (c *Cluster) Container(IDOrName string) *cluster.Container {
 	return nil
 }
 
-// nodes returns all the nodes in the cluster.
-func (c *Cluster) listNodes() []cluster.Node {
+// listNodes returns all the engines in the cluster.
```
aluzzardi: listEngines perhaps?

scheduler/node/node.go
```diff
@@ -0,0 +1,86 @@
+package node
```
aluzzardi: Is this package really necessary? All it contains is a single struct.

Also, if it was called `scheduler.Node` it would perhaps be more clear that this is a scheduler thing (much like if it was called `scheduler.Candidate` or `scheduler.Destination`)

scheduler/node/node.go
```diff
@@ -0,0 +1,86 @@
...
+	UsedMemory  int64
+	UsedCpus    int64
+	TotalMemory int64
+	TotalCpus   int64
+
+	IsHealthy bool
+}
+
+// NewNode creates a node from an engine
+func NewNode(e *cluster.Engine) *Node {
```
aluzzardi: This works great for now.

However, I believe it would cause problems with other cluster drivers since it's Engine (therefore) Swarm specific.

I believe eventually, the engine to node convertion should be handled by Cluster, but for now it's good.

Thoughts?

aluzzardi:
A few comments but overall looks good.

I'm not entirely sold on the `Node` name (even though I suggested it), but I don't have better alternatives. 

`Candidate`, `Destination`, `Placement`?

/cc @vieux @abronan @tnachen

scheduler/node/node.go
vieux: I believe it has to be a package, otherwise you always end up with import cycle.
It's used by cluster/swarm and scheduler. At least I didn't managed to do without it.

scheduler/node/node.go
vieux: It was in the engine at the beginning, but I moved it here because, for now (it might change) we are also doing engine -> node in api/api.go (in `proxyRandom)

scheduler/node/node.go
aluzzardi: Only `cluster/swarm` imports `scheduler`, not the other way around, right?

vieux: yes, but `scheduler` imports the strategies and filters, so we cannot use scheduler.Node in the strategies and filters

vieux: the only way I see (but I don't really like) is to move the strategies and filters into the scheduler package itself. but it fee less clean.

`Node` even though it's a package, it's nested in `scheduler/` so it's not so bad.

abronan:
@aluzzardi the `Node` name looks OK to me. Not sure if changing for another one would help.

scheduler/node/node.go
aluzzardi: Yup

scheduler/node/node.go
aluzzardi: Yes, you're right!

aluzzardi:
LGTM

@vieux: Ready to merge?

Issue #498, opened 2015-03-23
Integration tests for the file discovery backend


crosbymichael:
**Problem**

Integration tests for the file discovery backend

**Solution**

Write integration tests to verify the file discovery backend is working correctly

rot26:
#dibs

groob:
#dibs

chanwit:
@groob any progess about this? If not, would you mind if I take over this one?

groob:
@chanwit go ahead, I ran out of steam.

chanwit:
Thanks @groob ! Review is very welcome.

chanwit:
@abronan @aluzzardi This issue can be closed for now.

PR #581, opened 2015-04-06
integration: test for file discovery

chanwit:
This is an integration test for file discovery. This PR fixes issue #498.

### DIFF
--- a/test/integration/file-discovery.bats
+++ b/test/integration/file-discovery.bats
@@ -1,0 +1,28 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# create a blank temp file for discovery
+DISCOVERY_FILE=$(mktemp)
+
+function teardown() {
+	swarm_manage_cleanup
+	rm -f $DISCOVERY_FILE
+	stop_docker
+}
+
+function setup_file_discovery() {
+	for host in ${HOSTS[@]}; do
+		echo "$host" >> $DISCOVERY_FILE
+	done
+}
+
+@test "docker info should return the number of nodes with file discovery" {
+	start_docker 2
+	setup_file_discovery
+	swarm_manage file://$DISCOVERY_FILE
+
+	run docker_swarm info
+	[ "$status" -eq 0 ]
+	[[ "${lines[3]}" == *"Nodes: 2" ]]
+}


test/integration/discovery.bats
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bats
+
+load helpers
+
+DISCOVERY_FILE=/tmp/test_file_discovery
```
vieux: can we use `mkfile` ?

test/integration/discovery.bats
```diff
@@ -0,0 +1,31 @@
...
+
+function teardown() {
+	stop_manager
+	stop_docker
+	rm -f $DISCOVERY_FILE
+}
+
+function start_manager_with_file_discovery() {
+	rm -f $DISCOVERY_FILE
```
vieux: is this needed here ?

test/integration/discovery.bats
chanwit: `mkfile` is not available everywhere. Maybe I'll use `cat /dev/null > /tmp/test_file_discovery` to create an empty file.

test/integration/discovery.bats
chanwit: I'll make it better.

chanwit:
@vieux PTAL, thank you.

chanwit:
@vieux please hold on.
I should rewrite this after #582 getting merged.

vieux:
@chanwit OK

test/integration/discovery.bats
aluzzardi: Could we use `mktemp` instead?

The goal of the integration tests is to run them on Jenkins or something like that automatically after each merge, so we kinda control the environment.

aluzzardi:
Agreed @chanwit, this will be heavily inspired from #582 once we are done with it - as every integration tests for every discovery service.

Very exciting!

test/integration/discovery.bats
chanwit: I'll try `mktemp` this time.

test/integration/file-discovery.bats
```diff
@@ -0,0 +1,27 @@
...
+
+function teardown() {
+	swarm_manage_cleanup
+	rm $DISCOVERY_FILE
+	stop_docker
+}
+
+function setup_file_discovery() {
+	# create a temp file
+	DISCOVERY_FILE=$(mktemp)
```
aluzzardi: **Maybe** it's better to put this outside of the function.

If the test fails before calling this (`start_docker` fails), `DISCOVERY_FILE` will be undefined and the teardown will `rm` with no arguments.

Not a huge deal.

test/integration/file-discovery.bats
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function teardown() {
+	swarm_manage_cleanup
+	rm $DISCOVERY_FILE
```
aluzzardi: It's probably worth using `rm -f`

test/integration/file-discovery.bats
```diff
@@ -0,0 +1,27 @@
...
+function setup_file_discovery() {
+	# create a temp file
+	DISCOVERY_FILE=$(mktemp)
+	for host in ${HOSTS[@]}; do
+		echo "$host" >> $DISCOVERY_FILE
+	done
+}
+
+@test "docker info should return the number of nodes with file discovery" {
+	start_docker 3
```
aluzzardi: 2 nodes are probably enough for this

aluzzardi:
Minor comments, otherwise LGTM

Thank you @chanwit !

Issue #504, opened 2015-03-23
Add integration tests to the zookeeper discovery backend


crosbymichael:
**Problem**

Add integration tests to the zookeeper discovery backend

**Solution**

Add integration tests to the zookeeper discovery backend

chanwit:
@crosbymichael maybe I could help for this issue.

vieux:
@chanwit go for it :D

chanwit:
@crosbymichael @vieux done. This issue can be now closed :)

abronan:
@chanwit :+1:

PR #582, opened 2015-04-06
integration: test for ZooKeeper discovery

chanwit:
ZooKeeper discovery needs an integration test. This PR fixes issue #504.

/cc @vieux @aluzzardi

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -9,7 +9,7 @@ function teardown() {
 
 @test "docker info should return the number of nodes" {
 	start_docker 3
-	start_manager
+	swarm_manage
 	run docker_swarm info
 	[ "$status" -eq 0 ]
 	[[ "${lines[3]}" == *"Nodes: 3" ]]
@@ -17,7 +17,7 @@ function teardown() {
 
 @test "docker ps -n 3 should return the 3 last containers, including non running one" {
        start_docker 1
-       start_manager
+       swarm_manage
        run docker_swarm run -d busybox sleep 42
        run docker_swarm run -d busybox false
        run docker_swarm ps -n 3
@@ -34,7 +34,7 @@ function teardown() {
 
 @test "docker ps -l should return the last container, including non running one" {
        start_docker 1
-       start_manager
+       swarm_manage
        run docker_swarm run -d busybox sleep 42
        sleep 1 #sleep so the 2 containers don't start at the same second
        run docker_swarm run -d busybox true
--- a/test/integration/helpers.bash
+++ b/test/integration/helpers.bash
@@ -37,15 +37,58 @@ function wait_until_reachable() {
 }
 
 # Start the swarm manager in background.
-function start_manager() {
-	${SWARM_ROOT}/swarm manage -H $SWARM_HOST "$@" `join , ${HOSTS[@]}` &
+function swarm_manage() {
+	local discovery
+	if [ $# -ge 0 ]; then
+		discovery=`join , ${HOSTS[@]}`
+	else
+		discovery="$@"
+	fi
+
+	${SWARM_ROOT}/swarm manage -H $SWARM_HOST $discovery &
 	SWARM_PID=$!
 	wait_until_reachable $SWARM_HOST
 }
 
+# Start swarm join for every engine with the discovery as parameter
+function swarm_join() {
+	local i=0
+	for h in ${HOSTS[@]}; do
+		echo "Swarm join #${i}: $h $@"
+		${SWARM_ROOT}/swarm join --addr=$h "$@" &
+		SWARM_JOIN_PID[$i]=$!
+		((++i))
+	done
+	wait_until_swarm_joined $i
+}
+
+# Wait until a swarm instance joins the cluster.
+# Parameter $1 is number of nodes to check.
+function wait_until_swarm_joined {
+	local attempts=0
+	local max_attempts=10
+
+	until [ $attempts -ge $max_attempts ]; do
+		run docker -H $SWARM_HOST info
+		if [[ "${lines[3]}" == *"Nodes: $1"* ]]; then
+			break
+		fi 
+		echo "Checking if joined successfully for the $((++attempts)) time" >&2
+		sleep 1
+	done
+	[[ $attempts -lt $max_attempts ]]
+}
+
 # Stops the manager.
-function stop_manager() {
+function swarm_manage_cleanup() {
 	kill $SWARM_PID
+}
+
+# Clean up Swarm join processes
+function swarm_join_cleanup() {
+	for pid in ${SWARM_JOIN_PID[@]}; do
+		kill $pid
+	done
 }
 
 # Run the docker CLI against swarm.
--- a/test/integration/main.bats
+++ b/test/integration/main.bats
@@ -3,7 +3,7 @@
 load helpers
 
 @test "test that create doesn't accept any arugments" {
-        run swarm create derpderpderp
+	run swarm create derpderpderp
 	[ "$status" -ne 0 ]
 }
 
--- a/test/integration/zk-discovery.bats
+++ b/test/integration/zk-discovery.bats
@@ -1,0 +1,35 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# Address on which Zookeeper will listen (random port between 7000 and 8000).
+ZK_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 7000 ))
+
+# Container name for integration test
+ZK_CONTAINER_NAME=swarm_integration_zk
+
+function start_zk() {
+	docker run --name $ZK_CONTAINER_NAME -p $ZK_HOST:2181 -d jplock/zookeeper
+}
+
+function stop_zk() {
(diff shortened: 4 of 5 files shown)


vieux:
@chanwit could you please start zookeeper in it's own container, instead of adding a new `Dockerfile` ?

aluzzardi:
Yup - is there a zookeeper container out there that we could use?

Other than that, this PR is much appreciated - thanks @chanwit !

aluzzardi:
Oh, I commented too fast.

So, could we run `docker run swarm join [...]` instead of doing that (as we recommend doing for users)?

I see the issue you were trying to solve - `docker run swarm` would run the release binary instead of the one compiled in.

A few solutions:
- We could simply `docker create swarm join [...]`, then replace the swarm binary inside the container with the built one, and finally `docker start`
- We could `docker build` swarm and then run that image

@vieux @abronan Thoughts?

vieux:
As we are currently using `swarm manage` from the host, we could also run the multiple `swarm join` on the host as well (using the containers IP)

aluzzardi:
Yep, @vieux, that could work.

So today in the helpers we have `start_manager` that starts the manager and does the cleanup automatically etc etc.

I think this should be renamed to `swarm_manage`. Then we could also have `swarm_join`, very similar implementation (actually the two could share some common code) that does a join.

In jours tests then you could do something like:
```
start_docker 3
swarm_manage
swarm_join zk://
```

swarm_join would iterate over `DOCKER_HOSTS` and spawn the according joins.

Thoughts?

chanwit:
@vieux Yes!! That would do the trick and I can get rid of the image building steps.

@aluzzardi I'll try to write function `swarm_join` and I am sure I need your comments for it :)

chanwit:
@vieux I got some hanging when running multiple `swarm join` instances. Would this possible to be a kind of race condition between instances?

Running them through docker does not cause this.

chanwit:
@vieux sorry my bad. `bats` will not stop if some processes are still running.

chanwit:
PTAL @abronan @aluzzardi @vieux

test/integration/api.bats
```diff
@@ -9,7 +9,7 @@ function teardown() {
 
 @test "docker info should return the number of nodes" {
 	start_docker 3
-	start_manager
+	swarm_manage `join , ${HOSTS[@]}`
```
chanwit: @aluzzardi I refactor `swarm_manage` to make it reusable across tests.
Should I leave `join , ${HOSTS[@]}`? Or would you prefer something function `join_hosts`?

aluzzardi: Awesome.

I think it could be something like: You can optionally pass a parameter to `swarm_manage`, in which case it'll use it as the address you want to manage (such as `swarm manage etcd://...`).

By default, if no parameters are passed, it will default to `join , ${HOSTS[@]}`.

I think it's a good tradeoff since only discovery tests will actually want to pass a custom discovery address, all the other tests will be happy with the default one and it'll make them simpler.

test/integration/helpers.bash
```diff
@@ -37,17 +37,28 @@ function wait_until_reachable() {
...
 
+function swarm_join() {
+	for h in ${HOSTS[@]}; do
+		echo "Swarm join $h $@"
+		${SWARM_ROOT}/swarm join --addr=$h "$@" &
+	done
+}
+
 # Stops the manager.
 function stop_manager() {
```
aluzzardi: For sake of consistency, this should probably be something like `swarm_manage_cleanup`?

Since we are not talking about a manager anymore, maybe it makes sense to simply have a `_cleanup` function

test/integration/helpers.bash
```diff
@@ -37,17 +37,28 @@ function wait_until_reachable() {
...
 	kill $SWARM_PID
 }
 
+function stop_swarm_agents() {
```
aluzzardi: Same as above

test/integration/helpers.bash
```diff
@@ -37,17 +37,28 @@ function wait_until_reachable() {
...
+	kill `pidof swarm`
```
aluzzardi: This is a bit dangerous since it will kill everything - the agents, the manager, and other swarms running on the host.

What about instead doing something like keeping an array of `PIDs` and then killing them here?

Similar to what we do with IDs `start_docker`  / `stop_docker`?

aluzzardi:
Awesome work !

A few comments but overall looks good.

test/integration/zk-discovery.bats
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bats
+
+load helpers
+
+ZK_HOST=127.0.0.1:2181
```
aluzzardi: Instead of hardcoding the port, you could do something similar to:
https://github.com/chanwit/swarm/blob/test-zk-discovery/test/integration/helpers.bash#L11

Example:
```bash
# Address on which Zookeeper will listen (random port between 7000 and 8000).
ZK_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 7000 ))
```

test/integration/zk-discovery.bats
```diff
@@ -0,0 +1,32 @@
...
+
+function start_zk() {
+	docker run --name zk -p $ZK_HOST:2181 -d jplock/zookeeper
```
aluzzardi: Instead of hardcoding `zk`, perhaps it should be a variable declared above.

Maybe something longer `swarm_integration_zk` to avoid conflicts?

test/integration/zk-discovery.bats
```diff
@@ -0,0 +1,32 @@
...
+}
+
+@test "zk discovery should be working correctly" {
+	start_zk
+	start_docker 3
+
+	swarm_manage zk://${ZK_HOST}/test
+	swarm_join   zk://${ZK_HOST}/test
+
+ 	sleep 10 # wait for all nodes to join
```
aluzzardi: Is there a way to not sleep so long?

test/integration/zk-discovery.bats
```diff
@@ -0,0 +1,32 @@
...
+function teardown() {
+	stop_swarm_agents
+	stop_manager
+	stop_docker
+	stop_zk
+}
+
+@test "zk discovery should be working correctly" {
+	start_zk
+	start_docker 3
```
aluzzardi: I think just having one or two engines should prove this works, no need for so many (will speed up the test)

test/integration/helpers.bash
```diff
@@ -37,17 +37,28 @@ function wait_until_reachable() {
...
 # Start the swarm manager in background.
-function start_manager() {
-	${SWARM_ROOT}/swarm manage -H $SWARM_HOST "$@" `join , ${HOSTS[@]}` &
+function swarm_manage() {
+	${SWARM_ROOT}/swarm manage -H $SWARM_HOST "$@" &
 	SWARM_PID=$!
 	wait_until_reachable $SWARM_HOST
 }
 
+function swarm_join() {
```
aluzzardi: Can we have a comment on top of this to explain what it does?

test/integration/zk-discovery.bats
aluzzardi: And if we must sleep, maybe we could do it as short as possible, like we do for `wait_until_reachable`: https://github.com/chanwit/swarm/blob/test-zk-discovery/test/integration/helpers.bash#L29

chanwit: May be I need to `docker_swarm info` until I get `Nodes: 3` or timeout then?

test/integration/zk-discovery.bats
chanwit: If you're OK, I'll go with just 1 engine.

test/integration/helpers.bash
chanwit: OK

test/integration/api.bats
chanwit: I see. The default param is great! I'll follow that.

test/integration/helpers.bash
chanwit: Yep

test/integration/helpers.bash
chanwit: Was trying with no success, so I kill pidof swarm. Will give it another try.

chanwit:
ping @aluzzardi

test/integration/main.bats
```diff
@@ -3,7 +3,7 @@
 load helpers
 
 @test "test that create doesn't accept any arugments" {
-        run swarm create derpderpderp
```
aluzzardi: Not sure what changed here (spaces / tabs)?

Also, not entirely sure if we need this test at all - it sure doesn't hurt.

aluzzardi:
Awesome! :+1: 

Thanks for this, we really really needed it.

A very minor comment, otherwise LGTM

test/integration/main.bats
chanwit: Most of whitespace used so far are tabs, but this line somehow contained spaces. My editor just re-formatted it to all tabs.

chanwit: it's the part of #501 added by @technolo-g

PR #583, opened 2015-04-07
Allow multiple IPs in Consul Discovery Service

### LINKED ISSUE #576: Help incorrectly shows ability to specify multiple consul IPs
The help message for Swarm shows that you can specify multiple consul IPs (which would be nice): https://github.com/docker/swarm/blob/master/help.go#L21

However, it doesn't actually (yet): 

`swarm version 0.2.0 (HEAD)`

```
$ /usr/local/bin/swarm manage -H tcp://0.0.0.0:2375 --strategy spread consul://10.0.0.5,10.0.1.5/swarm
INFO[0000] Listening for HTTP                            addr=0.0.0.0:2375 proto=tcp
FATA[0000] Put http://10.0.0.5,10.0.1.5/v1/kv/swarm/: dial tcp: lookup 10.0.0.5,10.0.1.5: no such host
```

Trying the command with each IP on it's own works.

Not sure if it was put there in anticipation of the feature, but it's a bit confusing that it's there but doesn't work.

I saw this PR: https://github.com/docker/swarm/pull/386. Not sure if it is being implemented from someone else.

abronan:
Fixes #576 

Open to discussion, because this is mostly hacking our way around the Go consul client.. Needs some testing as well.

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -1,7 +1,9 @@
 package consul
 
 import (
+	"errors"
 	"fmt"
+	"net/http"
 	"path"
 	"strings"
 	"time"
@@ -11,10 +13,18 @@ import (
 	consul "github.com/hashicorp/consul/api"
 )
 
+var (
+	// ErrConsulNotReachable thrown when no consul agent's api is reachable
+	ErrConsulNotReachable = errors.New("No consul agents reachable")
+	defaultConsulPort     = "8500"
+	defaultWaitTime       = 5
+)
+
 // ConsulDiscoveryService is exported
 type ConsulDiscoveryService struct {
 	heartbeat time.Duration
 	client    *consul.Client
+	agents    string
 	prefix    string
 	lastIndex uint64
 }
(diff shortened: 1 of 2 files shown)


abronan:
Needs some gofmt..

discovery/consul/consul.go
```diff
@@ -103,8 +166,12 @@ func (s *ConsulDiscoveryService) waitForChange() <-chan uint64 {
...
 			_, meta, err := kv.List(s.prefix, option)
 			if err != nil {
-				log.WithField("name", "consul").Errorf("Discovery error: %v", err)
-				break
+				err = s.switchConsulAgent()
+				if err != nil {
+					log.WithField("name", "consul").Errorf("Discovery error: %v", err)
+					break
+				}
+				continue RETRY
```
vieux: I don't see a `for` within a `for` is this label necessary ?

discovery/consul/consul.go
```diff
@@ -25,42 +34,89 @@ func init() {
...
-	p := &consul.KVPair{Key: s.prefix, Value: nil}
-	if _, err = kv.Put(p, nil); err != nil {
-		return err
-	}
-	_, meta, err := kv.Get(s.prefix, nil)
-	if err != nil {
-		return err
-	}
-	s.lastIndex = meta.LastIndex
+	s.prefix = "consul/"
```
vieux: why do you hardcode `consul` before we were able to choose it was better. no ?

discovery/consul/consul.go
abronan: No it's not :D

discovery/consul/consul.go
abronan: Yes, I will put it back to `path` value.

vieux:
@MrMMorris what do you think of this ? 

Looking at the code, I see it's not supported at all by consul (unlike etcd or zookeeper) so I wonder if we should simply fix the usage and only accept one `addr`

vieux:
same question @david-guenault @JeffChien ^

MrMMorris:
@vieux so consul doesn't allow the multiple IPs?

MrMMorris:
it definitely doesn't make sense to only accept one IP as consul should be clustered

vieux:
@MrMMorris to be confirmed by @abronan but apparently your are supposed to give one IP and then use `chit-chat protocol` to find the others

MrMMorris:
@vieux but then what happens when the initial node you reference disappears?

abronan:
To my limited knowledge, consul client only takes the IP of a single consul server (unlike etcd which takes a list of IPs to create the client). I don't know if on node failure Consul knows about the other agents and switches automatically to another agent. I probably missed something but for now either we create one client per consul agent (painful if you have a lot of them) or we switch to another consul agent by creating a new client on failure.

MrMMorris:
Consul client config allows for specifying multiple Consul server IP's. This is my current client conf:

```json
{
  "start_join": [
    "10.0.0.5",
    "10.0.0.6",
    "10.0.1.5"
],
  "disable_remote_exec": true,
  "domain": "consul.",
  "data_dir": "/opt/consul/data",
  "log_level": "INFO",
  "node_name": "comcast-dev",
  "server": false,
  "client_addr": "0.0.0.0",
  "bind_addr": "0.0.0.0",
  "datacenter": "ec2",
  "rejoin_after_leave": true,
  "leave_on_terminate": false
}
```

abronan:
@MrMMorris Which client are you using? For the Go client, if I take a look at https://github.com/hashicorp/consul/blob/master/api/api.go and the `Config` struct there is only a single string field for the consul server address.

MrMMorris:
@abronan not sure. I am downloading from https://dl.bintray.com/mitchellh/consul/0.5.0_linux_amd64.zip

and setting config based on https://www.consul.io/docs/agent/options.html#start_join

abronan:
From the issue hashicorp/consul#849, Consul go client only takes a single consul server's IP thus we need to handle the failover logic ourselves.

ping @vieux @aluzzardi thoughts?

vieux:
it's your call if you are confident with merging this let's go otherwise we only support one IP

abronan:
I'm a bit skeptical about this and I wouldn't merge it as it is. The proper way to use the consul client is clearly described hashicorp/consul#849 using Consul DNS. This PR is mostly a work-around to do the failover ourselves.

abronan:
@MrMMorris What do you think about this? I'd like first to hear about a good use case to avoid using Consul DNS and just point to a set of Consul servers.

MrMMorris:
@abronan I am fine with having to use Consul DNS as I keep finding more reasons to implement it anyways.

abronan:
@MrMMorris OK thanks for the quick followup. I'm closing the PR for now.

PR #584, opened 2015-04-07
[SwarmCluster] Remove extra RLock in node.Container

noxiouz:
```node.Container``` uses an array of containers from ```Containers```, which is
built under an internal RLock, so the external RLock is useless.
Allocate enough memory for the array of containers.


vieux:
ping @aluzzardi @abronan

aluzzardi:
Thanks @noxiouz - could you please rebase?

noxiouz:
done

aluzzardi:
LGTM

Issue #555, opened 2015-04-02
Optimize atomic operations in swarm 


smothiki:
A short term fix for the race conditions was merged in PR https://github.com/docker/swarm/pull/545 .
The main operation that requires atomicity is applying filters to a set of nodes and selecting a node form the filtered list according to the scheduler strategy . Once a node selection is atomic then create container operation on that node need node be locked.

abronan:
Closed by #1261

PR #587, opened 2015-04-07
Proposal : Optimize locking for create container request

smothiki:
Signed-off-by: sivaram mothiki <sivaram@opdemand.com>
fixes https://github.com/docker/swarm/issues/555
As discussed in https://github.com/docker/swarm/pull/545 node selection for scheduling a container is the only operation that has to be atomic in the create container call. We have optimized this by 
1. Defined Scheduled item struct which takes image and container names can be implemented to add more items in the struct w.r.t other filters.
2. Before node.create call add the image and container to the scheduled queue
3. Once the create call is successful remove the scheduled item i,e container along with the image from the queue.
Tested for filter ``affinity:container!=~pulsar-quacking*`` by creating pulsar-quacking_v2.web.4 , pulsar-quacking_v2.web.2 , pulsar-quacking_v2.web.3 containers simultaneously whose images are not present on any host  
Sample Logs 
```
Apr 03 01:10:54 deis-04 sh[17973]: time="2015-04-03T01:10:54Z" level=info msg="HTTP request received" method=POST uri="/v1.17/containers/create?name=pulsar-quacking_v2.web.4"
Apr 03 01:10:54 deis-04 sh[17973]: time="2015-04-03T01:10:54Z" level=info msg="affinity container:" id=0 name=affinity
Apr 03 01:10:54 deis-04 sh[17973]: time="2015-04-03T01:10:54Z" level=info msg="affinity container:pulsar-quacking_v2.web.2" id=1 name=affinity
Apr 03 01:10:54 deis-04 sh[17973]: time="2015-04-03T01:10:54Z" level=info msg="affinity container:" id=0 name=affinity
Apr 03 01:10:54 deis-04 sh[17973]: time="2015-04-03T01:10:54Z" level=info msg="affinity container:pulsar-quacking_v2.web.3" id=1 name=affinity
Apr 03 01:10:54 deis-04 sh[17973]: time="2015-04-03T01:10:54Z" level=info msg="addtoqueue image:172.17.8.103:5000/pulsar-quacking:v2 container:pulsar-quacking_v2.web.4" id="3NQ5:GL5P:WGZV:Z5ZN:ZENF:BSQH:4VFJ:SFIL:MKDO:CRGK:2OUC:JHFY" name=deis-02
``` 
To avoid race condition for pulsar-quacking_v2.web.4 while selecting a node , while applying filters it checks that pulsar-quacking_v2.web.2 , pulsar-quacking_v2.web.3 are already present in the queue. 
```
$docker ps 
2017b63bd0d3        172.17.8.103:5000/pulsar-quacking:v2              "/runner/init start    7 seconds ago       Up 6 seconds            172.17.8.102:49153->5000/tcp                                                      deis-03/pulsar-quacking_v2.web.2   
71e0243d2805        172.17.8.103:5000/pulsar-quacking:v2              "/runner/init start    9 seconds ago       Up Less than a second   172.17.8.100:49153->5000/tcp                                                      deis-01/pulsar-quacking_v2.web.3   
b27cb3ac0ee0        172.17.8.103:5000/pulsar-quacking:v2              "/runner/init start    10 seconds ago      Up 6 seconds            172.17.8.101:49153->5000/tcp                                                      deis-02/pulsar-quacking_v2.web.4   
6bcbf8cfab1f        172.17.8.103:5000/pulsar-quacking:v2              "/runner/init start    2 minutes ago       Up 2 minutes            172.17.8.103:49153->5000/tcp                                                      deis-04/pulsar-quacking_v2.web.1  
```

This is a proposal and it works for affinities and PR is open for discussions and Design suggestions.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"container/list"
 	"crypto/tls"
 	"errors"
 	"fmt"
@@ -30,6 +31,7 @@ func NewEngine(addr string, overcommitRatio float64) *Engine {
 		containers:      make(map[string]*Container),
 		healthy:         true,
 		overcommitRatio: int64(overcommitRatio * 100),
+		ScheduleQueue:   list.New(),
 	}
 	return e
 }
@@ -53,6 +55,14 @@ type Engine struct {
 	eventHandler    EventHandler
 	healthy         bool
 	overcommitRatio int64
+	ScheduleQueue   *list.List
+}
+
+//ScheduledItem FIX ME : works for affinities as it needs the conatainer name and Image
+//Need to add more types to support all filters
+type ScheduledItem struct {
+	Image     string
+	Container string
 }
 
 // Connect will initialize a connection to the Docker daemon running on the
@@ -347,6 +357,23 @@ func (e *Engine) TotalCpus() int64 {
 	return e.Cpus + (e.Cpus * e.overcommitRatio / 100)
 }
 
+//AddtoQueue adds a scheduled item to scheduled queue
+func (e *Engine) AddtoQueue(config *dockerclient.ContainerConfig, name string) {
+	item := &ScheduledItem{
+		Image:     config.Image,
+		Container: name,
+	}
+	e.ScheduleQueue.PushFront(*item)
+}
+
+func (e *Engine) removeFromQueue(config *dockerclient.ContainerConfig, name string) {
+	for i := e.ScheduleQueue.Back(); i != nil; i = i.Prev() {
+		if i.Value.(ScheduledItem).Container == name && i.Value.(ScheduledItem).Image == config.Image {
+			e.ScheduleQueue.Remove(i)
+		}
+	}
+}
+
 // Create a new container
 func (e *Engine) Create(config *dockerclient.ContainerConfig, name string, pullImage bool) (*Container, error) {
 	var (
@@ -356,7 +383,7 @@ func (e *Engine) Create(config *dockerclient.ContainerConfig, name string, pullI
 	)
 
 	newConfig := *config
-
+	defer e.removeFromQueue(config, name)
 	// nb of CPUs -> real CpuShares
 	newConfig.CpuShares = config.CpuShares * 1024 / e.Cpus
 
@@ -378,7 +405,6 @@ func (e *Engine) Create(config *dockerclient.ContainerConfig, name string, pullI
 	// Register the container immediately while waiting for a state refresh.
 	// Force a state refresh to pick up the newly created container.
 	e.refreshContainer(id, true)
-
 	e.RLock()
 	defer e.RUnlock()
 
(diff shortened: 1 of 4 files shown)


vieux:
@smothiki I'm sorry we just merged a huge refactor, can you rebase your PR ? Thanks

smothiki:
okay

smothiki:
Apologies for a wrong commit . Will update the code with the new rebased code .

cluster/engine.go
```diff
@@ -53,6 +55,14 @@ type Engine struct {
 	eventHandler    EventHandler
 	healthy         bool
 	overcommitRatio int64
+	ScheduleQueue   *list.List
+}
+
+//ScheduledItem FIX ME : works for affinities as it needs the conatainer name and Image
+//Need to add more types to support all filters
+type ScheduledItem struct {
```
vieux: I believe the struct `Node` in `scheduler/node` is what you want. (okay it's a big bigger, but it's our "ScheduledItem")

smothiki: ```scheduler/node``` has the list of containers that are already created but not the list of containers that are scheduled but not created.  . I think if we try to inject or add container before its creation it hits back during refreshloop while updating list of containers, as we don't have complete details to make a container struct or Image struct until the container gets created

vieux: yes, I believe you are right

aluzzardi: Ideally (as in, in a perfect world) scheduled containers (not yet created) should be in `Node` and not `Engine`.

Here's the reasoning:
`Engine` is a type representing a Docker Engine, verbatim
`Node` is an internal structure of the scheduler on which it operates on. Right now it's pretty much just an Engine, but that's where we can put data types to be used by the scheduler and the scheduler only (how many failed attempts, bla bla).

Not sure if it's easy to do though.

smothiki: I had thought about keeping ```scheduled Item in node.go``` , but doesn't it create cyclic dependency . As ```node.go imports cluster.Engine``` and again we have to remove the item from the queue once the container is scheduled so inside ```engine.create``` we have to do ```node.removefromqueue```. for which engine has to import scheduler/node .  To enforce that scheduled item should go from Node.go add an Image field and container name field in the [swarm/container.go] (https://github.com/docker/swarm/blob/master/cluster/container.go#L8) so that we can avoid the seperate stucts for scheduling . I'll try updating container.go and see if it works and update the PR

scheduler/filter/affinity.go
```diff
@@ -36,6 +36,9 @@ func (f *AffinityFilter) Filter(config *dockerclient.ContainerConfig, nodes []*n
 				for _, container := range node.Containers {
 					containers = append(containers, container.Id, strings.TrimPrefix(container.Names[0], "/"))
 				}
+				for _, container := range node.ScheduledList("container") {
```
noxiouz: Probably, it's faster
```go
containers = append(containers, node.ScheduledList("container")...)
```

smothiki: yeah , Sorry Bad programming

vieux:
I believe a similar concept was added in https://github.com/docker/swarm/pull/1261

Issue #558, opened 2015-04-02
Make heartbeat type as unsigned int


snrism:
Current heartbeat type is integer and can accept negative values or single digit interval time. This can lead to frequent registration with the manager and can be flaky. To improve this, I believe we can make two changes:

- Make heartbeat type as unsigned
- Define a minimal heartbeat interval (say 10sec). However this can be optional.

Any thoughts if this will be useful?

Thanks,
Sriram

mgoelzer:
Hi @snrism, I agree with you on making it an unsigned int.  It might also be a good idea to require it to be strictly >0, because currently 0 is allowed and I'm not sure if that's a bug or a real use case.

snrism:
@mgoelzer Indeed. I will open a pull request and see what folks think about it.

vieux:
@snrism I agree with @mgoelzer unsigned + > 0

snrism:
@vieux: thanks, I am working on it, will open a PR soon...

PR #588, opened 2015-04-08
Fix heartbeat type and only accept > 0 value

snrism:
This PR addresses the following open issue https://github.com/docker/swarm/issues/558

The changes involve:
+ Updating hearbeat type to be uint64
+ Check if heartbeat value is > 0

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -24,7 +24,7 @@ func init() {
 }
 
 // Initialize is exported
-func (s *ConsulDiscoveryService) Initialize(uris string, heartbeat int) error {
+func (s *ConsulDiscoveryService) Initialize(uris string, heartbeat uint64) error {
 	parts := strings.SplitN(uris, "/", 2)
 	if len(parts) < 2 {
 		return fmt.Errorf("invalid format %q, missing <path>", uris)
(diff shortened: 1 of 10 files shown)


discovery/consul/consul.go
```diff
@@ -40,7 +41,11 @@ func (s *ConsulDiscoveryService) Initialize(uris string, heartbeat int) error {
 		return err
 	}
 	s.client = client
-	s.heartbeat = time.Duration(heartbeat) * time.Second
+	if heartbeat < 1 {
```
vieux: the check is already done in `manage.go` and `join.go` I don't think it's necessary to do it here. (it's a lot of duplicated code in every discovery service)

vieux:
@snrism please run `golint -min_confidence=0.9 ./...` and fix the issues, thanks

discovery/consul/consul.go
snrism: Hmm, sure..will remove that in the discovery service and update it

snrism:
@vieux I removed the check in discovery backends and removed unwanted code. Let me know if this looks OK. 

Thanks,
Sriram

snrism:
@vieux: I combined the commits into a single commit and removed all unwanted code. I did "golint -min_confidence=0.9 ./...", but looks like the CI is failing again. any thoughts?

vieux:
LGTM

vieux:
ping @abronan

abronan:
LGTM

PR #594, opened 2015-04-08
Add Rlocks to swarm cluster

noxiouz:
Add forgotten RLocks to Cluster methods (protect c.engines).
Use WaitGroup to wait for goroutines.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -148,6 +148,9 @@ func (c *Cluster) newEntries(entries []*discovery.Entry) {
 }
 
 func (c *Cluster) hasEngine(addr string) bool {
+	c.RLock()
+	defer c.RUnlock()
+
 	for _, engine := range c.engines {
 		if engine.Addr == addr {
 			return true
@@ -255,7 +258,7 @@ func (c *Cluster) listNodes() []*node.Node {
 	c.RLock()
 	defer c.RUnlock()
 
-	out := []*node.Node{}
+	out := make([]*node.Node, 0, len(c.engines))
 	for _, n := range c.engines {
 		out = append(out, node.NewNode(n))
 	}
@@ -268,7 +271,7 @@ func (c *Cluster) listEngines() []*cluster.Engine {
 	c.RLock()
 	defer c.RUnlock()
 
-	out := []*cluster.Engine{}
+	out := make([]*cluster.Engine, 0, len(c.engines))
 	for _, n := range c.engines {
 		out = append(out, n)
 	}
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -199,10 +199,15 @@ func (c *Cluster) RemoveImage(image *cluster.Image) ([]*dockerclient.ImageDelete
 
 // Pull is exported
 func (c *Cluster) Pull(name string, callback func(what, status string)) {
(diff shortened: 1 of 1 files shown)


cluster/swarm/cluster.go
```diff
@@ -196,10 +199,15 @@ func (c *Cluster) RemoveImage(image *cluster.Image) ([]*dockerclient.ImageDelete
 
 // Pull is exported
 func (c *Cluster) Pull(name string, callback func(what, status string)) {
-	size := len(c.engines)
-	done := make(chan bool, size)
+	var wg sync.WaitGroup
+
+	c.RLock()
 	for _, n := range c.engines {
+		wg.Add(1)
```
aluzzardi: Why not `wg.Add(len(c.engines))` before entering the for loop?

cluster/swarm/cluster.go
```diff
@@ -196,10 +199,15 @@ func (c *Cluster) RemoveImage(image *cluster.Image) ([]*dockerclient.ImageDelete
 
 // Pull is exported
 func (c *Cluster) Pull(name string, callback func(what, status string)) {
-	size := len(c.engines)
-	done := make(chan bool, size)
+	var wg sync.WaitGroup
+
+	c.RLock()
```
aluzzardi: I think this is troublesome - Pull() could last hours, we cannot afford to `RLock` the entire Cluster.

cluster/swarm/cluster.go
```diff
@@ -255,7 +262,7 @@ func (c *Cluster) listNodes() []*node.Node {
 	c.RLock()
 	defer c.RUnlock()
 
-	out := []*node.Node{}
+	out := make([]*node.Node, 0, len(c.engines))
```
aluzzardi: What's the point of optimizing this?

If it's for readability concerns, I guess we could either use the original version, or if we use `make` we shouldn't append but rather do something like `out[i] = n`

cluster/swarm/cluster.go
noxiouz: With wg.Add in the loop the wg is a part of the task distribution. 
If we decided to launch less goroutines then length of c.engines (e.g. via condition), it would lead to a deadlock.
Also it's better to remove a blank line between wg.Add & go func.

The last argument: http://golang.org/pkg/sync/#example_WaitGroup :D

cluster/swarm/cluster.go
noxiouz: Ok, so to prevent the race I copy values to an array under a lock.

cluster/swarm/cluster.go
noxiouz: There's no point  except a performance. If the language provides a mechanism to save memory allocations, it could be used to speedup our code. Also we may avoid creating useless pointers.
I hope go compiler will become smart enough to optimize such cases by itself.

Anyway if it's a code style thing, I can agree.

noxiouz:
Should I fix something?

cluster/swarm/cluster.go
aluzzardi: Yep, sounds good.

cluster/swarm/cluster.go
aluzzardi: Fair enough :)

aluzzardi:
LGTM

Issue #602, opened 2015-04-08
Docker API Parity


aluzzardi:
Umbrella issue for missing API endpoints.

* #585: `docker load`
* #455: `docker build`
* `docker save`

Also, in order for private images to work, we need auth support (#374).

jimmyxian:
@aluzzardi Also, "docker import " is not implement
I will focus on these APIs'implementation.

vieux:
@jimmyxian thanks can you start by adding a PR to update api/README.md to say it's not implemented ? Thanks

vieux:
@jimmyxian i spoke too fast :D

vieux:
@jimmyxian can you please take a look at `docker save` as you did load, it should be quite similar ?

jimmyxian:
@vieux OK, i will be working on it. :)

PR #603, opened 2015-04-09
Proposal: Add build support on a random node

### LINKED ISSUE #455: Add docker build
I'm working on service that allows users to create developer environments. I'm planning to use swarm instead of (or with) self written tool for load balancing.
User should be able to create dev environment by building container from Dockerfile.

vieux:
Fix #455 
Related to #602

```
docker build -t myimage 
docker run -e affinity:image==myimage myimage
```

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -92,7 +92,7 @@ func (e *Engine) connectClient(client dockerclient.Client) error {
 		return err
 	}
 
-	if err := e.refreshImages(); err != nil {
+	if err := e.RefreshImages(); err != nil {
 		e.client = nil
 		return err
 	}
@@ -155,8 +155,8 @@ func (e *Engine) RemoveImage(image *Image) ([]*dockerclient.ImageDelete, error) 
 	return e.client.RemoveImage(image.Id)
 }
 
-// Refresh the list of images on the engine.
-func (e *Engine) refreshImages() error {
+// RefreshImages refreshes the list of images on the engine.
+func (e *Engine) RefreshImages() error {
 	images, err := e.client.ListImages()
 	if err != nil {
 		return err
@@ -275,7 +275,7 @@ func (e *Engine) refreshLoop() {
 		}
 
 		if err == nil {
-			err = e.refreshImages()
+			err = e.RefreshImages()
 		}
 
 		if err != nil {
@@ -490,7 +490,7 @@ func (e *Engine) handler(ev *dockerclient.Event, _ chan error, args ...interface
 	case "pull", "untag", "delete":
 		// These events refer to images so there's no need to update
 		// containers.
-		e.refreshImages()
+		e.RefreshImages()
 	case "start", "die":
 		// If the container is started or stopped, we have to do an inspect in
 		// order to get the new NetworkSettings.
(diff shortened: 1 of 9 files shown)


garagatyi:
I think that build mostly will be used to create containers, not pushing it somewhere. Thats why it doesn't make sense to create it on overloaded nodes, where containers can't be started.

vieux:
@garagatyi yes, you can still do 
```
docker build -t user/myimage 
docker push user/myimage
docker run user/myimage ...
```

vieux:
following @aluzzardi idea, "random" nos uses the filters and strategy selected. build will likely select the best node that match your needs.

aluzzardi:
LGTM

aluzzardi:
@garagatyi: How about now?

cluster/swarm/cluster.go
```diff
@@ -295,3 +295,15 @@ func (c *Cluster) Info() [][2]string {
 
 	return info
 }
+
+// RANDOMENGINE returns a random engine.
+func (c *Cluster) RANDOMENGINE() (*cluster.Engine, error) {
```
abronan: Nit-picking on the name but why keeping **random*** if it defaults to the strategy used? (can be something like *PickEngine* or else?)

abronan:
@aluzzardi way better than random :D

aluzzardi:
@vieux: If possible, would it be too much of a hassle to add an integration test?

Eventually, before 0.3, it would be awesome to have an integration test for every single API endpoint so we can confidently say "we are 100% compatible with the Docker API".

Creating them as we add API endpoints will be less of a pain later :)

garagatyi:
@vieux thanks for the improvement
@aluzzardi LGTM now

vieux:
ping @aluzzardi

aluzzardi:
@vieux I thought you were adding integration tests for this one before the review?

aluzzardi:
Also, there's one commit that doesn't seem to pass tests (6961b54) - could you squash it with something else?

Ideally every single commit should build / pass tests. It will make our lives much easier should we need to git bisect someday.

vieux:
@aluzzardi PTAL

cluster/engine.go
```diff
@@ -156,7 +156,7 @@ func (e *Engine) RemoveImage(image *Image) ([]*dockerclient.ImageDelete, error)
 }
 
 // Refresh the list of images on the engine.
```
aluzzardi: golint is complaining about this

aluzzardi:
Any way to know whether the image was built "manually"?

I'm wondering if `run` could work without the need of affinity - otherwise, we won't be "compatible" with the docker API since `docker build && docker run` will fail (CI system for example).

vieux:
rebased

vieux:
@aluzzardi as we discussed it should be fixed in docker itself, I'll open another issue.

aluzzardi:
Yep, looks good!

Last item: Could you please create an issue to refactor `build`?

Ideally we would add a `cluster.Build` function (along with `Engine.Build` and support in `dockerclient`), in the same way it was done for `load`: https://github.com/docker/swarm/pull/609

Issue #585, opened 2015-04-07
Idea of "docker load" api implementation


jimmyxian:
I want to implement the 'docker load' api in swarm, the following steps is my idea to do this:

* (1) swarm cache the load-file(save in a temp file)\[api/api.go\]
* (2) call cluster.Load(temp_file_path)
	* Need to add a "func Load(file_path string)" interface in cluster[cluster/cluster.go]
* (3) In swarm cluster,  implement Load interface, send load request to every node[cluster/swarm/cluster.go]
    * call node.load(temp_file_path)
* (4) In node.load function, call n.client.LoadImage(temp_file_path)[cluster/swarm/node.go]
    * Need to add and implement LoadImage interface in dockerclient
	
Well, first i will submit a PR(add and implement LoadImage interface) in dockerclient, then implement "docker load" api in swarm.

Is my idea right ?

vieux:
I think so, as we decided that `docker pull` pulls on every node, I think it's a good idea to have `docker load` load on every machine.

jimmyxian:
@vieux OK. I will do it

aluzzardi:
Fixed by @jimmyxian

Issue #492, opened 2015-03-23
Integration tests for API


jfrazelle:
Right now, the API is not being covered by unit tests since it's hard to properly test. Integration tests could be written covering every supported CLI command.

Currently, we are missing the following tests:
- [x] docker attach
- [x] docker build
- [x] docker commit
- [x] docker cp
- [x] docker create
- [x] docker diff
- [x] docker events
- [x] docker exec
- [x] docker export
- [x] docker history
- [x] docker images
- [ ] docker import
- [x] docker info
- [x] docker inspect
- [x] docker kill
- [x] docker load
- [ ] docker login
- [ ] docker logout
- [x] docker logs
- [x] docker port
- [x] docker pause
- [x] docker ps
- [x] docker pull
- [ ] docker push
- [x] docker rename
- [x] docker restart
- [x] docker rm
- [x] docker rmi
- [x] docker run
- [x] docker save
- [x] docker search
- [x] docker start
- [x] docker stats
- [x] docker stop
- [x] docker tag
- [x] docker top
- [x] docker unpause
- [x] docker version
- [x] docker wait

Help very appreciated as it's a colossal amount of work!

dpwspoon:
https://zapier.com/engineering/continuous-integration-jenkins-docker-github/

aluzzardi:
We are missing the following integration tests (defined in `api.bats`):

```
ocker attach
docker build
docker commit
docker cp
docker create
docker diff
docker events
docker exec
docker export
docker history
docker images
docker import
docker inspect
docker kill
docker load
docker login
docker logout
docker logs
docker port
docker pause
docker pull
docker push
docker rename
docker restart
docker rm
docker rmi
docker run
docker save
docker search
docker start
docker stats
docker stop
docker tag
docker top
docker unpause
docker version
docker wait
```

Help very appreciated as it's a colossal amount of work!

chanwit:
:)

chanwit:
Could we have a proper checklist? So works can be done individually and randomly.

aluzzardi:
Sure, @chanwit, sounds like a good idea!

jimmyxian:
@jfrazelle  @aluzzardi 
I will be working on it.

aluzzardi:
Thanks @jimmyxian 

Do you want to work on all of them or just a few ones? I can check the boxes on the ones you'd like to tackle to mark them as assigned

jimmyxian:
[...]

PR #609, opened 2015-04-10
Add support "docker load" api

jimmyxian:
issue:https://github.com/docker/swarm/issues/585
Tested locally and works
```
[root@localhost home]# docker -H 127.0.0.1:2375 load -i /home/busybox.tar
[root@localhost home]# 
[root@localhost home]# 
[root@localhost home]# docker -H 127.0.0.1:2375 load -i /home/code/test.go 
host81:Loading Image... 500 Internal Server Error: Untar exit status 1 archive/tar: invalid tar header
host80:Loading Image... 500 Internal Server Error: Untar exit status 1 archive/tar: invalid tar header
```


abronan:
LGTM

cc @aluzzardi @vieux

abronan:
Just came over this with @vieux and the output must be similar to the pull command (not spawning another line to say the image is downloaded or loaded):

```
[root@localhost home]# docker -H 127.0.0.1:2375 load < /home/ubuntu.tar 
host81:Loading Image... : loaded
host80:Loading Image... : loaded
```

If you can conform to this output, that would be perfect.

aluzzardi:
Is there a way to do this in streaming rather than two steps (download tar / upload tar)?

aluzzardi:
As asked in #603: would it be possible to add an integration test? Eventually we are going to test every single API endpoint and doing that as we add endpoints will be less of a pain later.

You could probably wait for @vieux to add an integration test for `docker build` in #603 and take inspiration from that.

jimmyxian:
@abronan OK, i will follow the pull command. Also, the docker api "load"  itself output nothing if success, i don't know whether it'OK to output the successful message?
@aluzzardi i'm trying to write the every api integration test locally with bats. i think all the apis test will come soon.  :)

aluzzardi:
@jimmyxian That is freaking awesome :)

How about the streaming vs downloading locally? Do you think it's doable?

jimmyxian:
@aluzzardi yeah.streaming is quit good. i think it's doable.  i will do it by one producer muti consumer model.... do it this in this PR？or a new one

aluzzardi:
@jimmyxian If you could squeeze it in this PR it would be fantastic.

I think we already have multiple examples in the code base, for logs and events (@vieux correct me if I'm wrong)

jimmyxian:
@aluzzardi @vieux @abronan 
Streaming is Done.  The docker api "load" itself output nothing if success. For swarm, also output nothing if successful, is it OK?

vieux:
@jimmyxian yes it's fine.

aluzzardi:
Could you please rebase?

jimmyxian:
@aluzzardi Done to rebase.

cluster/swarm/cluster.go
```diff
@@ -231,6 +232,57 @@ func (c *Cluster) Pull(name string, callback func(what, status string)) {
 	}
 }
 
+// Load image
+func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)) {
+	size := len(c.engines)
+	done := make(chan bool, size)
```
aluzzardi: Would you mind using a `WaitGroup`, like in `Pull()` (we just changed that because apparently it's a best practice):

https://github.com/docker/swarm/blob/master/cluster/swarm/cluster.go#L213

test/integration/docker-load.bats
```diff
@@ -0,0 +1,30 @@
...
+# temp file for saving image
+IMAGE_FILE=$(mktemp)
+
+function teardown() {
+	stop_docker
+	swarm_manage_cleanup
+	rm -f $IMAGE_FILE
+}
+
+@test "docker load should return success,every node should load the image" {
```
aluzzardi: Could you move this into `api.bats`?

aluzzardi:
Looks great, some minor comments, thanks!

This may be impossible, but do you think tests for `cluster.Load` could be added? The overall coverage went down by 1.07% with this PR.

aluzzardi:
Integration test related to #492

jimmyxian:
ping @aluzzardi
Used WaitGroup and moved the test in api.bats
Also, I will think about the cluster.Load test and commit a new PR about this

vieux:
@jimmyxian would you mind adding the test in this PR ? We try to not merge the test and the code at the same time.

jimmyxian:
@vieux OK. i will do that :)

aluzzardi:
Opened #672 to track the missing test

PR #612, opened 2015-04-11
fix #589 - allow pipe in constraint regexp

### LINKED ISSUE #589: Regexp filtering limitation
# What I do

I have 5 docker server in the pool of my swarm manager.
Two servers are started with the following label : 
  * server1 : ̀`type=app`
  * server2 : `type=db`
 
I then want to start a container on any nodes with types `app` **or** `db` : 

```bash
docker run -d -e constraint:type==/app\|db/ nginx
```

# What I get

I get an the following error on client side : 
```bash
FATA[0000] Error response from daemon: Value '/app|db/' is invalid
```

And this one on proxy side : 
```bash
time="2015-04-08T11:15:38+02:00" level=info msg="HTTP request received" method=POST uri="/v1.16/containers/create" 
time="2015-04-08T11:15:38+02:00" level=error msg="HTTP error: Value '/app|db/' is invalid" status=500 
```

# What I expected

The [scheduler README](https://github.com/docker/swarm/blob/master/scheduler/filter/README.md) says [Go regexp are supported](https://github.com/docker/swarm/blame/master/scheduler/filter/README.md#L174), so I was expecting the container to be created on any of these nodes.

I can see this is because of the constraint value [validation here](https://github.com/docker/swarm/blob/cdc9c7a47b0267dac77a170fed72a5bd26a752a4/scheduler/filter/expr.go#L55)

Swarm should not limit us when building complex filtering strategies, **the full golang regexp behaviour should be available**

chanwit:
This PR fixes missing pipe ( | ) for the `or` operation in constraint expressions, reported in #589.

### DIFF
--- a/scheduler/filter/expr.go
+++ b/scheduler/filter/expr.go
@@ -52,7 +52,7 @@ func parseExprs(key string, env []string) ([]expr, error) {
 						// allow leading = in case of using ==
 						// allow * for globbing
 						// allow regexp
-						matched, err := regexp.MatchString(`^(?i)[=!\/]?(~)?[a-z0-9:\-_\s\.\*/\(\)\?\+\[\]\\\^\$]+$`, parts[1])
+						matched, err := regexp.MatchString(`^(?i)[=!\/]?(~)?[a-z0-9:\-_\s\.\*/\(\)\?\+\[\]\\\^\$\|]+$`, parts[1])
 						if err != nil {
 							return nil, err
 						}
--- a/scheduler/filter/expr_test.go
+++ b/scheduler/filter/expr_test.go
@@ -32,7 +32,7 @@ func TestParseExprs(t *testing.T) {
 	assert.NoError(t, err)
 
 	// Allow regexp in value
-	_, err = parseExprs("constraint", []string{"constraint:node==/(?i)^[a-b]+c*$/"})
+	_, err = parseExprs("constraint", []string{"constraint:node==/(?i)^[a-b]+c*(n|b)$/"})
 	assert.NoError(t, err)
 
 	// Allow space in value


chanwit:
@frntn could you please confirm that this fixes your issue.

frntn:
Ok for me.

I have successfully build a new swarm binary with your repo/branch
No error messages at container deployment using pipe `|` 
The constraint is applied as expected

test/integration/issue-589-constraint.bats
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+}
+
+@test "issue 589 node constraint regexp should support pipe" {
```
abronan: Could you remove the reference to the issue in the file and the test and put everything into `constraints.bats`? I think the test description is explicit enough and it belongs here.

test/integration/issue-589-constraint.bats
```diff
@@ -0,0 +1,23 @@
...
+	# start node-0 and node-1
+	start_docker 2
+	swarm_manage
+
+	# Reproduce issue 589
```
chanwit: @abronan also removing this comment?

abronan: You can leave the comment for now for easy reference :)

chanwit:
ping @abronan @aluzzardi

abronan:
LGTM

aluzzardi:
I think this should probably be tested in unit tests rather than integration tests.

What do you think @chanwit ?

chanwit:
@aluzzardi I got both unit and integration tests in this PR.
Unit test is here: https://github.com/docker/swarm/pull/612/files#diff-6c36ccfba3d706f6271fe995f77ffa72R35

I am OK to remove the integration test.

aluzzardi:
Yep, but that only tests if it actually parses correctly, not if it works.

Although I agree that testing Go's regex library might not be useful :)

So LGTM

PR #620, opened 2015-04-15
add create/start/stop/restart/kill integration test

jimmyxian:
add create/start/stop/restart/kill integration test

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -38,9 +38,17 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker create" {
-	skip
+	start_docker 3
+	swarm_manage
+
+	# create
+	run docker_swarm create --name test_container busybox sleep 1000
+        [ "$status" -eq 0 ]
+
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"test_container"* ]]
 }
 
 # FIXME
@@ -91,9 +99,20 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker kill" {
-	skip
+	start_docker 3
+	swarm_manage
+	# run 
+	run docker_swarm run -d --name test_container busybox sleep 1000
+	[ "$status" -eq 0 ]
+
+	# kill
+	run docker_swarm kill test_container
+	[ "$status" -eq 0 ]
+	# verify
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" == *"Exited"* ]]
 }
 
 # FIXME
@@ -177,9 +196,25 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker restart" {
-	skip
+	start_docker 3
+	swarm_manage
+	# run 
+	run docker_swarm run -d --name test_container busybox sleep 1000
+	[ "$status" -eq 0 ]
+
+	# make sure container is up
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" == *"Up"* ]]
+
+	# restart
+	run docker_swarm restart test_container
+	[ "$status" -eq 0 ]
+	# verify
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" == *"Up"* ]]
 }
 
 # FIXME
@@ -207,9 +242,19 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker start" {
-	skip
+	start_docker 3 
+	swarm_manage
+	# create
+	run docker_swarm create --name test_container busybox sleep 1000
+	[ "$status" -eq 0 ]
+
+	# start
+	run docker_swarm start test_container
+	[ "$status" -eq 0 ]
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"Up"* ]]
 }
 
 # FIXME
@@ -217,9 +262,19 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker stop" {
-	skip
+	start_docker 3
+	swarm_manage
+	# run 
+	run docker_swarm run -d --name test_container busybox sleep 500
+	[ "$status" -eq 0 ]
+
+	# stop
+	run docker_swarm stop test_container
+	[ "$status" -eq 0 ]
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" == *"Exited"* ]]
 }
 
 # FIXME
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -42,10 +42,15 @@ function teardown() {
 	start_docker 3
 	swarm_manage
 
+	# make sure no contaienr exist
+	run docker_swarm ps -qa
+	[ "${#lines[@]}" -eq 0 ]
+
 	# create
 	run docker_swarm create --name test_container busybox sleep 1000
         [ "$status" -eq 0 ]
 
+	# verify, created container exists
 	run docker_swarm ps -l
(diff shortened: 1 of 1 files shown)


test/integration/api.bats
```diff
@@ -48,3 +48,50 @@ function teardown() {
        [ "${#lines[@]}" -eq  2 ]
        [[ "${lines[1]}" == *"false"* ]]
 }
+
+
+@test "docker create/start/stop/restart/kill should return success" {
+	start_docker 1
```
vieux: can you use more than one daemon, like 3
so we are sure that the calls go to the right node

abronan: @vieux 2 might be enough then right?

vieux: yes but I prefer 3 :p

abronan: :ghost:

test/integration/api.bats
```diff
@@ -48,3 +48,50 @@ function teardown() {
...
+	swarm_manage
+	
+	#create
+	run docker_swarm create --name test_container busybox sleep 1000
+	[ "$status" -eq 0 ]
+	run docker_swarm ps -l
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${#lines[1]}" == *"test_container"* ]]
```
abronan: You need to remove the `#` in front of `lines[1]` (for all the following tests too)

abronan:
@jimmyxian Test fails on `[[ "${#lines[1]}" == *"test_container"* ]]` because of the comment above.

test/integration/api.bats
```diff
@@ -48,3 +48,50 @@ function teardown() {
...
        [[ "${lines[1]}" == *"false"* ]]
 }
+
+
+@test "docker create/start/stop/restart/kill should return success" {
+	start_docker 1
+	swarm_manage
+	
+	#create
+	run docker_swarm create --name test_container busybox sleep 1000
```
chanwit: This should be `run docker_swarm create --name test_container busybox sh` to make the container run forever.

chanwit:
This PR relates to #492.

aluzzardi:
Thanks @jimmyxian, much appreciated!

I've created a skeleton in #628 - basically created an integration test for every API call we are missing (the implementation right now is `skip`).

Could you please rebase this on top of #628 once we merge it?

jimmyxian:
@aluzzardi OK, i will have done half of the apis integration test. If all test ok locally, i will subbmit it.

chanwit:
:+1: 
@jimmyxian please submit it at small amounts. It's difficult to review all of them at a time :)

jimmyxian:
@chanwit OK. :)

aluzzardi:
Thanks @jimmyxian !

As @chanwit said, would it be possible to split it up into one test per "action" (create, start, ...)?

In `api.bats` I've created empty tests (they only do skip) for every single command the CLI supports.

Ideally, could you move each one of those tests into its own test function?

Agreed that for now it's not very useful, but each of those tests is going to grow: For `create`, we might want to test with different flags and see if they work. For `start`, `stop`, `restart`, we might want to do a `docker ps` and make sure the status got updated.

jimmyxian:
@aluzzardi OK, Thks for your advice.I will follow this.

test/integration/api.bats
```diff
@@ -38,9 +38,17 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker create" {
-	skip
+	start_docker 3
+	swarm_manage
+
```
aluzzardi: Could you do a `ps -qa` and check that `lines -eq 0` before doing the create?

test/integration/api.bats
```diff
@@ -91,9 +99,20 @@ function teardown() {
...
 
-# FIXME
 @test "docker kill" {
-	skip
+	start_docker 3
+	swarm_manage
+	# run 
+	run docker_swarm run -d --name test_container busybox sleep 1000
+	[ "$status" -eq 0 ]
+
```
aluzzardi: Could you do a `ps -l` here and check the container is running before the kill?

test/integration/api.bats
```diff
@@ -207,9 +242,19 @@ function teardown() {
...
 
-# FIXME
 @test "docker start" {
-	skip
+	start_docker 3 
+	swarm_manage
+	# create
+	run docker_swarm create --name test_container busybox sleep 1000
+	[ "$status" -eq 0 ]
+
```
aluzzardi: Could you check that the container is stopped here?

test/integration/api.bats
```diff
@@ -217,9 +262,19 @@ function teardown() {
...
 
-# FIXME
 @test "docker stop" {
-	skip
+	start_docker 3
+	swarm_manage
+	# run 
+	run docker_swarm run -d --name test_container busybox sleep 500
+	[ "$status" -eq 0 ]
+
```
aluzzardi: Check if container is running here

aluzzardi:
Few comments but looks great!

chanwit:
@jimmyxian This looks promising! But as @aluzzardi pointed out, every test should have its exit criteria  rather that only checking status codes.

For example, when calling `docker stop`, we need to ensure that the container is really stopped by querying the engine again.

jimmyxian:
@aluzzardi  OK, thanks for your advice..I will modify it.

test/integration/api.bats
jimmyxian: After creating the container, the status of the new container is null. I will check the lines of "docker ps -l" and check the container's name

jimmyxian:
ping @aluzzardi Modified it

test/integration/api.bats
```diff
@@ -91,9 +104,27 @@ function teardown() {
...
+	[[ "${lines[1]}" == *"test_container"* ]]
+	[[ "${lines[1]}" == *"Up"* ]]
+
+	# kill
+	run docker_swarm kill test_container
+	[ "$status" -eq 0 ]
+	# verify
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+        [[ "${lines[1]}" == *"test_container"* ]]
```
vieux: can you please fix the indent here ?

test/integration/api.bats
```diff
@@ -177,9 +208,27 @@ function teardown() {
...
+	[[ "${lines[1]}" == *"test_container"* ]]
+	[[ "${lines[1]}" == *"Up"* ]]
+
+	# restart
+	run docker_swarm restart test_container
+	[ "$status" -eq 0 ]
+	# verify
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+        [[ "${lines[1]}" == *"test_container"* ]]
```
vieux: same here

test/integration/api.bats
```diff
@@ -207,9 +256,26 @@ function teardown() {
...
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"test_container"* ]]
+
+	# start
+	run docker_swarm start test_container
+	[ "$status" -eq 0 ]
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+        [[ "${lines[1]}" == *"test_container"* ]]
```
vieux: and here

jimmyxian: @vieux ...ok

jimmyxian:
ping @vieux  Done to fix the format

test/integration/api.bats
aluzzardi: Can you verify if that's the case with docker as well?

You might have found another bug in swarm!

aluzzardi:
One very last comment - otherwise LGTM

aluzzardi:
Hey, for some reason my comment is now showing up here.

I added a last comment to `docker start`

test/integration/api.bats
jimmyxian: OK, i will check it. :)

jimmyxian: @aluzzardi Yeah, I checked it. The docker is same as that.

jimmyxian:
@aluzzardi Yeah, I checked it. After creating the container, the status of the new container is null.

aluzzardi:
Awesome! Thanks!

LGTM

PR #622, opened 2015-04-15
Remove prefixes from DiscoveryService implementations

KushalP:
All of these structs are public. Previously when they were used, they'd
look like the following:

 - `consul.ConsulDiscoveryService`
 - `etcd.EtcdDiscoveryService`
 - `file.FileDiscoveryService`
 - `nodes.NodesDiscoveryService`
 - `token.TokenDiscoveryService`
 - `zookeeper.ZkDiscoveryService`

As a result, they stuttered when read. This commit removes the
duplication of the package names on the struct so that they read better.

See also: https://golang.org/ref/spec#Import_declarations

Signed-off-by: Kushal Pisavadia <kushi.p@gmail.com>

### DIFF
--- a/discovery/consul/consul.go
+++ b/discovery/consul/consul.go
@@ -11,8 +11,8 @@ import (
 	consul "github.com/hashicorp/consul/api"
 )
 
-// ConsulDiscoveryService is exported
-type ConsulDiscoveryService struct {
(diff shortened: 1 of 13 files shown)


LK4D4:
LGTM

abronan:
LGTM

PR #623, opened 2015-04-15
Fix golint errors

aluzzardi:


vieux:
@aluzzardi I believe the only issue was `NewEventsHandler` as `eventsHandler` doesn't need to be exported.

Can you revert the change and unexport `NewEventsHandler` ?

vieux:
LGTM

PR #626, opened 2015-04-16
Coverage improvements

aluzzardi:
- Make script/coverage runnable locally.
- Move the goveralls push to travis instead of script/coverage.
- Refactor code into a loop.
- Properly handle packages with no tests.
- Exclude API from coverage.


vieux:
LGTM

PR #627, opened 2015-04-16
API Cleanup

aluzzardi:
Some refactoring around the API to split up `api.go` (one of the biggest files of swarm) into sub components.


aluzzardi:
@vieux @abronan: Thoughts? This is a proposal, feel free to reject if the split doesn't make sense.

vieux:
LGTM

abronan:
LGTM

aluzzardi:
Rebased

PR #629, opened 2015-04-16
stress test: First short at stress testing Swarm in an automated way.

aluzzardi:

### DIFF
--- a/test/stress/stress.bats
+++ b/test/stress/stress.bats
@@ -1,0 +1,86 @@
+#!/usr/bin/env bats
+
+load ../integration/helpers
+
+NODES=10
+CONTAINERS=100
+
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+}
+
+# Start N containers on the cluster in parallel that will sleep randomly.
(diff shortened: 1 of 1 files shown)


chanwit:
How many node swarm should be able to handle properly before getting 1.0?
Do you have some numbers in mind?

ehazlett:
:+1: nice!  i'll test this out :)

aluzzardi:
@chanwit 100-ish

abronan:
Current status: 300 containers on 30 nodes, slowly ramping up :smile:

aluzzardi:
Yeah, I went up to 1000 containers on 100 nodes but started having hiccups :)

abronan:
@aluzzardi My machine started having hiccups way before the 100 nodes milestone :sob: 

It's a great first shot and gives an idea of Swarm is capable of. Would be great if in the future, we could distribute the load between multiple machines/docker daemons for this stress tests.

abronan:
Anyway, LGTM

vieux:
LGTM

PR #586, opened 2015-04-07
Implement docker attach-ws api

jimmyxian:
Signed-off-by: Xian Chaobo <xianchaobo@huawei.com>

PR #636, opened 2015-04-17
Add attach-ws api integration test

jimmyxian:
https://github.com/docker/swarm/pull/586 Add test for attach-ws api
I build a docker image with websocket CLI(WSSH) wirtten in Nodejs to test attatch-ws api. if connected successfull, it returns two lines.
```
Session Open
Session Closed
```

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -12,6 +12,26 @@ function teardown() {
 	skip
 }
 
+@test "docker attach through websocket" {
+	CLIENT_API_VERSION="v1.17"
+	start_docker 2
+	swarm_manage
+
+	#create a container
+        run docker_swarm run -d --name test_container busybox sleep 1000
+
+	# test attach-ws api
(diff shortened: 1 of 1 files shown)


test/integration/api.bats
```diff
@@ -12,6 +12,26 @@ function teardown() {
 	skip
 }
 
+@test "docker attach through websocket" {
+	CLIENT_API_VERSION="v1.17"
```
abronan: We should probably expose this at a higher level, in `helpers.bats`? /cc @aluzzardi

abronan:
Test passes successfully and LGTM. Just wondering if we should enforce the `CLIENT_API_VERSION` at a higher level.

/cc @aluzzardi

jimmyxian:
@abronan @aluzzardi  Yeah, i think it's good for us to maintain after moving it in helpers.bats. any advice?

aluzzardi:
I think it's okay to keep it there for now - only that test is using it.

LGTM

aluzzardi:
Related to #492

PR #637, opened 2015-04-17
Add commit/exec integration test

jimmyxian:
Add commit/exec integration test

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -28,9 +28,31 @@ function teardown() {
       [ "${#lines[@]}" -eq 1 ]
 }
 
-# FIXME
 @test "docker commit" {
-	skip
+	start_docker 3
+	swarm_manage
+
+	run docker_swarm run -d --name test_container busybox sleep 500
+	[ "$status" -eq 0 ]
+
+	# make sure container exists
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"test_container"* ]]
+
+	# no comming name before commit 
+	run docker_swarm images
+	[ "$status" -eq 0 ]
+	[[ "${output}" != *"commit_image_busybox"* ]]
+
+	# commit container
+	run docker_swarm commit test_container commit_image_busybox
+	[ "$status" -eq 0 ]
+
+	# verify after commit 
+	run docker_swarm images
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"commit_image_busybox"* ]]
 }
 
 # FIXME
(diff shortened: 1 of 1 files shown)


test/integration/api.bats
```diff
@@ -13,24 +13,39 @@ function teardown() {
 }
 
 @test "docker build" {
-      start_docker 3
```
aluzzardi: Watch out for tabs / spaces - I believe we are using tabs in `*.bats` files and your editor replaced them by spaces.

test/integration/api.bats
```diff
@@ -13,24 +13,39 @@ function teardown() {
...
+
+	run docker_swarm run -d --name test_container busybox sleep 500
+	[ "$status" -eq 0 ]
+
+	# commit container
+	run docker_swarm commit test_container commit_image_busybox
+	[ "$status" -eq 0 ]
+	
+	# cluster refresh image during 30 seconds        
+	sleep 35
```
aluzzardi: I think that's a problem on Swarm that your integration test found.

We should force an image refresh after a commit (like we do for build).

test/integration/api.bats
```diff
@@ -172,9 +198,24 @@ function teardown() {
...
+
+	run docker_swarm run -d --name test_container busybox sleep 500
+	[ "$status" -eq 0 ]
+
+	# rename container
+	run docker_swarm rename test_container rename_container
+	[ "$status" -eq 0 ]
+
+	# cluster refresh state of container during 30 seconds	
+	sleep 35
```
aluzzardi: Same here - we should refresh

test/integration/api.bats
```diff
@@ -172,9 +198,24 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker rename" {
```
aluzzardi: Actually, I think we don't really support renaming in Swarm (this is a new feature).

We should definitely upstate the `Store` with the new name.

test/integration/api.bats
jimmyxian: Yeah, we should update the images after commit\pull\tag\load... I will check it and fix this problem.

test/integration/api.bats

test/integration/api.bats
jimmyxian: I submit a PR to force refresh images when commit/tag/pull #650

test/integration/api.bats
jimmyxian: Actually, I think we should also force refresh container when rename, what do you think about this?

jimmyxian:
@aluzzardi 
Removed rename test And use exec -i

jimmyxian:
ping @aluzzardi

aluzzardi:
LGTM

PR #638, opened 2015-04-17
Simplify code using gofmt's simplify command

KushalP:
This code was passed through `gofmt -s`, which simplifies code. The tool
removed the blank identifiers from these `for range` blocks.

More information: https://golang.org/cmd/gofmt/#hdr-The_simplify_command

Signed-off-by: Kushal Pisavadia <kushi.p@gmail.com>


KushalP:
Apologies. It looks like this PR only works on Go 1.4. I'll close this PR.

Issue #625, opened 2015-04-16
Get Node Info via API


snrism:
Is there a way to get the node info via API? I would like to get the "docker info" output of individual nodes in the cluster (this is required so that I can read the standard filters and user-defined labels to schedule the containers)? If yes, any pointer is appreciated. If this is an open issue, I am happy to send a PR.

Thanks,
Sriram

snrism:
I guess, by gathering the label values from engine Info and appending this to the cluster info, we can get this information in "docker info" and via the /info api call. I will send a PR for this.

PR #639, opened 2015-04-17
Add Node Labels to Info 

snrism:
This pull request addresses https://github.com/docker/swarm/issues/625

The goal is to include the label values in the "/info" API. This is helpful for an external application which interacts with swarm via REST APIs to get the constraint values from each host and to schedule the containers based on labels.

Just to highlight: This update however changes the "docker info" output to show as following:

Containers: 16
Strategy: spread
Nodes: 3
 host1: 192.168.111.78:2375
  └ Containers: 5
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 2.053 GiB
  └ Labels: storagedriver=aufs, executiondriver=native-0.2, kernelversion=3.13.0-39-generic, operatingsystem=Ubuntu 14.04.1 LTS, service=security
 host2: 192.168.111.79:2375
  └ Containers: 5
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 2.053 GiB
  └ Labels: storagedriver=aufs, executiondriver=native-0.2, kernelversion=3.13.0-39-generic, operatingsystem=Ubuntu 14.04.1 LTS, service=perf
 host3: 192.168.111.80:2375
  └ Containers: 6
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 2.053 GiB
  └ Labels: storagedriver=aufs, executiondriver=native-0.2, kernelversion=3.13.0-39-generic, operatingsystem=Ubuntu 14.04.1 LTS, service=monitor

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"sort"
+	"strings"
 	"sync"
 
 	log "github.com/Sirupsen/logrus"
@@ -310,6 +311,12 @@ func (c *Cluster) Info() [][2]string {
 		info = append(info, [2]string{" └ Containers", fmt.Sprintf("%d", len(engine.Containers()))})
 		info = append(info, [2]string{" └ Reserved CPUs", fmt.Sprintf("%d / %d", engine.UsedCpus(), engine.TotalCpus())})
 		info = append(info, [2]string{" └ Reserved Memory", fmt.Sprintf("%s / %s", units.BytesSize(float64(engine.UsedMemory())), units.BytesSize(float64(engine.TotalMemory())))})
+		labels := make([]string, 0, len(engine.Labels))
+		for k, v := range engine.Labels {
+			labels = append(labels, k+"="+v)
+		}
+		sort.Strings(labels)
+		info = append(info, [2]string{" └ Labels", fmt.Sprintf("%s", strings.Join(labels, ", "))})
 	}
 
 	return info
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -114,11 +114,12 @@ function teardown() {
 }
 
 @test "docker info" {
-	start_docker 3
+	start_docker 1 --label foo=bar
 	swarm_manage
 	run docker_swarm info
 	[ "$status" -eq 0 ]
-	[[ "${lines[3]}" == *"Nodes: 3" ]]
+	[[ "${lines[3]}" == *"Nodes: 1" ]]
+	[[ "${output}" == *"└ Labels:"*"foo=bar"* ]]
 }
 
 # FIXME


vieux:
Hey @snrism can you fix the golint ? Thanks

snrism:
@vieux fixed that. thanks.

cluster/engine.go
```diff
@@ -347,6 +347,15 @@ func (e *Engine) TotalCpus() int64 {
 	return e.Cpus + (e.Cpus * e.overcommitRatio / 100)
 }
 
+// NodeLabels returns the assigned and standard labels
+func (e *Engine) NodeLabels() string {
+	labels := []string{}
```
noxiouz: I don't want to be annoying :D, but 
```go
labels = make([]string, 0, len(e.Labels)
```
It will accelerate this method on 25-30% depending of the amount of keys.

snrism: @noxiouz I initially did that and while updating the code changed it.. Lemme make that. Thanks for your suggestion

aluzzardi:
Overall intent looks good.

However, I wouldn't be putting `NodeLabels` inside Engine - it's a pretty print utility and won't benefit anything but `docker info`.

How about putting that code directly into `Info` or another helper function of `Cluster`?

@abronan @vieux: Thoughts?

snrism:
@aluzzardi agreed it makes sense to move it within cluster..once agreed, I can update the code to parse the labels inside the Info function (before appending the info). Thanks!

aluzzardi:
@vieux @abronan What do you think about my previous comment?

I know we're kinda already doing this for Filters() in the scheduler (returning a docker info friendly output)

abronan:
Agreed, it makes more sense to have it directly into `Info`. Not sure if we need an helper function for this. We can just move the code of `NodeLabels` into Info. It is concise enough.

snrism:
@aluzzardi @abronan Cleaned up the code to add the label info directly inside the Info function. Hope this works..thanks

cluster/swarm/cluster.go
```diff
@@ -310,6 +311,11 @@ func (c *Cluster) Info() [][2]string {
 		info = append(info, [2]string{" └ Containers", fmt.Sprintf("%d", len(engine.Containers()))})
 		info = append(info, [2]string{" └ Reserved CPUs", fmt.Sprintf("%d / %d", engine.UsedCpus(), engine.TotalCpus())})
 		info = append(info, [2]string{" └ Reserved Memory", fmt.Sprintf("%s / %s", units.BytesSize(float64(engine.UsedMemory())), units.BytesSize(float64(engine.TotalMemory())))})
+		labels := make([]string, 0, len(engine.Labels))
+		for k, v := range engine.Labels {
+			labels = append(labels, k+"="+v)
+		}
```
abronan: Would be great if you can add `sort.Strings(labels)` here to have a sorted output for better readability (with a lot of entries it can be difficult to locate a label)

snrism: Done. Thanks for your suggestion.

abronan:
LGTM

cc/ @aluzzardi

aluzzardi:
@snrism Awesome!

Would you mind adding this to our integration tests?
https://github.com/docker/swarm/blob/master/test/integration/api.bats

You can pass additional parameters to `start_docker` (such as `--label`), then you can check in `$output` if you find those labels.

snrism:
@aluzzardi Added the label info check in api integration tests. Does that look OK? Thanks.

aluzzardi:
Very much indeed! LGTM

Thanks @snrism !

PR #642, opened 2015-04-19
Cleanup: Move CLI related files to a subpackage.

aluzzardi:
Bonus points: `godep go run main.go` works.

Integration tests could actually use `go run` instead of assuming the swarm binary was already compiled.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -1,0 +1,123 @@
+package cli
+
+import (
+	"fmt"
+	"os"
+	"path"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/codegangsta/cli"
+	"github.com/docker/swarm/discovery"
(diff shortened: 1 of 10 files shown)


aluzzardi:
Well, since it was a tiny change, the integration tests actually do use go run now.

/cc @chanwit @jimmyxian

aluzzardi:
Removed the CLI from test coverage - just like the API, it's going to be tested in integration.

chanwit:
This is great @aluzzardi. It's now guarantee that integration tests will be executed against the current codes!

jimmyxian:
Great ! @aluzzardi

vieux:
we could have use `godep go run *.go` otherwise LGTM

abronan:
LGTM

PR #646, opened 2015-04-19
Cleanup: cast int64 to make it works everywhere

chanwit:


vieux:
@chanwit where do you see tests failing without this fix ?

chanwit:
@vieux on my ARM cluster, which is not 64-bit.

vieux:
LGTM

abronan:
LGTM

PR #649, opened 2015-04-20
Add attach/diff/port/top integration test

jimmyxian:
Add attach/diff/port/top integration test


vieux:
LGTM

PR #651, opened 2015-04-20
add dockerfile and scripts to make buildable and testable on jenkins

jfrazelle:
This doesn't work yet but the goal would be to have it run the integration tests w docker in docker.

you can try with:
```
docker build --rm --force-rm -f Dockerfile.test -t swarmtest .
docker run --rm -it --privileged swarmtest script/testall
```

I think you all will know better how the bats test needs the daemon started so I am opening this now, so hopefully you all will be able to spot the fix easily :)

### DIFF
--- a/Dockerfile.test
+++ b/Dockerfile.test
@@ -1,0 +1,58 @@
+# Dockerfile for testing swarm
+# How to use:
+#   docker build --rm --force-rm -f Dockerfile.test -t swarm-test .
+#   docker run --rm -it --privileged swarm-test script/testall
+FROM debian:jessie
+
+# compile and runtime deps
+# https://github.com/docker/docker/blob/master/project/PACKAGERS.md#build-dependencies
+# https://github.com/docker/docker/blob/master/project/PACKAGERS.md#runtime-dependencies
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        ca-certificates \
+        curl \
+        e2fsprogs \
+        gcc \
+        git \
+        iptables \
+        libsqlite3-dev \
+        make \
+        mercurial \
+        procps \
+        xz-utils \
+    && rm -rf /var/lib/apt/lists/*
+
+# install golang
+ENV GO_VERSION 1.4.2
+RUN curl -sSL https://golang.org/dl/go${GO_VERSION}.src.tar.gz | tar -v -C /usr/local -xz \
+    && mkdir -p /go/bin
+ENV PATH /go/bin:/usr/local/go/bin:$PATH
+ENV GOPATH /go
+RUN cd /usr/local/go/src && ./make.bash --no-clean 2>&1
+
+# install bats
+RUN cd /usr/local/src/ \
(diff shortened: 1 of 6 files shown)


jfrazelle:
actually i think i might have it now

jfrazelle:
nm idk

Dockerfile.test
```diff
@@ -0,0 +1,58 @@
+# Dockerfile for testing swarm
+# How to use:
+#   docker build --rm --force-rm -f Dockerfile.test -t swarm-test .
+#   docker run --rm -it --privileged swarm-test script/testall
+FROM debian:jessie
```
aluzzardi: How about starting from `aluzzardi/docker` so it comes pre-installed with docker and no need to install its dependencies?

Dockerfile.test
```diff
@@ -0,0 +1,58 @@
+# Dockerfile for testing swarm
```
aluzzardi: I think this Dockerfile could be located in `test/integration` (and simply named `Dockerfile`) and take care of only running integration tests (we are still using travis for unit)

Dockerfile.test
```diff
@@ -0,0 +1,58 @@
...
+ENV GOPATH /go
+RUN cd /usr/local/go/src && ./make.bash --no-clean 2>&1
+
+# install bats
+RUN cd /usr/local/src/ \
+    && git clone https://github.com/sstephenson/bats.git \
+    && cd bats \
+    && ./install.sh /usr/local
+
+# install go tools
```
aluzzardi: These dependencies wouldn't be needed if we only run integration

script/dind
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
```
aluzzardi: If we start from a docker image, this would be unnecessary

script/integration-test
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -e
+
+DEST=$1
+
+# subshell so that we can export PATH without breaking other things
+(
+	source "$(dirname "$BASH_SOURCE")/.integration-daemon-start")
```
aluzzardi: If I'm not mistaken, only one daemon per host can be run, right?

In that case, couldn't just we assume that docker is already running and available at `DOCKER_HOST` and not start it ourselves?

script/integration-test
```diff
@@ -0,0 +1,25 @@
...
+	
+	# we need to wrap up everything in between integration-daemon-start and
```
aluzzardi: I **think** you could use `trap`: It calls some code before exiting the script (even if it was Ctrl-C'ed)

script/integration-test
aluzzardi: If this only runs integration tests (without starting the daemon), I think we could move it to `test/integration/run.sh` - not sure about that though.

Thoughts?

aluzzardi:
Thanks a lot for this!

aluzzardi:
Replaced by #696 (although heavily inspired by this).

PR #652, opened 2015-04-21
Add pause/unpause/inspect integration test

jimmyxian:
Add pause/unpause/inspect integration test

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -87,7 +87,6 @@ function teardown() {
 }
 
 @test "docker inspect" {
-skip
 	start_docker 3
 	swarm_manage
 	# run container
@@ -103,10 +102,12 @@ skip
 	run docker_swarm inspect test_container
 	[ "$status" -eq 0 ]
 	[[ "${lines[1]}" == *"AppArmorProfile"* ]]
+	# the specific information of swarm node
+	[[ ${output} == *'"Node": {'* ]]
+	[[ ${output} == *'"Name": "node-'* ]]
 }
 
 @test "docker inspect --format" {
-skip
 	start_docker 3
 	swarm_manage
 	# run container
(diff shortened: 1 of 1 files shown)


abronan:
Weirdly, `pause` cannot complete in my setup. It makes the `dind` container go into the `<defunct>` state and the test stays blocked forever (probably on `docker unpause`).

/cc @aluzzardi @vieux

abronan:
Does it work for you @jimmyxian?

aluzzardi:
LGTM once @abronan issue is solved

aluzzardi:
Related to #492

test/integration/api.bats
```diff
@@ -86,9 +86,41 @@ function teardown() {
 	[[ "${lines[3]}" == *"Nodes: 3" ]]
 }
 
-# FIXME
 @test "docker inspect" {
-	skip
+	start_docker 3
```
aluzzardi: @vieux: Should we check the `Node` field that we add to `docker inspect` here?

test/integration/api.bats
```diff
@@ -121,9 +153,27 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker pause" {
-	skip
+	start_docker 3
```
aluzzardi: LGTM

Please check if it works on your machine as @abronan suggested.

test/integration/api.bats
```diff
@@ -232,9 +282,36 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker unpause" {
-	skip
+	start_docker 3
```
aluzzardi: LGTM

jimmyxian:
@abronan @aluzzardi @vieux 
Yeah, It's a bug. If the state of the container is paused, it can't be removed(`rm -f`), so it block in stop_docker. So three ways to solve:
+ upause it after paused sucessfully
+ modify stop_docker in helpers.bash: unpause all the paused container before `rm -f`
+ fix it in docker: `rm -f` should remove container in any states

abronan:
@jimmyxian Imho for now the better solution would be to `pause` and then `unpause` the container for `stop_docker` to complete inside the `docker pause` test. As we are using `pause` only in this test, we don't want to have an extra check for paused containers after each set of tests.

Regarding the fix in docker, it's a well known issue that ghosted containers cannot be cleaned up properly. So just adding the `unpause` to the test will do for now.

Just tested and it works fine, If you could add this to the existing test that would be great! Thanks!

aluzzardi:
Agree with @abronan 

Also, maybe we should add a check of `Node` in inspect - I'll let @vieux decide.

test/integration/api.bats
vieux: yes very good idea!

ew should look for a `  "Node": {` object and a maybe `"Name": "node-"` as all the node start with `node-`

jimmyxian:
@abronan @aluzzardi @vieux  thks
Add `unpause` in pause test and check node infomation in `inspect`

abronan:
@jimmyxian Perfect. Thanks!

LGTM

PR #654, opened 2015-04-21
Add images/pull/tag integration test

jimmyxian:
Add images/pull/tag integration test

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -136,9 +136,68 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker images" {
-	skip
+	start_docker 3
+	swarm_manage
+
+	# no image exist
+	run docker_swarm images 
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -eq 1 ]
+	[[ "${lines[0]}" == *"REPOSITORY"* ]]
+
+	# pull image
+	run docker_swarm pull busybox
+	[ "$status" -eq 0 ]
+
+	# show all node images, including reduplicated
+	run docker_swarm images
+	[ "$status" -eq 0 ]
+	# check pull busybox, if download sucessfully, the busybox have one tag(lastest) at least
+	# if there are 3 nodes, the output lines of "docker images" are greater or equal 4(1 header + 3 busybox:latest)
+	# The following(pull/tag) is the same reason.
+	[ "${#lines[@]}" -ge 4 ]
+	# Every line should contain "busybox" exclude the first head line 
+	for((i=1; i<${#lines[@]}; i++)); do
+		[[ "${lines[i]}" == *"busybox"* ]]
+	done
+}
+
+@test "docker images --filter node=<Node name>" {
+	start_docker 3
+	swarm_manage
+
+	# no image exist
+	run docker_swarm images 
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -eq 1 ]
+	[[ "${lines[0]}" == *"REPOSITORY"* ]]
+
+	# make sure every node has no image
+	for((i=0;i<3;i++)); do
+		run docker_swarm images --filter node=node-$i
+		[ "$status" -eq 0 ]
+		[ "${#lines[@]}" -eq 1 ]
+		[[ "${lines[0]}" == *"REPOSITORY"* ]]
+	done
+
+	# pull image
+	run docker_swarm pull busybox
+	[ "$status" -eq 0 ]
+
+	# make sure images have pulled
+	run docker_swarm images
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -ge 4 ]
+	[[ "${lines[1]}" == *"busybox"* ]]
+
+	# verify
+	for((i=0; i<3; i++)); do 
+		run docker_swarm images --filter node=node-$i
+		[ "$status" -eq 0 ]
+		[ "${#lines[@]}" -ge 2 ]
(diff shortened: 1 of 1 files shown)


abronan:
`tag` test fails on my setup on `[[ "${lines[*]}" == *"tag_busybox"* ]]`. Not sure why though.. Is it the proper way to test on multiple lines output? Shouldn't we `grep tag_busybox` from the output instead?

/ping @aluzzardi

vieux:
@jimmyxian @abronan we could `docker images tag_busybox`

test/integration/api.bats
```diff
@@ -222,9 +312,30 @@ function teardown() {
...
+	[ "${#lines[@]}" -ge 2 ]
+	[[ "${lines[*]}" == *"busybox"* ]]
+	[[ "${lines[*]}" != *"tag_busybox"* ]]
+
+	# tag image
+	run docker_swarm tag busybox tag_busybox:test
+	[ "$status" -eq 0 ]
+
+	# verify
+	run docker_swarm images
```
vieux: please use `run docker_swarm images tag_busybox`

abronan:
@vieux +1. Reduces the output and makes the test easier.

test/integration/api.bats
```diff
@@ -222,9 +312,30 @@ function teardown() {
...
+
+	run docker_swarm pull busybox
+	[ "$status" -eq 0 ]
+
+	# make sure the image of busybox exists 
+	# the comming image of tag_busybox not exsit
+	run docker_swarm images
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -ge 2 ]
+	[[ "${lines[*]}" == *"busybox"* ]]
```
aluzzardi: Please use `$output` instead of `${lines[*]}`

aluzzardi:
Related to #492

test/integration/api.bats
```diff
@@ -68,9 +68,68 @@ function teardown() {
...
 }
 
-# FIXME
 @test "docker images" {
-	skip
+	start_docker 3
+	swarm_manage
+
+	# no image exist
+	run docker_swarm images
```
aluzzardi: I think here you could run `images -q` and simply check that the output has 0 lines

test/integration/api.bats
```diff
@@ -68,9 +68,68 @@ function teardown() {
...
+
+	# no image exist
+	run docker_swarm images 
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -eq 1 ]
+	[[ "${lines[0]}" == *"REPOSITORY"* ]]
+
+	# make sure every node has no image
+	for((i=0;i<3;i++)); do
+		run docker_swarm images --filter node=node-$i
```
aluzzardi: `images -q`

test/integration/api.bats
```diff
@@ -68,9 +68,68 @@ function teardown() {
...
+	# if there are 3 nodes, the output lines of "docker images" are greater or equal 4(1 header + 3 busybox:latest)
+	# so use -ge here, the following(pull/tag) is the same reason
+	[ "${#lines[@]}" -ge 4 ]
+	# Every line should contain "busybox" exclude the first head line 
+	for((i=1; i<${#lines[@]}; i++)); do
+		[[ "${lines[i]}" == *"busybox"* ]]
+	done
+}
+
+@test "docker images --filter node=<Node name>" {
```
aluzzardi: I think the two `images` tests can be combined

test/integration/api.bats
```diff
@@ -162,9 +221,40 @@ function teardown() {
...
+	for host in ${HOSTS[@]}; do
+		run docker -H $host images -q
+		[ "$status" -eq 0 ]
+		[ "${#lines[@]}" -eq 0 ]
+	done
+
+	run docker_swarm pull busybox
+	[ "$status" -eq 0 ]
+
+	# swarm verify
```
aluzzardi: This is already checked by `docker images` - I think the node verify is enough

test/integration/api.bats

test/integration/api.bats

jimmyxian:
@aluzzardi @abronan @vieux Thks
Aleady modifid by the comments

abronan:
`tag` is still failing for me.

If I echo the `output` during the test I get:

```
REPOSITORY TAG IMAGE ID CREATED VIRTUAL SIZE
busybox latest 4986bf8c1536 3 months ago 2.433 MB
busybox latest 4986bf8c1536 3 months ago 2.433 MB
busybox latest 4986bf8c1536 3 months ago 2.433 MB
```

There is no tagged image so I wonder if there is actually a bug with `tag`.. Will check on a proper cluster.

abronan:
Ok the test fails because it immediately issues a `docker images` right after the `docker tag` but the state is not refreshed so the output is the one I gave above: without the tagged image.

Just tested on a real cluster, if I `tag` an image and I do a `docker images` directly after, the `tag` does not appear in the output.

aluzzardi:
It's probably the same issue we had with `build` and a few others, we need to force the refresh.

abronan:
Ok this works great after rebasing off of Master. @jimmyxian could you please rebase so I can check again and confirm this works as expected. Thanks!

jimmyxian:
@abronan My bad.Not rebase correctly yesterday
Done to Rebase Again. Thks

abronan:
@jimmyxian No worries, this was me having a hard time to figure out where was the problem as I missed the PR with the force-refresh. Double-checked and works fine. Thanks again!

LGTM

/ping @aluzzardi

aluzzardi:
LGTM

PR #663, opened 2015-04-23
Add events/search/stats/version/wait integration test

jimmyxian:
Add events/search/stats/version/wait integration test

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -772,9 +772,8 @@ function teardown() {
 	start_docker 3
 	swarm_manage
 
-	TEMP_NAME=noimage$RANDOM
 	# search image (not exist), the name of images only [a-z0-9-_.] are allowed
-	run docker_swarm search $TEMP_NAME
+	run docker_swarm search does_not_exist
 	[ "$status" -eq 0 ]
 	[ "${#lines[@]}" -eq 1 ]
 	[[ "${lines[0]}" == *"DESCRIPTION"* ]]
@@ -830,9 +829,13 @@ function teardown() {
 	# storage the real time output of stats in TEMP_FILE
 	docker_swarm stats test_container > $TEMP_FILE &
 
+	local attempts=0
+	local max_attempts=5
 	# make sure TEMP_FILE exists and is not empty
-	sleep 3
-	[ -s $TEMP_FILE ]
+	until [ -s $TEMP_FILE ] || [ $attempts -ge $max_attempts ]; do
+		sleep 0.5
+		attempts=$[$attempts+1]
+	done
 
 	# if "CPU %" in TEMP_FILE, status is 0
 	run grep "CPU %" $TEMP_FILE
@@ -958,11 +961,12 @@ function teardown() {
 }
 
 @test "docker wait" {
+	TEMP_FILE=$(mktemp)
 	start_docker 3
 	swarm_manage
-	
-	# run after 10 seconds, test_container will exit
-	run docker_swarm run -d --name test_container busybox sleep 10
+
+	# run after 3 seconds, test_container will exit
+	run docker_swarm run -d --name test_container busybox sleep 3
 	[ "$status" -eq 0 ]
 
 	# make sure container exists and is up
@@ -971,9 +975,20 @@ function teardown() {
 	[[ "${lines[1]}" ==  *"test_container"* ]]
 	[[ "${lines[1]}" ==  *"Up"* ]]
 
-	# wait until exist(after 10 seconds)
-	run docker_swarm wait test_container
+	# wait until exist(after 3 seconds)
+	docker_swarm wait test_container > $TEMP_FILE &
+	WAIT_PID=$!
+
+	# after 5 secondes, cat $TEMP_FILE      
+	sleep 5
+	run cat $TEMP_FILE
 	[ "$status" -eq 0 ]
 	[ "${#lines[@]}" -eq 1 ]
 	[[ "${lines[0]}" == "0" ]]
+
+	# after sucess, if the wait process still exist then kill
+	run ps -p $WAIT_PID
+	if [ "$status" -eq 0 ]; then
+		kill -9 $WAIT_PID
+	fi
 }
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -955,14 +955,13 @@ function teardown() {
 	[ "${#lines[@]}" -ge 8 ]
 
 	# verify
-	client_reg='^Client version: [0-9]+\.[0-9]+\.[0-9]+$'
-	server_reg='^Server version: swarm/[0-9]+\.[0-9]+\.[0-9]+$'
+	client_reg='^Client version: [0-9]+\.[0-9]+\.[0-9]+.*$'
+	server_reg='^Server version: swarm/[0-9]+\.[0-9]+\.[0-9]+.*$'
 	[[ "${lines[0]}" =~ $client_reg ]]
 	[[ "${lines[5]}" =~ $server_reg ]]
 }
 
 @test "docker wait" {
-	TEMP_FILE=$(mktemp)
 	start_docker 3
 	swarm_manage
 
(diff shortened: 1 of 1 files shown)


test/integration/api.bats
```diff
@@ -423,9 +443,27 @@ function teardown() {
...
 
-# FIXME
 @test "docker search" {
-	skip
+	start_docker 3
+	swarm_manage
+
+	TEMP_NAME=noimage$RANDOM
+	# search image (not exist), the name of images only [a-z0-9-_.] are allowed
+	run docker_swarm search $TEMP_NAME
```
aluzzardi: I think this is a bit overkill - could you please `docker_swarm search does_not_exist`?

test/integration/api.bats
```diff
@@ -450,9 +488,32 @@ function teardown() {
...
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" == *"test_container"* ]]
+	[[ "${lines[1]}" == *"Up"* ]]
+
+	# storage the real time output of stats in TEMP_FILE
+	docker_swarm stats test_container > $TEMP_FILE &
+
+	# make sure TEMP_FILE exists and is not empty
+	sleep 3
```
aluzzardi: No sleeps like this please - could you do a loop that retries and sleeps a little bit between tries? Like in `wait_until_reachable`

test/integration/api.bats
```diff
@@ -508,12 +569,39 @@ function teardown() {
 	skip
 }
 
-# FIXME
 @test "docker version" {
-	skip
+	start_docker 3
```
aluzzardi: LGTM

test/integration/api.bats
```diff
@@ -508,12 +569,39 @@ function teardown() {
...
 }
 
-# FIXME
 @test "docker wait" {
-	skip
+	start_docker 3
+	swarm_manage
+	
+	# run after 10 seconds, test_container will exit
+	run docker_swarm run -d --name test_container busybox sleep 10
```
aluzzardi: 10 is way too long (we don't want to slow down the tests).

3 seconds should be plenty

test/integration/api.bats
```diff
@@ -508,12 +569,39 @@ function teardown() {
...
+	[ "$status" -eq 0 ]
+
+	# make sure container exists and is up
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"test_container"* ]]
+	[[ "${lines[1]}" ==  *"Up"* ]]
+
+	# wait until exist(after 10 seconds)
+	run docker_swarm wait test_container
```
aluzzardi: The problem here is that if `wait` doesn't work, this will be stuck forever.

How about using the `timeout` tool?

```
run docker_swarm run -d --name test_container busybox sleep 3
[...]
run timeout 5 docker_swarm wait test_container
[ "$status" -eq 0 ]
```

test/integration/api.bats
```diff
@@ -116,9 +116,29 @@ function teardown() {
 	[[ "${lines[*]}" ==  *"diff.txt"* ]]
 }
 
-# FIXME
 @test "docker events" {
-	skip
+	TEMP_FILE=$(mktemp)
```
aluzzardi: @vieux Do we want to check the node here in events?

test/integration/api.bats

test/integration/api.bats
jimmyxian: Yeah

jimmyxian:
@aluzzardi  Thanks for your advice. ( `timeout 5s + COMMAND` ), when I use `timeout 5 + docker_swarm ...`, return an error `"No such file or directory"`. So achieve the same effect ( to kill the blocked process )  in another way.

test/integration/api.bats
vieux: Yes. @jimmyxian can you start the container on a specific node and check the that right node is inside the event ?

jimmyxian:
ping @aluzzardi

aluzzardi:
Everything looks good here besides that `sleep 5`.

`timeout` is part of core utils and should probably be in your system (on my Ubuntu it's `/usr/bin/timeout`)

test/integration/api.bats
```diff
@@ -450,9 +491,36 @@ function teardown() {
...
+	[[ "${lines[1]}" == *"test_container"* ]]
+	[[ "${lines[1]}" == *"Up"* ]]
+
+	# storage the real time output of stats in TEMP_FILE
+	docker_swarm stats test_container > $TEMP_FILE &
+
+	local attempts=0
+	local max_attempts=5
+	# make sure TEMP_FILE exists and is not empty
+	until [ -s $TEMP_FILE ] || [ $attempts -ge $max_attempts ]; do
```
aluzzardi: You can make use of "retry" here: #693

test/integration/api.bats
```diff
@@ -508,12 +576,51 @@ function teardown() {
...
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"test_container"* ]]
+	[[ "${lines[1]}" ==  *"Up"* ]]
+
+	# wait until exist(after 3 seconds)
+	docker_swarm wait test_container > $TEMP_FILE &
+	WAIT_PID=$!
+
+	# after 5 secondes, cat $TEMP_FILE      
+	sleep 5
```
aluzzardi: Here you could use `retry` as well from #693

maybe something like `retry 10 0.5 cat $TEMP_FILE` (will try 10 times every 0.5 seconds, so 5 seconds total)

test/integration/api.bats
```diff
@@ -508,12 +576,51 @@ function teardown() {
...
 
-# FIXME
 @test "docker wait" {
-	skip
+	TEMP_FILE=$(mktemp)
+	start_docker 3
+	swarm_manage
+
+	# run after 3 seconds, test_container will exit
+	run docker_swarm run -d --name test_container busybox sleep 3
```
aluzzardi: Maybe `sleep 1` is enough?

test/integration/api.bats
aluzzardi: Also, I believe we shouldn't be doing it like that.

As I mentioned, a `timeout` might work.

jimmyxian: Yeah. I'm doing on it.

aluzzardi:
Please rebase - that will fix existing problems already fixed by #713

jimmyxian:
ping @aluzzardi 
Updated by the comments and check again, THKS

test/integration/api.bats
```diff
@@ -880,12 +945,43 @@ function teardown() {
...
+	[ "$status" -eq 0 ]
+
+	# make sure container exists and is up
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"test_container"* ]]
+	[[ "${lines[1]}" ==  *"Up"* ]]
+
+	# wait until exist(after 1 seconds)
+	timeout 5 docker -H $SWARM_HOST wait test_container > $TEMP_FILE
```
aluzzardi: I think you can `run timeout 5 ...` and just check `$output`, no need to redirect

test/integration/api.bats
```diff
@@ -880,12 +945,43 @@ function teardown() {
...
+	start_docker 3
+	swarm_manage
+
+	# version
+	run docker_swarm version
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -ge 8 ]
+
+	# verify
+	client_reg='^Client version: [0-9]+\.[0-9]+\.[0-9]+$'
```
aluzzardi: This is failing on Jenkins.

I believe because we use docker master, and the version is `1.7.0-dev`.

aluzzardi:
Related to #492

test/integration/api.bats
```diff
@@ -880,12 +945,39 @@ function teardown() {
...
+	# make sure container exists and is up
+	run docker_swarm ps -l
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" ==  *"test_container"* ]]
+	[[ "${lines[1]}" ==  *"Up"* ]]
+
+	# wait until exist(after 1 seconds)
+	run timeout 5 docker -H $SWARM_HOST wait test_container
+
+	[ "${#lines[@]}" -eq 1 ]
```
aluzzardi: This line is not necessary since we are checking below `"${output}" == "0" ` (without any `*` so this is exact matching)

aluzzardi:
Great, only one very very minor comment which you could ignore.

I'm going to try locally as well.

PR #665, opened 2015-04-23
fix consul help

vieux:


vieux:
@abronan follow up #583

abronan:
LGTM

PR #666, opened 2015-04-23
Add note against using devicemapper as a storage driver to run the integration tests

abronan:


vieux:
LGTM (did you try with overlay?)

chanwit:
+1 :):)

abronan:
@vieux I need to upgrade my kernel to 3.18+ on Swarm agents, but this should work fine with overlay.

PR #675, opened 2015-04-26
Add Integration test for dependency filters

snrism:
Integration test for 3 dependency filters

Shared Volumes
Links
Shared Network Stack

### DIFF
--- a/test/integration/dependency.bats
+++ b/test/integration/dependency.bats
@@ -1,0 +1,83 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+}
+
+@test "shared volumes dependency" {
+	start_docker 2
+	swarm_manage
+
+	# Running the second container with shared volumes.
+	run docker_swarm run --name b1 -e constraint:node==node-0 -d busybox:latest sleep 500
+	[ "$status" -eq 0 ]
+	run docker_swarm run --name b2 --volumes-from=/b1 -d busybox:latest sh
+	[ "$status" -eq 0 ]
+
+	# check if containers share volume.
+	run docker_swarm inspect -f "{{ .HostConfig.VolumesFrom }}" b2
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"[/b1]"* ]]
+
+	# check if both containers are started on the same node
+	run docker_swarm inspect -f "{{ .Node.Name }}" b1
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"node-0"* ]]
+
+	run docker_swarm inspect -f "{{ .Node.Name }}" b2
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"node-0"* ]]
+}
+
+@test "links dependency" {
+	start_docker 2
+	swarm_manage
+
+	# Running the second container with link dependency.
+	run docker_swarm run --name b1 -e constraint:node==node-1 -d busybox:latest sleep 500
+	[ "$status" -eq 0 ]
+	run docker_swarm run --name b2 --link=/b1:foo -d busybox:latest sh
+	[ "$status" -eq 0 ]
+
+	# check if containers share link.
+	run docker_swarm inspect -f "{{ .HostConfig.Links }}" b2
+	[ "$status" -eq 0 ]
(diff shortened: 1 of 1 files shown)


vieux:
@snrism could we also check that both containers end up on the same node ?

snrism:
@vieux Sure..let me update that.

snrism:
@vieux Can you check if it looks OK now? Thanks

vieux:
@snrism thanks ping me when you are done

vieux:
@snrism I see you added some contraint but you never check that the node are on the right one.
you can take a look at https://github.com/vieux/swarm/blob/affinity_labels/test/integration/affinities.bats for an example.

snrism:
@vieux Updated, you can check now. I added "-e constraint:node==node-0" for 3 tests and the test cases passes.

snrism:
@vieux Oops.. Sorry, will add that.

snrism:
@viuex: When I rebased from master and ran the integration test now, I see that the "swarm_manage" step fails. Is this related to https://github.com/docker/swarm/commit/c7746f9463e7c709174eb8998335332bb9b263ce ?

Do i have to set something or run any additional commands? Thanks.

snrism:
@viuex: updated the test to check for dependencies and node. Let me know if this looks OK. Thanks.

test/integration/dependency.bats
```diff
@@ -0,0 +1,71 @@
...
+
+        # Running the second container with shared volumes.
+        run docker_swarm run --name b1 -e constraint:node==node-0 -d busybox:latest sleep 500
+        [ "$status" -eq 0 ]
+        run docker_swarm run --name b2 --volumes-from=/b1 -d busybox:latest sh
+        [ "$status" -eq 0 ]
+
+        # check if containers share volume.
+        run docker_swarm inspect -f "{{ .HostConfig.VolumesFrom }}" b2
+	[ "$status" -eq 0 ]
```
aluzzardi: Indentation looks wrong

test/integration/dependency.bats
```diff
@@ -0,0 +1,71 @@
...
+        [[ "${output}" == *"[/b1]"* ]]
+
+	# check if container "b2" is started on the same node
```
aluzzardi: I know this is a bit redundant since we are forcing `node==node-0` for b1, but could you actually check if `b1` is running on `node-0` as well? It will also make the test easier to understand.

test/integration/dependency.bats
```diff
@@ -0,0 +1,71 @@
...
+        [ "$status" -eq 0 ]
+        run docker_swarm run --name b2 --link=/b1:foo -d busybox:latest sh
+        [ "$status" -eq 0 ]
+
+        # check if containers share link.
+        run docker_swarm inspect -f "{{ .HostConfig.Links }}" b2
+        [ "$status" -eq 0 ]
+        [[ "${output}" == *"[/b1:/b2/foo]"* ]]
+
+	# check if container "b2" is started on the same node
```
aluzzardi: Indentation

test/integration/dependency.bats
```diff
@@ -0,0 +1,71 @@
...
+        run docker_swarm inspect -f "{{ .Node.Name }}" b2
```
aluzzardi: Same comment regarding checking if b1 is running on node-1

test/integration/dependency.bats
```diff
@@ -0,0 +1,71 @@
...
+        [ "$status" -eq 0 ]
+        run docker_swarm run --name b2 --net=container:/b1 -d busybox:latest sh
+        [ "$status" -eq 0 ]
+
+        # check if containers have shared network stack.
+        run docker_swarm inspect -f "{{ .HostConfig.NetworkMode }}" b2
+        [ "$status" -eq 0 ]
+        [[ "${output}" == *"container:/b1"* ]]
+
+	# check if container "b2" is started on the same node
```
aluzzardi: Same here.

aluzzardi:
Thanks @snrism, looks great! A few nits.

Also, would you mind checking the indentation of the file? It should be all tabs.

snrism:
@aluzzardi Thanks for the comments. I have checked if the first container is hosted on the same node and checked for indentation. Does this look ok?

aluzzardi:
Perfect, thanks!

PR #683, opened 2015-04-28
Fast exit from create on invalid command-line

icecrime:
The `create` command now fails immediately when invoked with an invalid command-line rather than going as far as initializing discovery and creating a cluster.


abronan:
@icecrime Good call!

LGTM

vieux:
LGTM

PR #684, opened 2015-04-28
support pulling private images from docker pull

vieux:

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"bytes"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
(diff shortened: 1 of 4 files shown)


vieux:
ping @aluzzardi

api/handlers.go
```diff
@@ -251,6 +252,11 @@ func postImagesCreate(c *context, w http.ResponseWriter, r *http.Request) {
 	wf := NewWriteFlusher(w)
 
 	if image := r.Form.Get("fromImage"); image != "" { //pull
+		authConfig := dockerclient.AuthConfig{}
+		buf, err := base64.URLEncoding.DecodeString(r.Header.Get("X-Registry-Auth"))
```
aluzzardi: This looks good, however I was wondering why passing the auth only when using `fromImage`?

If we receive a `X-Registry-Auth` from the client, shouldn't we always pass it?

vieux: when `fromImage` is not provided it means import, and we don't need `authConfig` for import

aluzzardi:
LGTM

PR #685, opened 2015-04-28
Fix integration test cleanup.

aluzzardi:
- Do not run swarm using a function since it forks a subshell and we end
  up killing it instead of the binary.
- Build a binary at each run instead of using go run. This is faster and
  fixes the cleanup for the same reason as the subshell.


abronan:
Tested and indeed fixes the issue.

LGTM

vieux:
LGTM

PR #686, opened 2015-04-29
strip quotes to allow chaining with docker-machine

chanwit:
We have some scenario that uses `swarm manage $(machine config ...)` and machine returns quotes around TLS-related files. Docker cli consumes these strings normally, but Swarm does not. This is a trivial fix to remove quotes from the filenames to allow that. 

The problem occurs when chaining `machine config` to `swarm`, both on Windows/Cygwin and Linux.

### DIFF
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -7,6 +7,7 @@ import (
 	"io/ioutil"
 	"path"
 	"strconv"
+	"strings"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
@@ -78,9 +79,9 @@ func manage(c *cli.Context) {
 			log.Fatal("--tlscacert must be provided when using --tlsverify")
 		}
 		tlsConfig, err = loadTLSConfig(
-			c.String("tlscacert"),
-			c.String("tlscert"),
-			c.String("tlskey"),
+			strings.Trim(c.String("tlscacert"), `"`),
+			strings.Trim(c.String("tlscert"), `"`),
+			strings.Trim(c.String("tlskey"), `"`),
 			c.Bool("tlsverify"))
 		if err != nil {
 			log.Fatal(err)


ehazlett:
@chanwit i'm not sure we want to strip them as this could cause potential issues with paths i believe.  could you try to use `eval $(swarm manage $(machine config ...))` -- that should work properly.  for machine you must use eval for env example.

chanwit:
@ehazlett Thank you! So this would be a wrong place to fix then.

I am wondering how `docker` cli deal with this as I call `docker $(machine config ...)` and everything just works fine.
Maybe there is a proper filepath expansion in `docker` ?

ehazlett:
@chanwit i'm not sure on the docker side.  with machine, you must use `eval` to be safe.  It perhaps just worked but to be safe use eval.

Thanks for the feedback!

aluzzardi:
Let us know what we should do with this PR - not very familiar with the matter.

chanwit:
Here's my possible solution at the moment.
  - Leave the quotes as is
  - Look into the Docker cli codes to see how Docker deals with quotes around TLS files.
  - Rewrite the PR

vieux:
`"` are interpreted by the shell, I added some debug and 

`./swarm m --tlscert "dsad" --tlscacert "twa" --tls --tlskey "12"` 

prints

`twa dsad 12` without any quote, so I don't see how stripping quotes coud help.

chanwit:
@vieux yes, " should be interpreted by the shell. But in the case of chaining with `machine`, quotes are passed directly to `swarm` and it's caused this problem.

Could you try your debug binary with any `machine config`?
`./swarm m $(machine config a_node)`
I am sure you're going to see some quotes there.

aluzzardi:
You should be using `eval`, as @ehazlett mentioned.

Assuming `echo` is `swarm` and `python` is `machine`:
```
$ echo $(python -c "print 'swarm \"path\"'")
swarm "path"
$ eval echo $(python -c "print 'swarm \"path\"'")
swarm path
```

chanwit:
OK, I shall close this PR then.

PR #691, opened 2015-05-01
Embed ContainerConfig into Container

aluzzardi:
The cluster configuration is now embedded into the container. It's guaranteed to be correct (as in, generated by `BuildContainerConfig`) at all times.


aluzzardi:
Hello? Anyone here? :)

ping @vieux @abronan

abronan:
@aluzzardi I bugged when I saw the "embedded into **the** container" :smile:

LGTM, better this way.

vieux:
LGTM

PR #693, opened 2015-05-01
integration: Add the re-usable "retry" helper.

aluzzardi:
Retry was kind of extracted from wait_until_reachable. The goal is to be
able to wait for a command to succeed, or fail if it doesn't succeed
after a defined number of attempts.

Indirectly improve the error logging: the output of the retry command
won't show up unless it constantly failed.

### DIFF
--- a/test/integration/helpers.bash
+++ b/test/integration/helpers.bash
@@ -36,15 +36,29 @@ function swarm() {
 	"$SWARM_BINARY" "$@"
 }
 
+# Retry a command $1 times until it succeeds. Wait $2 seconds between retries.
+function retry() {
+	local attempts=$1
+	shift
+	local delay=$1
+	shift
+	local i
+
+	for ((i=0; i < attempts; i++)); do
+		run "$@"
+		if [[ "$status" -eq 0 ]] ; then
+			return 0
+		fi
+		sleep $delay
+	done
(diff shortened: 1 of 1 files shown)


aluzzardi:
/cc @jimmyxian 

Can you verify that the integration tests all work on your side with this PR?

vieux:
LGTM

abronan:
LGTM as well, waiting to hear back from @jimmyxian ;)

jimmyxian:
@aluzzardi @abronan @vieux 
Great! I will try to use it..

test/integration/helpers.bash
```diff
@@ -36,15 +36,29 @@ function swarm() {
...
+	local delay=$1
+	shift
+	local i
+
+	for ((i=0; i < attempts; i++)); do
+		run "$@"
+		if [[ "$status" -eq 0 ]] ; then
+			return 0
+		fi
+		sleep $delay
```
jimmyxian: for ((i=0; i < attempts; i++)); do
                sleep $delay
		run "$@"
		if [[ "$status" -eq 0 ]] ; then
			return 0
		fi
		
maybe we should change the 'sleep' position, put sleep function first. The test will be ok!

PR #695, opened 2015-05-01
DO NOT MERGE - REVIEW ONLY require username and password for swarm create

vieux:
Close #695

The server side part, isn't live yet please review the code.

Added a `-u,--username` and `-p.--password` to `swarm create`
Alternatively you can do a swarm create and it will ask for the username and password (without echo, like docker login)

You can combine with: 
```
$ swarm create -u vieux
Password:

53128764320641268482352786912
```

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -51,6 +51,11 @@
 			"Rev": "443437f5ea04da9d62bf3e05d7951f7d30e77d96"
 		},
 		{
+			"ImportPath": "github.com/docker/docker/pkg/term",
+			"Comment": "v1.4.1-216-g12fef2d",
+			"Rev": "12fef2d8df4e5da024418e324f7e3c3e82220a27"
+		},
+		{
 			"ImportPath": "github.com/docker/docker/pkg/units",
 			"Comment": "v1.4.1-3245-g443437f",
 			"Rev": "443437f5ea04da9d62bf3e05d7951f7d30e77d96"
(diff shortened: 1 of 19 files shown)


vieux:
ping @abronan @aluzzardi @jlhawn

vieux:
I know the tests are broken, I did it on purpose so we don't merge.

aluzzardi:
As discussed IRL, how about having `destroy` as well, requiring the same auth (both client and server side)

aluzzardi:
Also, this set aside, how about moving from discovery-staging to discovery?

vieux:
I agree on the destroy but let's change the url in another PR

aluzzardi:
How about this workflow:

```
$ swarm create
Warning: Anonymous cluster created. Anonymous clusters can be destroyed by anyone with knowledge of the token. Please provide a `--username` in order to create an authenticated cluster.
53128764320641268482352786912
```

```
$ swarm create -u vieux
Password:
53128764320641268482352786912
```

/cc @bfirsh @shykes

vieux:
ping @bfirsh @shykes

moxiegirl:
@vieux Do we need to update the user-facing documentation for this? I saw you hit the token README

moxiegirl:
@vieux We need to add to `docs/discover.md` , `install-manual.md`, `install-w-machine.md` that tells users that they will need to do this. I'm assuming that it is the `docker login` credentials.  That can happen after your merge if you are moving fast.

Yam408:
H

PR #696, opened 2015-05-02
integration tests: Support for running within Docker.

aluzzardi:
With this change, integration tests can run using Docker with zero dependencies:

```
$ git clone https://github.com/docker/swarm.git
$ swarm/test/integration/run.sh
```

The environment to run the tests (containing docker, go and bats) is pulled and the code+tests are injected through a volume.

The environment was built with `test/integration/Dockerfile`

By default `run.sh` will run all tests but you can specify which tests to run, for instance:
```
$ test/integration/run.sh constraints.bats affinities.bats
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ script:
   - script/validate-gofmt
   - go vet ./...
   - fgt golint ./...
-  # Lint our test files and make sure they are not space indented.
-  - fgt find test -type f -exec grep -Hn -e "^ " {} \;
+  # Lint shell files and make sure they are not space indented.
+  - fgt find test/ -type f \( -name "*.sh" -or -name "*.bash" -or -name "*.bats" \) -exec grep -Hn -e "^ " {} \;
   - go test -v -race ./...
   - script/coverage
   - goveralls -service=travis-ci -coverprofile=goverage.report
--- a/test/integration/Dockerfile
+++ b/test/integration/Dockerfile
@@ -1,0 +1,33 @@
+# Dockerfile for swarm integration test environment.
+# Use with run_in_docker.sh
+
+FROM dockerswarm/dind:1.6.0
+
+# Install dependencies.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+        build-essential \
+        file \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Install golang
+ENV GO_VERSION 1.4.2
+RUN curl -sSL https://golang.org/dl/go${GO_VERSION}.src.tar.gz | tar -v -C /usr/local -xz \
+    && mkdir -p /go/bin
+ENV PATH /go/bin:/usr/local/go/bin:$PATH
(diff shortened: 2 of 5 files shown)


aluzzardi:
Hey @jfrazelle, does this work for Jenkins?

/cc @vieux @abronan

aluzzardi:
Also, very important: Tests are failing right now. This is very work in progress but I'd need @vieux @abronan and @jimmyxian help to figure out why they are not working.

aluzzardi:
Replaces #651

aluzzardi:
Using `vfs` the tests don't slow down by much and the environment is more reproducible on every host.

```
STORAGE_DRIVER=aufs bats test/integration/constraints.bats  3.75s user 1.17s system 13% cpu 37.104 total
STORAGE_DRIVER=vfs bats test/integration/constraints.bats  4.05s user 1.39s system 13% cpu 40.089 total
```

aluzzardi:
Depends on #700

jimmyxian:
@aluzzardi OK, i will test all api integration test.

jfrazelle:
so this `swarm/test/integration/run.sh` bash command would need to be run in a container

what we are trying to avoid is if someone makes a pr changing that bash script to `rm -rf /` etc

so ideally i need something like:

```
docker build blah blah
docker run blah blah
```

its how all the other projects are set up

aluzzardi:
@jfrazelle I'll look into this.

One thing though: Since we need `--privileged` (like other projects do I suppose), I still can mount / and `rm -rf` the heck out of it, it will just take me 2 lines instead of 1 ;)

I'm also worried about performance - the tests already take half an hour to run.

jfrazelle:
well ya but it removes that in the container not on the host ;)

jfrazelle:
docker/docker tests take like 12-15 min so i mean up to you, it can't hurt

aluzzardi:
Well...

```
$ docker run -it --privileged debian:jessie bash
root@1b01bc4b5967:/# mount /dev/sda1 /mnt/
root@1b01bc4b5967:/# ls /mnt/
[this is my host and I can wipe it out]
```

aluzzardi:
@jfrazelle How does this look?

Humans can just use `run.sh`

Jenkins can directly use `test_runner.sh` from a container.

For performance, the image is not built but rather pulled and the code is injected through a volume:

```
git clone <...>
docker pull dockerswarm/swarm-test-env
docker run --rm -t --privileged -v <git root>:/go/src/github.com/docker/swarm ./test_runner.sh
echo $?
```

jfrazelle:
cool will try

aluzzardi:
cant push until it's merged :) (it's an automated build)

How about we merge this as is if you agree on the concept? I still have some work to do before we can put it on jenkins (some tests are SLOW and some are broken)

jfrazelle:
yeah totally :)

jfrazelle:
let me know when ready :)

aluzzardi:
Ready!

PR #700, opened 2015-05-02
integration tests: run with native/vfs by default.

aluzzardi:
Signed-off-by: Andrea Luzzardi <aluzzardi@gmail.com>

PR #703, opened 2015-05-04
integration: Use the brand new dind image.

aluzzardi:
We now have dockerswarm/dind built using https://github.com/aluzzardi/dind:
- Much more lightweight.
- Exists in every version since Docker 1.0.0.

Depends on #700 as it cannot run on aufs.

### DIFF
--- a/test/integration/helpers.bash
+++ b/test/integration/helpers.bash
@@ -7,8 +7,8 @@ SWARM_ROOT=${SWARM_ROOT:-${BATS_TEST_DIRNAME}/../..}
 SWARM_BINARY=`mktemp`
 
 # Docker image and version to use for integration tests.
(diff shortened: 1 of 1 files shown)


vieux:
`image affinity` seems to fail with this PR and on an unrelated note, `docker cp` is stuck on both master and this PR

vieux:
@aluzzardi I wonder what you changed since the last image was able to run on aufs /cc @jfrazelle

aluzzardi:
@vieux not really sure but:

1/ It wasn't supposed to work: https://github.com/docker/docker/blob/master/daemon/graphdriver/aufs/aufs.go#L48

2/ @jfrazelle, correct me if I'm wrong, but we run jenkins on debian therefore no aufs anwyay?

jfrazelle:
there is aufs, we test all graph drivers, prs are all aufs

PR #707, opened 2015-05-04
improve tag image

jimmyxian:
Improve tag image so that it can refresh image in cluster.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -410,3 +410,17 @@ func (c *Cluster) RenameContainer(container *cluster.Container, newName string) 
 	st.Name = newName
 	return c.store.Replace(container.Id, st)
 }
+
+// TagImage tag an image
+func (c *Cluster) TagImage(IDOrName string, repo string, tag string, force bool) error {
+	c.RLock()
+	defer c.RUnlock()
+
+	// check image
+	image := c.Image(IDOrName)
+	if image == nil {
+		return fmt.Errorf("could not find image: no such id: %s", IDOrName)
+	}
+
+	return image.Engine.TagImage(IDOrName, repo, tag, force)
+}
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -419,7 +419,7 @@ func (c *Cluster) TagImage(IDOrName string, repo string, tag string, force bool)
 	// check image
 	image := c.Image(IDOrName)
 	if image == nil {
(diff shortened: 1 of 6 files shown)


jimmyxian:

vieux:
I don't see the improvement here.

jimmyxian:
@vieux 
I think `Engine.RefreshImages()` should not be a public function. It should not be called in docker api. If all the three API (`docker tag\docker commit\docker build`)  do as the following way, it will be  become a private function.
```
docker api ---> cluster api ---> engine api  ---> dockerclient api
```
I don't known if it's meaningful, waiting for your reply. Tks
/cc @aluzzardi @abronan

vieux:
@jimmyxian I agree, it's an improvment regarding the quality of the code, not the feature.

LGTM

cluster/swarm/cluster.go
```diff
@@ -410,3 +410,17 @@ func (c *Cluster) RenameContainer(container *cluster.Container, newName string)
...
+
+// TagImage tag an image
+func (c *Cluster) TagImage(IDOrName string, repo string, tag string, force bool) error {
+	c.RLock()
+	defer c.RUnlock()
+
+	// check image
+	image := c.Image(IDOrName)
+	if image == nil {
+		return fmt.Errorf("could not find image: no such id: %s", IDOrName)
```
vieux: can you please use `fmt.Errorf("No such image %s", IDOrName)`

jimmyxian: OK

jimmyxian:
ping @vieux 
Modified by the comment. check again, thks

aluzzardi:
I think I'm the one who suggested this, but now I'm having second thoughts.

I'm worried that the more we don't rely on proxy and use our own implementation, the more things will break since we need to re-implement them every time.

I realized this with `ps`: We don't proxy `ps` (we can't), and realized `--filter` doesn't work and since docker 1.6 they added `--filter label=k=v` which we needed to add.

I don't know if this is the path we want to take.

@vieux ?

vieux:
@aluzzardi I have the same concerns.

aluzzardi:
Perhaps we should address these in 0.4 with #748

jimmyxian:
@aluzzardi Agree, using proxy can reduce a lot of work in our swarm. If we implement in swarm, Docker API changes, we also have to change it in swarm.
@aluzzardi @vieux  Shall I close this one?

aluzzardi:
I think so yeah

PR #710, opened 2015-05-04
remove unused fields from node

vieux:


vieux:
We only need `Used*` and `Total*`

abronan:
LGTM

PR #712, opened 2015-05-05
integration: Use aufs by default.

aluzzardi:
The problem has been fixed in the dind images.


abronan:
LGTM

PR #716, opened 2015-05-05
integration: Run tests with Docker master

aluzzardi:
- Support for using a different CLI (set by `DOCKER_BINARY`)
- `run.sh` will now extra the CLI from the docker image and use it to run tests
- Updated `DOCKER_IMAGE` to `dockerswarm/dind-master`, which is automatically built on every merge of `docker/docker`
- Added a test to verify that the client and server versions are the same

### DIFF
--- a/test/integration/api.bats
+++ b/test/integration/api.bats
@@ -321,7 +321,8 @@ function teardown() {
 	swarm_manage
 	run docker_swarm info
 	[ "$status" -eq 0 ]
-	[[ "${output}" == *"Nodes: 1 "* ]]
+	echo $output
+	[[ "${output}" == *"Nodes: 1"* ]]
 	[[ "${output}" == *"└ Labels:"*"foo=bar"* ]]
 }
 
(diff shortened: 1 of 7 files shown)


aluzzardi:
To all writing integration tests: @vieux @abronan @jimmyxian @chanwit 

Please remember: Never use `docker` unless you are combining it with `-H`. If you want to talk to the host docker engine (like in `zk-discovery.bats`), now you have to use `docker_host`.

chanwit:
@aluzzardi ack

aluzzardi:
@vieux please take a quick look at the changes.

Some are okay (the ordering changed and I've future-proofed by checking `$output` instead of `$lines`) but `inspect --format={.Node.IP}` is broken because of #717

jimmyxian:
@aluzzardi OK

aluzzardi:
@vieux @abronan can you review this?

abronan:
@aluzzardi overall LGTM

Just, as this clearly aims to run this on Jenkins, shouldn't we package @jimmyxian's **wssh** image under `dockerswarm` instead? Just to keep control over this instead of relying on another repository. For now it's ok but just asking.

aluzzardi:
@vieux I need to merge this. Please have a look after.

@abronan Because one can still run the tests without jenkins

Issue #692, opened 2015-05-01
Need to check if network address in "swarm join" is a valid network interface


snrism:
Here is my setup:

host1: 192.168.111.78, host2: 192.168.111.79, host3: 192.168.111.80

If by mistake I join host2 with host1's IP into my swarm cluster with `swarm join --addr=192.168.111.78:2375 token://4432db3c4617b18655632bf091d9dfc7` (I used host1's address in host2), there is no error in host2 to indicate that I have started my `swarm join` with a wrong IP address nor there is any log in the master node to indicate this error.

When `docker` or `swarm manage` is started with such misconfig, the agent catches this error. However `swarm join` doesnt catch this error. 

I worked on this issue (https://github.com/snrism/swarm/commit/661e8f5855f8aea5d7963cb33b228baa8369b981) to check if the IP (currently only IPv4) is a valid network interface in the host. I am not sure if my implementation is the right way to check this, but appreciate any feedback if this works?

arbedout:
Cool idea. This may be a problem if the interface that swarm manager can reach is not in the list of interfaces returned by `net.InterfaceAddrs()` though. (Certain NAT and HTTP proxy setups come to mind). 

Could this be just a `log.Printf()` instead of a `log.Fatalf()`? That way the user is warned but os.Exit() won't be called, and they're free to continue if they know what they're doing.

snrism:
@arbedout thanks

@aluzzardi any comments/feedback?

aluzzardi:
@snrism so, I'm thinking of `swarm join` running as a container inside docker. The address provided to `--addr` won't be visible by the container (probably).

Thoughts?

snrism:
@aluzzardi good point. Starting swarm inside a container will not bring this issue, but running the swarm binary in the host certainly generated this problem when I bootstrapped my application. I guess I can adjust my setup to spinup swarm inside a container, but do you think it is worth documenting this issue somewhere so that folks start swarm in container as default? maybe some guidelines? I will takedown the PR related to this.

vieux:
@snrism I think @aluzzardi the good practice is the start the join in a container, that's the way docker-machine works today. As we don't want to force `--net host` I think it's better not to merge this PR/feature.

snrism:
@vieux @aluzzardi: Agreed. Closed the PR. Thanks

PR #718, opened 2015-05-05
Check if network address in "swarm join" is a valid network interface

snrism:
This pull request addresses https://github.com/docker/swarm/issues/692

I am re-stating the problem here:

Here is my setup:

host1: 192.168.111.78, host2: 192.168.111.79, host3: 192.168.111.80

If by mistake I join host2 with host1's IP into my swarm cluster with swarm join --addr=192.168.111.78:2375 token://4432db3c4617b18655632bf091d9dfc7 (I used host1's address in host2), there is no error in host2 to indicate that I have started my swarm join with a wrong IP address nor there is any log in the master node to indicate this error.

When docker or swarm manage is started with such misconfig, the agent catches this error. However swarm join doesnt catch this error. This PR checks if the IP (currently only IPv4, since in IPv6 it doesn't check for IPs like ::1/128) is a valid network interface in the host. Kindly let me know if this works.

### DIFF
--- a/cli/join.go
+++ b/cli/join.go
@@ -1,8 +1,10 @@
 package cli
 
 import (
+	"net"
 	"regexp"
 	"strconv"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


vieux:
I like the idea, @abronan @aluzzardi ?

aluzzardi:
@vieux I added some concerns on #692.

Could you take a look over there?

cli/join.go
```diff
@@ -15,6 +17,26 @@ func checkAddrFormat(addr string) bool {
...
+	newip := strings.Split(addr, ":")
+	// Handle loopback interface as a special case
+	if strings.EqualFold(newip[0], "localhost") {
+		return true
+	}
+	// Compare host's network interface addresses with requested address
+	addrs, err := net.InterfaceAddrs()
+	if err != nil {
+		log.Fatal(err)
+	}
```
chanwit: if err, should we `return false` just after logging fatal?
I think `addrs` is not valid after `err` is not `nil`.

snrism:
Closing this PR. Please  refer to https://github.com/docker/swarm/issues/692 for conclusion

PR #725, opened 2015-05-06
Discovery integration cleanup

aluzzardi:


test/integration/helpers.bash
```diff
@@ -86,32 +91,55 @@ function swarm_manage() {
...
+	echo "current: $current | nodes: $nodes" > log
+	for ((i=current; i < nodes; i++)); do
+		local h="${HOSTS[$i]}"
+		echo "Swarm join #${i}: $h $addr"
+		"$SWARM_BINARY" join --addr="$h" "$addr" &
 		SWARM_JOIN_PID[$i]=$!
-		((++i))
 	done
-	retry 30 1 [ -n "$(docker_swarm info | grep -q 'Nodes: $i')" ]
+	retry 10 0.5 check_discovery_nodes "$addr"
```
chanwit: This won't fail on the CI but will fail on my machines in case of `token` as mine are quite far and network lagging will be occurred. Is there anyway to workaround this number? Perhaps make it configurable like `$SWARM_ROOT` ?

chanwit:
@aluzzardi overall LGTM with few comments

aluzzardi:
@chanwit Can you try with this PR?

I think I fixed the issue - it wasn't the network lag, but rather the default  `--heartbeat` of 25 seconds, which meant that the manager only pulled discovery every 25 seconds to find new nodes with token discovery.

I've changed the heartbeat to 1 second therefore it shouldn't lag anymore.

chanwit:
@aluzzardi ah I see. The heartbeat rate fooled me then. I'll try it and let you know.

chanwit:
@aluzzardi sorry to not make it in time :(

aluzzardi:
No worries! Try with `master` and if it doesn't work we'll fix it

PR #727, opened 2015-05-06
integration: API tests split

aluzzardi:
Note: `test/integration/run.sh` as well as jenkins are going to run all tests.

However, `bats test/integration` is **not** as it doesn't do recursive directories.

You will have to `bats test/integration test/integration/api` if you want to test everything.

I don't think it's a problem, the direction we are going I guess is to use `bats` to run an individual test and rely on `run.sh` to run everything (since it already takes close to 10 minutes for the entire suite, it's not a big deal if it takes 30 extra seconds)

### DIFF
--- a/test/integration/affinities.bats
+++ b/test/integration/affinities.bats
@@ -47,7 +47,7 @@ function teardown() {
 	start_docker 2
 	swarm_manage
 
-	run docker -H ${HOSTS[0]}  build -t test $BATS_TEST_DIRNAME/testdata/build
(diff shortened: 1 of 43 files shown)


aluzzardi:
@vieux you are going to like this.

abronan:
@aluzzardi I like this too :ghost: no more time wasted on `api.bats` just to see if a single test works. :dancer: :dancers:

aluzzardi:
I have to merge this because of a push to master that went rogue.

PR #730, opened 2015-05-06
Integration performance

aluzzardi:
- Added `start_docker_with_busybox` which starts an engine with a preloaded busybox without pulling (just like `docker/docker` integration-cli).
- Use `start_docker_with_busybox` everywhere it's possible.
- Start only what we need.
- Improved correctness of a few tests at the same time.

### DIFF
--- a/test/integration/affinities.bats
+++ b/test/integration/affinities.bats
@@ -8,10 +8,10 @@ function teardown() {
 }
 
 @test "container affinty" {
-	start_docker 2
+	start_docker_with_busybox 2
 	swarm_manage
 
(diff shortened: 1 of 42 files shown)


aluzzardi:
**Baseline**:
```
real	8m57.970s
user	0m31.422s
sys	1m7.109s
```

**This PR**:
```
real	5m28.964s
user	0m38.587s
sys	1m22.076s
```

vieux:
does it work without using `run.sh` ? I see a bunch of errors on my machine

vieux:
commit, events and images

PR #732, opened 2015-05-06
fix name checks in create and rename

### LINKED ISSUE #729: rename: Container ID and Name collision
I experienced this in our tests for `rename`, but it might be affecting other:

> Conflict, The name c4 is already assigned to c4faf30b321817bad939eef1c839f61cd5e111f2ff833de80880af754bdadd50. You have to delete (or rename) that container to be able to assign c4 to a container again.

Here we looked up `c4` as a name and collided with a container whose id starts with c4.

vieux:
fix #729

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -275,14 +275,13 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if container := c.cluster.Container(name); container != nil {
-		httpError(w, fmt.Sprintf("Conflict, The name %s is already assigned to %s. You have to delete (or rename) that container to be able to assign %s to a container again.", name, container.Id, name), http.StatusConflict)
-		return
-	}
-
 	container, err := c.cluster.CreateContainer(cluster.BuildContainerConfig(config), name)
 	if err != nil {
-		httpError(w, err.Error(), http.StatusInternalServerError)
+		if strings.HasPrefix(err.Error(), "Conflict") {
+			httpError(w, err.Error(), http.StatusConflict)
+		} else {
+			httpError(w, err.Error(), http.StatusInternalServerError)
+		}
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


vieux:
ping @abronan @aluzzardi I don't see how to add a test

chanwit:
It seems you need to create the first container, grep its `id`, then use the first part of the `id` as the `name` of the second container.

maybe I can help you write the integration test for this, if you'd like?

aluzzardi:
Yep, what @chanwit said.

```
id=$(docker_swarm create busybox)
prefix=$(printf "$id" | cut -c 1-4)
docker_swarm create --name "$prefix" busybox
docker_swarm create --name test busybox
docker_swarm rename test "$prefix"
```

Now that I think of it, even the full ID should work right?
```
id=$(docker_swarm create busybox)
docker_swarm create --name "$id" busybox
docker_swarm create --name test busybox
docker_swarm rename test "$id"
```

vieux:
@aluzzardi I added a test, thanks @chanwit

test/integration/name_conflict.bats
```diff
@@ -0,0 +1,31 @@
...
+
+load helpers
+
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+}
+
+@test "name conflict short ID" {
+	start_docker 1
```
aluzzardi: `start_docker_with_busybox` as you don't want to pull `busybox` when you do the `create`

test/integration/name_conflict.bats
```diff
@@ -0,0 +1,31 @@
...
+	id=$(docker_swarm create busybox)
+	prefix=$(printf "$id" | cut -c 1-4)
+	docker_swarm create --name "$prefix" busybox
+	docker_swarm create --name test busybox
+	run docker_swarm rename test "$prefix"
+	[ "$status" -eq 1 ]
+}
+
+@test "name conflict full ID" {
+	start_docker 1
```
aluzzardi: same

test/integration/name_conflict.bats
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/env bats
```
aluzzardi: What do you think of putting those in api/create.bats and api/rename.bats?

Right now they don't test much

aluzzardi:
LGTM

Issue #644, opened 2015-04-19
Integration tests for discovery


aluzzardi:
The following services need integration tests:

* [x] consul 
* [x] etcd 
* [x] file 
* [x] token 
* [x] zookeeper

aluzzardi:
Hey @chanwit, if by any chance you get bored, do you think you could tackle `consul`?

Similar to `zookeeper`, there is an image ready for consul: https://github.com/progrium/docker-consul

No pressure, we'll do it eventually :)

chanwit:
@aluzzardi no problem. I'll do it, but it would take sometime anyway.

snrism:
@aluzzardi @chanwit I was working on it. Let me try to finish it in next few days and open a PR. If I am unable to get it done, will drop in a note soon.

chanwit:
@snrism great! let me know if you've got the PR up.

snrism:
@chanwit @aluzzardi opened a PR for consul integration https://github.com/docker/swarm/pull/733. If this works, I can work on etcd next in the next few days as well.

PR #733, opened 2015-05-07
Integration test for consul discovery service

snrism:
Add integration test for consul. Updates to https://github.com/docker/swarm/issues/644

### DIFF
--- a/test/integration/discovery-consul.bats
+++ b/test/integration/discovery-consul.bats
@@ -1,0 +1,52 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# Address on which Consul will listen (random port between 8000 and 9000).
+CONSUL_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 8000 ))
+
+# Container name for integration test
+CONTAINER_NAME=swarm_consul
+
+function check_leader() {
+	# Confirm Cluster leader election
+	docker_host logs $CONTAINER_NAME | grep "New leader elected"
+	# Check member state
+	docker_host logs $CONTAINER_NAME | grep "consul: member '$CONTAINER_NAME' joined, marking health alive"
+}
+
+function start_consul() {
+	docker_host run --name=$CONTAINER_NAME -h $CONTAINER_NAME -p $CONSUL_HOST:8500 -d progrium/consul -server -bootstrap-expect 1 -data-dir /test
+	# Check if consul cluster leader is elected
+	retry 30 1 check_leader
+}
+
+function stop_consul() {
+	docker_host rm -f -v $CONTAINER_NAME
+}
+
+function setup() {
+	start_consul
+}
+
+function teardown() {
+	swarm_manage_cleanup
+	swarm_join_cleanup
+	stop_docker
+	stop_consul
+}
+
+@test "consul discovery" {
+	# Start 2 engines and make them join the cluster.
+	start_docker 2
+	swarm_join "consul://${CONSUL_HOST}/test"
+
+	# Start a manager and ensure it sees all the engines.
+	swarm_manage "consul://${CONSUL_HOST}/test"
+	check_swarm_nodes
+
+	# Add another engine to the cluster and make sure it's picked up by swarm.
+	start_docker 1
+	swarm_join "consul://${CONSUL_HOST}/test"
+	retry 30 1 check_swarm_nodes
+}


aluzzardi:
Awesome! It failed once but now it's working. I'm going to try to run it 3/4 times to see if it's flaky.

Thanks a lot

snrism:
Strange, it passed when I first opened the PR. Any changes reqd?

test/integration/discovery-consul.bats
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# Address on which Consul will listen (random port between 7000 and 8000).
```
aluzzardi: You should pick a different range since you might conflict with `zookeeper` which is also using that port range.

aluzzardi:
**Might** be a port conflict between consul and zk.

aluzzardi:
https://jenkins.dockerproject.com/job/Swarm-PRs/65/console

aluzzardi:
```
✗ consul discovery
   (from function `retry' in file ./helpers.bash, line 80,
    from function `swarm_join' in file ./helpers.bash, line 130,
    in test file ./discovery-consul.bats, line 35)
     `swarm_join "consul://${CONSUL_HOST}/test"' failed
   Swarm join #0: 127.0.0.1:5548 consul://127.0.0.1:7983/test
   time="2015-05-07T04:34:06Z" level=fatal msg="Unexpected response code: 500 (No cluster leader)" 
   Swarm join #1: 127.0.0.1:5549 consul://127.0.0.1:7983/test
   time="2015-05-07T04:34:06Z" level=fatal msg="Unexpected response code: 500 (No cluster leader)" 
   Command "check_discovery_nodes consul://127.0.0.1:7983/test" failed 10 times. Output: 0 out of 2 hosts joined discovery
```

aluzzardi:
Googling the issue seem like it's a common one. Maybe we're not giving consul enough time to bootstrap or something.

Could you have a look?

snrism:
OK.. lemme check again and update

aluzzardi:
@snrism By the way, I see some unrelated tests failing. Would you mind doing a rebase on master? We fixed a bunch of problems over the last couple of days and now the suite is mostly reliable.

snrism:
@aluzzardi I did rebase with the current master and looks like I am still facing similar issue. 

Also, I now use new port range for consul discovery so that it doesn't conflict with zookeeper and have introduced sleep for 20sec before swarm join is called so that the cluster leader is elected. checking if there is a better way to do this. 

@chanwit any comments?

chanwit:
@snrism is there any way to query the readiness of Consul? If so, you can just use `retry` the check until it's ready before proceeding to the next step.

test/integration/discovery-consul.bats
```diff
@@ -0,0 +1,47 @@
...
+# Address on which Consul will listen (random port between 8000 and 9000).
+CONSUL_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 8000 ))
+
+# Container name for integration test
+CONSUL_CONTAINER_NAME=swarm_consul
+
+function start_consul() {
+	run docker_host run -h $CONSUL_CONTAINER_NAME -p $CONSUL_HOST:8500 -d progrium/consul -server -bootstrap-expect 1 -data-dir /test
+	[ "$status" -eq 0 ]
+	sleep 20
```
chanwit: Sleep is not good for waiting. May be you should check with `retry`.

Ahh right, doing `docker_host logs` on the consul container and `grep`ing something out of it might work.

snrism: yes, I wasn't happy about sleep..yes, figuring that out...thanks for the inputs.

test/integration/discovery-consul.bats
```diff
@@ -0,0 +1,53 @@
...
+
+# Container name for integration test
+CONTAINER_NAME=swarm_consul
+
+function check_leader() {
+	docker_host logs $CONTAINER_NAME | grep "New leader elected"
+}
+
+function start_consul() {
+	run docker_host run -h $CONTAINER_NAME -p $CONSUL_HOST:8500 -d progrium/consul -server -bootstrap-expect 1 -data-dir /test
```
aluzzardi: What about using `-bootstrap` instead of ` -bootstrap-expect 1`?

snrism: The documentation specifes to use of -bootstrap-expect. -bootstrap was used in prior versions and for manual setup. https://www.consul.io/docs/guides/bootstrapping.html.

snrism:
sorry folks, I missed the name is the previous build. my bad. hope this works.

test/integration/discovery-consul.bats
abronan: +1 for `-bootstrap-expect 1`, it's easier to make sense that this is a standalone consul container than with `-bootstrap` only

aluzzardi:
@snrism: No worries.

Feel free to prepend "DO NOT MERGE" to the title of this PR and push as many changes as you like to trigger the integration tests, then revert the title when you are done experimenting.

aluzzardi:
Some things you could try:

- Use `-bootstrap` instead of `-bootstrap-expect 1`
- Launch consul with `-h consul-node` then look for `consul: member 'consul-node' joined, marking health alive` in the logs

aluzzardi:
Also,

@chanwit: It looks like zookeeper is failing often when running this. I don't know if the problem comes from the `zk` tests or if it's somehow related to this PR. Could you try with/without this PR to find out where the problem is coming from? Thanks!

snrism:
@aluzzardi As noted earlier (I guess the conversation got hidden in the diff) the recommended way is to use -bootstrap-expect rather than -bootstrap. -bootstrap was used in prior versions and for manual setup. https://www.consul.io/docs/guides/bootstrapping.html.

I now check for "consul: member '$CONTAINER_NAME' joined, marking health alive" in the logs.

Zookeeper failed in this build as well. Lemme check into it to see if I can spot any issue.

aluzzardi:
I'll retrigger the tests a few times and see if it's still flaky.

So far it worked twice in a row (yay!)

chanwit:
Maybe I need to improve the ZK test the same way. I'll do it tonight.

snrism:
@aluzzardi I noticed that the `events.bats` test fails in my system as well and mostly due to the timing between the following calls (for me create event was not recorded in the TEMP_FILE):

`docker_swarm events > $TEMP_FILE &`
`run docker_swarm create --name test_container -e constraint:node==node-0 busybox sleep 100 `

If I introduce some delay between the calls, the test passes in my system. Not sure if you faced similar problem.

aluzzardi:
@snrism Yes, we have lots of flaky tests which we are still fixing.

Integration tests have been integrated with Jenkins for just a few days so we get the initial problems :)

As long as the test in this PR is good, we'll merge it, disregarding the other unrelated failures (which we are solving in parallel).

It looks good to me now - the only small issue is with formatting (travis failure):
`test/integration/discovery-consul.bats:41:        retry 30 1 check_leader`

This is using space indentation rather than tabs

test/integration/discovery-consul.bats
```diff
@@ -0,0 +1,55 @@
...
+
+function check_leader() {
+	# Confirm Cluster leader election
+	docker_host logs $CONTAINER_NAME | grep "New leader elected"
+	# Check member state
+	docker_host logs $CONTAINER_NAME | grep "consul: member '$CONTAINER_NAME' joined, marking health alive"
+}
+
+function start_consul() {
+	run docker_host run --name=$CONTAINER_NAME -h $CONTAINER_NAME -p $CONSUL_HOST:8500 -d progrium/consul -server -bootstrap-expect 1 -data-dir /test
```
aluzzardi: Small nit: There is no need to use `run` here and checking $status -eq 0 if you are expecting this to work.

Without `run`, any command that exits with non-zero will fail the test

test/integration/discovery-consul.bats
```diff
@@ -0,0 +1,55 @@
...
+function teardown() {
+	swarm_manage_cleanup
+	swarm_join_cleanup
+	stop_docker
+	stop_consul
+}
+
+@test "consul discovery" {
+	# Check if consul cluster leader is elected
+        retry 30 1 check_leader
```
aluzzardi: How about moving this to `start_consul`?

test/integration/discovery-consul.bats
```diff
@@ -0,0 +1,55 @@
...
+	docker_host logs $CONTAINER_NAME | grep "consul: member '$CONTAINER_NAME' joined, marking health alive"
+}
+
+function start_consul() {
+	run docker_host run --name=$CONTAINER_NAME -h $CONTAINER_NAME -p $CONSUL_HOST:8500 -d progrium/consul -server -bootstrap-expect 1 -data-dir /test
+	[ "$status" -eq 0 ]
+}
+
+function stop_consul() {
+	run docker_host rm -f -v $CONTAINER_NAME
```
aluzzardi: Same comment regarding `run`

aluzzardi:
Few minor nits otherwise this is a LGTM

test/integration/discovery-consul.bats
snrism: OK, I initially had it as part of start_consul..will revert back.

aluzzardi:
Awesome. Please change the PR title if you're ready to merge, I'll do it as soon as the tests go through.

snrism:
Done

aluzzardi:
Thanks!

PR #738, opened 2015-05-07
docker ps: Support for filters and improve state management.

### LINKED ISSUE #734: support `ps --filter`
ps filters are server side in docker:
https://github.com/docker/docker/blob/master/daemon/list.go#L97

We need to support them as well.

aluzzardi:
- `docker ps` now fully supports `--filter` flags
- Generate `Status` in `docker ps` dynamically. "Up X seconds" is now
  real-time.
- Misc cleanup.

Fixes  #734

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -106,28 +106,104 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Parse flags.
 	all := r.Form.Get("all") == "1"
 	limit, _ := strconv.Atoi(r.Form.Get("limit"))
 
-	out := []*dockerclient.Container{}
+	// Parse filters.
+	filters, err := dockerfilters.FromParam(r.Form.Get("filters"))
+	if err != nil {
+		httpError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	filtExited := []int{}
+	if i, ok := filters["exited"]; ok {
+		for _, value := range i {
+			code, err := strconv.Atoi(value)
+			if err != nil {
+				httpError(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			filtExited = append(filtExited, code)
+		}
+	}
+	if i, ok := filters["status"]; ok {
+		for _, value := range i {
+			if value == "exited" {
+				all = true
+			}
+		}
+	}
+
+	// Filtering: select the containers we want to return.
+	candidates := []*cluster.Container{}
 	for _, container := range c.cluster.Containers() {
-		tmp := (*container).Container
 		// Skip stopped containers unless -a was specified.
-		if !strings.Contains(tmp.Status, "Up") && !all && limit <= 0 {
+		if !container.Info.State.Running && !all && limit <= 0 {
 			continue
 		}
+
 		// Skip swarm containers unless -a was specified.
-		if strings.Split(tmp.Image, ":")[0] == "swarm" && !all {
+		if strings.Split(container.Image, ":")[0] == "swarm" && !all {
 			continue
 		}
+
+		// Apply filters.
+		if !filters.Match("name", strings.TrimPrefix(container.Names[0], "/")) {
+			continue
+		}
+		if !filters.Match("id", container.Id) {
+			continue
+		}
+		if !filters.MatchKVList("label", container.Config.Labels) {
+			continue
+		}
+		if !filters.Match("status", container.StateString()) {
+			continue
+		}
+
+		if len(filtExited) > 0 {
+			shouldSkip := true
+			for _, code := range filtExited {
+				if code == container.Info.State.ExitCode && !container.Info.State.Running {
+					shouldSkip = false
+					break
+				}
+			}
+			if shouldSkip {
+				continue
+			}
+		}
+
+		candidates = append(candidates, container)
+	}
+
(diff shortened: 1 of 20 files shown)


api/handlers.go
```diff
@@ -106,28 +106,104 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
...
 	limit, _ := strconv.Atoi(r.Form.Get("limit"))
 
-	out := []*dockerclient.Container{}
+	// Parse filters.
+	filters, err := dockerfilters.FromParam(r.Form.Get("filters"))
+	if err != nil {
+		httpError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	filtExited := []int{}
```
chanwit: I think `lint` will complain this?
should it be `var filtExited []int`.

test/integration/api/ps.bats
```diff
@@ -44,3 +44,61 @@ function teardown() {
 	[ "${#lines[@]}" -eq  2 ]
 	[[ "${lines[1]}" == *"false"* ]]
 }
+
+@test "docker ps --filter" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	# Running
+	firstID=$(docker_swarm run -d --name name1 --label "match=me" --label "second=tag" busybox sleep 10000)
```
chanwit: Does this help?
```
run docker_swarm run -d --name name1 --label "match=me" --label "second=tag" busybox sleep 10000
firstID=$output
[ $status -eq 0 ]
```
Or we just don't care about exit codes of these three runnings?

cluster/container.go
```diff
@@ -10,3 +16,50 @@ type Container struct {
 	Info   dockerclient.ContainerInfo
 	Engine *Engine
 }
+
+// Status returns a human-readable description of the state
+// Stoken from docker/docker/daemon/state.go
+func (c *Container) Status() string {
+	s := c.Info.State
+	if s.Running {
+		if s.Paused {
```
chanwit: Should we reuse `startedAt`.
`startedAt := units.HumanDuration(time.Now().UTC().Sub(s.StartedAt))`
All next three statements just need to use it.

test/integration/api/ps.bats
aluzzardi: Yeah, actually I've been discovering a few things that may lead to best practices.

In bats, by default, every command that doesn't exit with status 0 will fail automatically. As an additional benefit, we get its output on the logs.

Unless we expect a command to fail or we want to get its `stderr`, it's better to **not** use `run` I believe.

So I believe that instead of:
```
run foo
[ $status -eq 0 ]
```

We should always do
```
foo
```

And instead of
```
run foo
[ $status -eq 0 ]
local thing=$output
```

```
thing=$(foo)
```

If we want to get `foo` stdout/stderr, do pattern matching on `$output`, or check `$status` for something different than `0`, then `run` is the way to go.

aluzzardi: /cc @vieux @abronan @jimmyxian

cluster/container.go
aluzzardi: I just took this from `docker/docker/daemon/state.go` without touching :)

The second statement uses `FinishedAt` though, so we could only re-use it twice, not sure if it's worth it.

cluster/container.go
```diff
@@ -10,3 +16,50 @@ type Container struct {
 	Info   dockerclient.ContainerInfo
 	Engine *Engine
 }
+
+// Status returns a human-readable description of the state
```
aluzzardi: @vieux 

Should those two functions be in `dockerclient` instead?

test/integration/api/ps.bats
vieux: ok !

api/handlers.go
vieux: no it's fine

cluster/container.go
vieux: I guess so yes

aluzzardi: It's complicated because of the dependency on docker/docker/pkg/units

vieux:
LGTM

vieux:
ping @abronan

cluster/container.go
```diff
@@ -10,3 +16,50 @@ type Container struct {
...
+	if s.FinishedAt.IsZero() {
+		return ""
+	}
+
+	return fmt.Sprintf("Exited (%d) %s ago", s.ExitCode, units.HumanDuration(time.Now().UTC().Sub(s.FinishedAt)))
+}
+
+// StateString returns a single string to describe state
+// Stoken from docker/docker/daemon/state.go
+func (c *Container) StateString() string {
```
abronan: @aluzzardi do we ignore `ghosted` containers (`s.Ghost`)? What can we do about those in Swarm? Will we keep them or hide them from Swarm supervision?

vieux: I think ghost isn't used since docker 0.8 or something like this.

abronan: @vieux True! Just checked the latest API version and it's not there. This is still in `dockerclient` though so we should probably delete it now ;-)

abronan:
LGTM

test/integration/api/ps.bats
jimmyxian: OK!

cluster/container.go
chanwit: I see.

PR #739, opened 2015-05-07
Add support save images

jimmyxian:
Add support save images

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -62,6 +62,59 @@ func getVersion(c *context, w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(version)
 }
 
+// GET /images/get
+func getImages(c *context, w http.ResponseWriter, r *http.Request) {
+	if err := r.ParseForm(); err != nil {
+		httpError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	names := r.Form["names"]
+
+	// find an engine which has all images
+	// Engine.Addr : [found names]
+	dict := make(map[string][]string)
+	bFoundEngine := false
+	var foundEngineAddr string
+	for _, image := range c.cluster.Images() {
+		for _, name := range names {
+			if image.Match(name) {
+				// check if engine addr already exists
+				value, exists := dict[image.Engine.Addr]
+				if exists {
+					// check if name already exists
+					found := false
+					for _, tempName := range value {
+						if tempName == name {
+							found = true
+						}
+					}
+					if found == false {
+						dict[image.Engine.Addr] = append(value, name)
+						if len(names) == len(dict[image.Engine.Addr]) {
+							bFoundEngine = true
+							foundEngineAddr = image.Engine.Addr
+						}
+					}
+				} else {
+					dict[image.Engine.Addr] = []string{name}
+				}
+			}
+			if bFoundEngine {
+				break
+			}
+		}
+		if bFoundEngine {
+			break
+		}
+	}
+
+	if bFoundEngine {
+		proxy(c.tlsConfig, foundEngineAddr, w, r)
+	} else {
+		httpError(w, fmt.Sprintf("Not found an engine which has all images: %s", names), http.StatusNotFound)
+	}
+}
+
 // GET /images/json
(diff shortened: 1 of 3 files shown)


jimmyxian:

test/integration/api/save.bats
```diff
@@ -42,3 +42,31 @@ function teardown() {
...
 	rm -f $temp_file_name_o
 }
+
+@test "docker save muti-images" {
+	start_docker_with_busybox 1
+	start_docker 1
+	swarm_manage
+	
+	# tag busybox
+	run docker_swarm tag busybox testimage
```
aluzzardi: So, in order to make sure that `save` actually looks up at different machines I would probably do something like:

```
start_docker_with_busybox 2
docker -H ${HOSTS[0]} tag busybox test1
docker -H ${HOSTS[0]} tag busybox test2
docker_swarm save test1 test2
```

This makes sure `save` is looking at both hosts.

aluzzardi:
@vieux You are way more familiar than me with what's that supposed to do, could you take a look?

What kind of things should be tested? Maybe a load afterwards on a third (clean host, without even `busybox`, just `start_docker`), run `images` on it and make sure both `test1` and `test2` are there?

test/integration/api/save.bats
jimmyxian: Well, does not support multi images on multiple nodes. :)

api/handlers.go
```diff
@@ -62,6 +62,59 @@ func getVersion(c *context, w http.ResponseWriter, r *http.Request) {
...
+	if err := r.ParseForm(); err != nil {
+		httpError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	names := r.Form["names"]
+
+	// find an engine which has all images
+	// Engine.Addr : [found names]
+	dict := make(map[string][]string)
+	bFoundEngine := false
```
chanwit: I think we do not need this `bFoundEngine`.
Just check with `foundEngineAddr = ""` as Golang alreay inits it to an empty string.

api/handlers.go
```diff
@@ -62,6 +62,59 @@ func getVersion(c *context, w http.ResponseWriter, r *http.Request) {
...
+	var foundEngineAddr string
+	for _, image := range c.cluster.Images() {
+		for _, name := range names {
+			if image.Match(name) {
+				// check if engine addr already exists
+				value, exists := dict[image.Engine.Addr]
+				if exists {
+					// check if name already exists
+					found := false
```
chanwit: Could you change this to something like `nameAlreadyExisted` instead of `found`?

api/handlers.go
```diff
@@ -62,6 +62,59 @@ func getVersion(c *context, w http.ResponseWriter, r *http.Request) {
...
+func getImages(c *context, w http.ResponseWriter, r *http.Request) {
+	if err := r.ParseForm(); err != nil {
+		httpError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+	names := r.Form["names"]
+
+	// find an engine which has all images
+	// Engine.Addr : [found names]
+	dict := make(map[string][]string)
```
chanwit: Variable named `dict` is too general. I guess it's the `imageMap` ?

aluzzardi:
Regarding my previous comment on saving images which are on different machines:

1/ If it's something we want to support, we should add a test
2/ If it's something we **don't** want to support, we should add a test anyway and verify that it fails properly

vieux:
@jimmyxian @aluzzardi I say we don't support save on multi machine (at least for now) so we it would be nice to add a proper error and a test.

vieux:
Also, can we then rmi busybox and testimage; load the tar file; and do a images again ?

api/handlers.go
jimmyxian: OK

api/handlers.go
jimmyxian: OK

jimmyxian:
@aluzzardi  @vieux Yeah. I will update by the comments.

jimmyxian:
ping @aluzzardi @vieux @chanwit Modified by the comments, check again, thks

Issue #600, opened 2015-04-08
Virtual Container IDs


aluzzardi:
Re-scheduling containers (see #599) means they will change ID. This is unacceptable since it's the only link between the user and a container.

Swarm should create a **Virtual Swarm ID** for every container that maps to an actual **Physical ID** from the engine.

Only the **Virtual ID** is exposed to the user, and Swarm maintains a mapping between the two. A rescheduled container will have a different **Physical ID** but the same **Virtual ID**.

Virtual IDs could be stored as labels, see #288.

smothiki:
@aluzzardi what is the Use case for containers getting re-scheduled other than node fail-over. Any ways Swarm is trying to enforce a unique naming scheme for containers across the cluster . May I know more about how Virtual ID is useful.

aluzzardi:
This is work in progress, so feedback is very much welcome.

Yes, the only re-scheduling reason (for the foreseeable future) is node fail-over.

Some reasons why we would need this:
* **Centralized Container IDs**: Right now, each engine creates its own IDs. They might conflict with each other on different engines. Creating them in a central place guarantees uniqueness.
* **Consistent IDs**: When moving a container (for instance, in case of node fail-over), its ID will change. Presenting a Virtual ID to the user would prevent that.
* **ID instantly available**: Right now, the scheduler (through `docker create`) must be synchronous since we need to return an ID through the API as soon as we return the call. Having a Virtual ID would allow Swarm to generate immediately a Virtual ID, put the container into the scheduler and return immediately to the user. Whenever the container gets actually scheduled, the Virtual ID will be mapped to an Engine ID.

In order to work easily with Virtual IDs, we would use labels to tag containers with their Virtual ID.
Running `docker ps` directly on an engine could display Virtual IDs as a label.
Running `docker ps` through Swarm could display the Engine ID as a label.

tnachen:
So does all the docker commands will try to find the ID in both docker names/ids and also the labels virtual IDs then? ie: docker stop virtualId1

vieux:
@tnachen yes, that's the plan. I also think that `docker run` should return the virtual ID, so you would use virtual ID all the time, and almost never actual container ID

tnachen:
cool, the user experience definitely becomes a bit interesting. I think we have to tell the user this is a virtual ID, since when docker ps is called users will get confused which one to look at.
[...]

PR #745, opened 2015-05-07
Support for Swarm IDs

aluzzardi:
Swarm ID is an additional ID we have per container.

It's useful for a few reasons:
- We can know whether a container was created by Swarm or not
- User can address their containers by such ID
- It will remain constant in case of rebalancing


aluzzardi:
Fixes #600

api/handlers.go
```diff
@@ -251,6 +256,32 @@ func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		httpError(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	// Compatibility: Inject the Swarm ID label if not available.
+	// This **must** be done before inserting the `Node` field since it too has
+	// a `Label` field.
+	if !bytes.Contains(data, []byte("\"com.docker.swarm.id\":")) {
```
chanwit: probably
```
if !bytes.Contains(data, []byte(`"com.docker.swarm.id":`)) {
```
would be easier to read?

api/handlers.go
```diff
@@ -251,6 +256,32 @@ func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
...
+		label := fmt.Sprintf("%q:%q", "com.docker.swarm.id", container.Config.SwarmID())
```
chanwit: Not sure if we need a constant for `"com.docker.swarm.id"`

cluster/config.go
```diff
@@ -87,6 +87,17 @@ func (c *ContainerConfig) extractExprs(key string) []string {
 	return exprs
 }
 
+// SwarmID extracts the Swarm ID from the Config.
+// May return an empty string if not set.
+func (c *ContainerConfig) SwarmID() string {
+	return c.Labels[namespace+".id"]
```
chanwit: Oh right, I found the `namespace` const here. Is it possible to be reused in `api/handlers` as well?

cluster/config_test.go
```diff
@@ -29,6 +29,18 @@ func TestBuildContainerConfig(t *testing.T) {
...
+	config := BuildContainerConfig(dockerclient.ContainerConfig{})
+	assert.Empty(t, config.SwarmID())
+	config.SetSwarmID("foo")
+	assert.Equal(t, config.SwarmID(), "foo")
+
+	// Retrieve an existing ID.
+	config = BuildContainerConfig(dockerclient.ContainerConfig{Labels: map[string]string{namespace + ".id": "test"}})
+	assert.Equal(t, config.SwarmID(), "test")
+}
+
```
chanwit: @aluzzardi  is it possible to add another test for the fake id? I saw you mentioned about the backward compatibility for 0.2.0.

cluster/config.go
vieux: yes I need it too for another PR, I think we could do a `consts.go` in `/cluster` and export the variable there.

aluzzardi:
Thanks for the feedback!

I decided to remove backward compatibility altogether.

It's not a huge deal, backward compatibility was meant to do stuff like rebalancing on containers created prior to 0.3, but since rebalancing will be an opt-in feature (there must be a flag passed during `docker run`), we won't rebalance older containers anyway so it's not necessary.

PTAL

aluzzardi:
PTAL @vieux @abronan

vieux:
LGTM

abronan:
LGTM

PR #751, opened 2015-05-08
DO NOT MERGE - docs: user guide overhaul

aluzzardi:


docs/index.md
```diff
@@ -49,16 +52,24 @@ containers on the same host. For details, see
 
 ## Pre-requisites for running Swarm
 
-You must install Docker 1.4.0 or later on all nodes. While each node's IP need not
-be public, the Swarm manager must be able to access each node across the network.
+Each machine of your cluster must be running Docker 1.4.0 or later.
```
thaJeztah: Each machine "in" your cluster?

docs/index.md
```diff
@@ -49,16 +52,24 @@ containers on the same host. For details, see
...
-To enable communication between the Swarm manager and the Swarm node agent on
-each node, each node must listen to the same network interface (TCP port).
-Follow the set up below to ensure you configure your nodes correctly for this
-behavior.
+By default, Docker listens on `unix:///var/run/docker.sock` to allow only local connections.
 
-> **Note**: Swarm is currently in beta, so things are likely to change. We
-> don't recommend you use it in production yet.
+To enable communication between Swarm and Docker running on different machines,
+each Docker must be
```
thaJeztah: each Docker *Engine*?

docs/index.md
```diff
@@ -49,16 +52,24 @@ containers on the same host. For details, see
...
+By default, Docker listens on `unix:///var/run/docker.sock` to allow only local connections.
 
-> **Note**: Swarm is currently in beta, so things are likely to change. We
-> don't recommend you use it in production yet.
+To enable communication between Swarm and Docker running on different machines,
+each Docker must be 
+[configured] (https://docs.docker.com/articles/basics/#bind-docker-to-another-hostport-or-a-unix-socket)
+to listen to a TCP port.
+
+While each machine's IP need not be public, the Swarm manager must be able to access
```
thaJeztah: IP address

docs/index.md
```diff
@@ -49,16 +52,24 @@ containers on the same host. For details, see
...
+each machine across the network.
```
thaJeztah: s/each/the/ (sentence already starts with each)

docs/index.md
```diff
@@ -49,16 +52,24 @@ containers on the same host. For details, see
...
+
+Follow the set up below to configure your machines correctly for this behavior.
+
+> Warning: Changing the default `docker` daemon binding to a TCP port will
```
thaJeztah: `**Warning**:`

docs/index.md
```diff
@@ -49,16 +52,24 @@ containers on the same host. For details, see
...
+> increase your security risks by allowing anyone with access to that port
+> full Docker access; so it is not advisable on an open network.
```
thaJeztah: Might want to add that having access to the API is equivalent to having root access to the host. I think there's a note on that somewhere in the docker docs.

Perhaps add a link to https://docs.docker.com/articles/security/#docker-daemon-attack-surface

docs/index.md
```diff
@@ -69,71 +80,55 @@ The easiest way to get started with Swarm is to use the
...
-The following example uses the Docker Hub based `token` discovery service:
-
-1. Create a Swarm cluster using the `docker` command.
+    Start the Docker daemon with the `-H` flag. This ensures that the Docker remote API
+    is available over TCP to the *Swarm Manager*.
 
     ```bash
-    $ docker run --rm swarm create
-    6856663cdefdec325839a4b7e1de38e8 # <- this is your unique <cluster_id>
+	$ docker -H tcp://0.0.0.0:2375 -d
```
thaJeztah: I wonder if the example should show how to set up a non-secure API. Even though the warning is added above, users will still blindly follow the guide and end up with an insecure set-up.

/cc @ewindisch @diogomonica

docs/index.md
```diff
@@ -145,9 +140,57 @@ illustrates how to do this:
...
+## Discovery (optional)
+
+Instead of providing a comma-separated list of machines, Swarm can be configured to
+dynamically discover machines.
+
+This allows to add and remove machines from the cluster without having to reconfigure and restart
+the Swarm manager.
+
+For this to work, each machine will run a Swarm agent. The agent registers the referenced
+Docker daemon, monitors it, and updates the discovery backend with the machine's status.
```
thaJeztah: Not sure, but recent docs/communications address the daemon as "Docker Engine". Does that apply here?

docs/index.md
```diff
@@ -145,9 +140,57 @@ illustrates how to do this:
...
+
+Swarm supports a wide range of discovery backends, including the `Docker Hub`, `consul`, `zookeeper` and `etcd`.
```
thaJeztah: please add a comma before "and"

docs/index.md
```diff
@@ -145,9 +140,57 @@ illustrates how to do this:
...
+
+The following example uses the Docker Hub based `token` discovery service:
+
+1. Create a Swarm cluster using the `docker` command.
+
+    ```bash
```
thaJeztah: Might want to check if the Docker docs support this (i.e. the "bash" / language hint), it's not a standard Markdown feature, so it may not work (haven't checked)

docs/index.md
```diff
@@ -145,9 +140,57 @@ illustrates how to do this:
...
+    The `create` command returns a unique cluster ID (`cluster_id`). You'll need
+    this ID when starting the Swarm agent on a machine.
+
+2. Log into **each machine** and register the Swarm agents to the discovery service.
+
+    The machine's IP must be accessible from the Swarm Manager.
+    Use the following command and replace with the proper `machine_ip` and `cluster_id` to start an agent:
+
+        ```bash
+        docker run -d swarm join --addr=<machine_ip:2375> token://<cluster_id>
```
thaJeztah: Should the port number be fixed `2375` or a generic `:machine_port`? (Thinking out loud here)

docs/index.md
```diff
@@ -69,71 +80,55 @@ The easiest way to get started with Swarm is to use the
...
 
-        ```bash
-        docker run -d swarm join --addr=<node_ip:2375> token://<cluster_id>
-        ```
+      ```bash
+    	docker run -d -p <manager_port>:2375 swarm manage nodes://<first_machine_address>,<second_machine_addres>,<third_machine_address>,...
+      ```
 
-        For example:
+    The machines IP must be accessible from the manager.
```
thaJeztah: IP address

docs/index.md
```diff
@@ -145,9 +140,57 @@ illustrates how to do this:
...
+    $ docker run --rm swarm create
+    6856663cdefdec325839a4b7e1de38e8 # <- this is your unique <cluster_id>
+    ```
+
+    The `create` command returns a unique cluster ID (`cluster_id`). You'll need
+    this ID when starting the Swarm agent on a machine.
+
+2. Log into **each machine** and register the Swarm agents to the discovery service.
+
+    The machine's IP must be accessible from the Swarm Manager.
```
thaJeztah: IP address

docs/index.md
```diff
@@ -49,16 +52,24 @@ containers on the same host. For details, see
...
-Follow the set up below to ensure you configure your nodes correctly for this
-behavior.
+By default, Docker listens on `unix:///var/run/docker.sock` to allow only local connections.
 
-> **Note**: Swarm is currently in beta, so things are likely to change. We
-> don't recommend you use it in production yet.
+To enable communication between Swarm and Docker running on different machines,
+each Docker must be 
+[configured] (https://docs.docker.com/articles/basics/#bind-docker-to-another-hostport-or-a-unix-socket)
+to listen to a TCP port.
```
thaJeztah: to listen *on* a TCP port? (Not sure)

aluzzardi:
Thanks @thaJeztah, this is an early draft and I'm more concerned about the content rather than the presentation.

It's a big change to the way we introduce Swarm to the public (skip the discovery step and make it optional)

thaJeztah:
@aluzzardi cool - didn't know what kind of review you were looking for. Was reading and adding notes along, so no problem.

aluzzardi:
@thaJeztah Thanks, those are useful! I'm definitely not the best docs writer out there :)

Any feedback on skipping the discovery step?

docs/index.md
```diff
@@ -69,71 +80,55 @@ The easiest way to get started with Swarm is to use the
...
-    1. Start the Docker daemon with the `-H` flag. This ensures that the Docker remote API on *Swarm Agents* is available over TCP for the *Swarm Manager*.
-
-        ```bash
-    		$ docker -H tcp://0.0.0.0:2375 -d
-        ```
+2. Start the Swarm manager on any machine (even on your laptop).
 
-    2. Register the Swarm agents to the discovery service. The node's IP must be accessible from the Swarm Manager. Use the following command and replace with the proper `node_ip` and `cluster_id` to start an agent:
+    The simplest way to set up the Swarm manager is to provide a comma-separated
+    list of machine addresses:
```
thaJeztah: Perhaps make a mention that, although using a static list of IP-addresses is the *easiest* way, in most cases it's not the recommended way. Add that a Discovery service is more flexible and "to set up a Discovery service, see [link to discovery section]"

docs/index.md
```diff
@@ -145,9 +140,57 @@ illustrates how to do this:
     	$ unset DOCKER_TLS_VERIFY
       ```
 
+## Discovery (optional)
```
thaJeztah: "Setting up a Discovery service" (or similar?)

Not sure if the "(optional)" should be added; I "get" why you added it, but if the introduction text is clear enough, it may be redundant.

Perhaps repeat here that in most cases, a comma-separated list is not the recommended way, and (I think) mainly for testing/development setups.

thaJeztah:
@aluzzardi added some quick notes.

In general, I think this makes sense; making it easier for people to get Swarm up and running will surely benefit the project. "Hey, let's test this thing and see what it does".

It would even be more awesome if "Machine" was added, especially: https://github.com/docker/machine/pull/715
Perhaps even using Docker-in-docker to set up "fake" Swarm nodes on a single machine (never got around myself to see if, and how that would work); A "getting started with Swarm in under 5 minutes", and move from there.

PR #755, opened 2015-05-08
Proposal Node failover

smothiki:
Signed-off-by: sivaram mothiki <smothiki@engineyard.com>

When a node fails Swarm has to reschedule the running containers from the failed node.
Considerations:
1. A cluster should know which container is running on which node so created meta container type struct in cluster. meta container stores container config which is used to create containers and current engine its running on and previous engine it ran on. For the first time creation of container Current and previous are same.
2. Created start method in cluster which in-turn calls start method in engine to start containers after creation.
3. Only Running containers should be started. So refreshloop always calls  refreshContainers(true) which updates containers info every time.
4. Cluster maintains a map of healthy and unhealthy nodes. when a healthy node becomes unhealthy cluster moves the healthy node to unhealthy nodes and calls failover which reschedules the containers to different nodes.
5. when an unhealthy node becomes healthy. calls adjust method which removes the containers that are rescheduled to avoid conflict.

This is a proposal PR and is open for discussions and Design suggestions.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -274,9 +274,9 @@ func (e *Engine) refreshLoop() {
 		var err error
 		select {
 		case <-e.ch:
-			err = e.refreshContainers(false)
+			err = e.refreshContainers(true)
 		case <-time.After(stateRefreshPeriod):
-			err = e.refreshContainers(false)
+			err = e.refreshContainers(true)
 		}
 
 		if err == nil {
@@ -391,6 +391,14 @@ func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool) (*
 	defer e.RUnlock()
 
 	return e.containers[id], nil
+}
+
+//Start starts a docker container
+func (e *Engine) Start(name string) error {
+	if err := e.client.StartContainer(name, nil); err != nil {
+		return err
+	}
+	return nil
 }
 
 // Destroy and remove a container from the engine.
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -7,6 +7,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/docker/pkg/units"
@@ -22,11 +23,19 @@ import (
 type Cluster struct {
 	sync.RWMutex
 
-	eventHandler cluster.EventHandler
-	engines      map[string]*cluster.Engine
-	scheduler    *scheduler.Scheduler
-	options      *cluster.Options
-	store        *state.Store
+	eventHandler   cluster.EventHandler
+	engines        map[string]*cluster.Engine
+	scheduler      *scheduler.Scheduler
+	options        *cluster.Options
+	store          *state.Store
+	metaContainers map[string]*metaContainer
+}
+
+type metaContainer struct {
+	Name    string
+	Config  cluster.ContainerConfig
+	Current *cluster.Engine
+	Prev    *cluster.Engine
 }
 
 // NewCluster is exported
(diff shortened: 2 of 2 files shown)


smothiki:
docker swarm evetns 

1. created mysql on deis-03
2. made deis-03 failvoer 
3. re schedule mysql on deis-02
4. reconnect deis-03 and remove mysql as it has already been rescheduled
```
2015-04-22T20:23:35.000000000Z 8fbda0e636db3f8bf26042677a6f618efccfd26d124b91635ca6ba2f0d877eab: (from mysql:5 node:deis-03) create
2015-04-22T20:23:35.000000000Z 8fbda0e636db3f8bf26042677a6f618efccfd26d124b91635ca6ba2f0d877eab: (from mysql:5 node:deis-03) start
2015-04-22T20:24:00.000000000Z (from swarm node:deis-03) engine_disconnect
2015-04-22T20:24:18.000000000Z 9e22bb2c8fd11c346e8024f81c31ad10eca59f2c8321f7ce454ab02e3213e6f4: (from mysql:5 node:deis-02) create
2015-04-22T20:24:18.000000000Z 9e22bb2c8fd11c346e8024f81c31ad10eca59f2c8321f7ce454ab02e3213e6f4: (from mysql:5 node:deis-02) start
2015-04-22T20:25:56.000000000Z (from swarm node:deis-03) engine_reconnect
2015-04-22T20:25:58.000000000Z 8fbda0e636db3f8bf26042677a6f618efccfd26d124b91635ca6ba2f0d877eab: (from mysql:5 node:deis-03) destroy
```
During deis-03 disconnect event mysql in deis-03 shows pending 
```
core@deis-04 ~ $ docker -H 172.17.8.103:2395 ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS                  PORTS               NAMES
9e22bb2c8fd1        mysql:5             "/entrypoint.sh mysq   6 seconds ago       Up Less than a second   3306/tcp            deis-02/some-mysql    
8fbda0e636db        mysql:5             "/entrypoint.sh mysq   49 seconds ago      Pending                 3306/tcp            deis-03/some-mysql   
```
During deis-03 reconnect event mysql in deis-03 is removed as it is already running in deis-02
```
core@deis-04 ~ $ docker -H 172.17.8.103:2395 ps
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS              PORTS               NAMES
9e22bb2c8fd1        mysql:5             "/entrypoint.sh mysq   2 minutes ago       Up 2 minutes        3306/tcp            deis-02/some-mysql    
4c93f95213cc        redis:latest        "/entrypoint.sh redi   About an hour ago   Up About an hour    6379/tcp            deis-02/some-redis3   
16ee2fe9f663        redis:latest        "/entrypoint.sh redi   18 hours ago        Up 18 hours         6379/tcp            deis-02/some-redis 
```

aluzzardi:
Thanks for contributing this, @smothiki.

Node failover is something we care about (I actually have a rebalancing branch here: https://github.com/aluzzardi/swarm/tree/rebalancing).

I have a few concerns regarding rebalancing, and that's why it's not in master yet. The same concerns apply to this proposal.

* **Relying on `name` to identify containers**. This approach is unreliable as names are not guaranteed to be unique across the fleet and also a `docker rename` could have happened (either through swarm or locally on the machine). To address the issue, we are implementing `Swarm ID`s: every container gets a globally unique ID set as a label. PR #745

* **In-memory state**. This proposal keeps health state in-memory through `metaContainer`. This means two things: It won't work if you restart Swarm and it won't work on a multi-master setup, which we'll be adding soon. To address this, we are adding a store to which Swarm will be able to add ContainerConfig, Names, and all the other stuff needed to re-create containers. PR #756

* **Enabled by default**. This proposal will rebalance **all** containers. This means that a stateful container (such as `mysql`) will get recreated from scratch if the node goes down, with no data in it.

abronan:
Replaced by #1578 (with some elements introduced by Node Management in #1569)

Thank you @smothiki for opening this PR in the first place.

PR #756, opened 2015-05-08
Change existing discovery backends to 'kv' using metatada storage backends in a new 'store' package

abronan:
This PR creates a new `store` package that contains metadata storage backends for *Consul*, *Etcd* and *Zookeeper*. These are unified behind a common `Store` interface.

We can now use these storage backends to store metadata about the cluster and finally have some sort of state replication (if we use a *Consul*, *Etcd* or *Zookeeper* cluster behind a Swarm setup).

This allows to add any kind of metadata storage that can mostly implement this interface (like Riak, Scalaris, or any other distributed KV store).

This is a huge PR, this needs proper testing and review first. There are also a few minor *TODOs* left.


abronan:
Test will fail because I kept *_test.go files that points to local instances of *Consul*, *Etcd* and *Zookeeper* referencing some old stuff...

abronan:
Also this broke a few things for `node`, `token` and `file` discovery. This is because I changed the `Initialize` method to accept a `DiscoveryTLS`, and also the `CreateEntries` to take arrays of `[]byte` instead of arrays of `string`. I'm actually working on correcting this part and all the associated tests and changing the documentation accordingly.

abronan:
golint :facepunch:

And zookeeper fails but that was to be expected :ghost:

cluster/options.go
```diff
@@ -4,8 +4,16 @@ import "crypto/tls"
...
 type Options struct {
-	TLSConfig       *tls.Config
+	TLS             *TLSConfig
 	OvercommitRatio float64
 	Discovery       string
 	Heartbeat       uint64
 }
+
+// TLSConfig is exported
+type TLSConfig struct {
```
aluzzardi: What's the point of this? `tls.Config` contains a ready to use config

cluster/options.go
```diff
@@ -4,8 +4,16 @@ import "crypto/tls"
 
 // Options is exported
 type Options struct {
```
aluzzardi: We probably most definitely don't want to mix Docker certs and storage certs.

store/README.md
```diff
@@ -0,0 +1,145 @@
...
+	WatchRange(prefix string, filter string, heartbeat uint64, callback WatchCallback) error
+	CancelWatchRange(prefix string) error
+	AtomicPut(key string, oldValue []byte, newValue []byte, index uint64) (bool, error)
+	AtomicDelete(key string, oldValue []byte, index uint64) (bool, error)
+}
+```
+
+You can get inspiration from existing backends to create a new one. This interface could be subject to changes to improve the experience of using the library.
+
+### Put
```
aluzzardi: Nit: That information is already in our godoc (or should be). We can duplicate it here, but it will become obsolete pretty soon :)

cluster/options.go
abronan: the etcd go client does not use that `tls.Config` unfortunately and builds its own given cert files :-(

cluster/options.go
abronan: Yes that was what I was thinking too, we need to be able to use certs separately (even though I don't know what it's going to look like in terms of UI with the `cli`)

mavenugo:
@abronan Thanks. This is useful. Is it reasonable to ask to move the `store` package under `pkg` ?
The concept of having a Facade like the store is quiet useful for other projects as well and hence having this in a `pkg` makes sense and in the future, we can even move it out to another repo if it make sense.

store/README.md
abronan: Yes true those informations are redundant ;)

cluster/options.go
vieux: @abronan opens a PR on the etcd repo :D

abronan:
@mavenugo The idea to put this into swarm is to let the lib mature and incubate here first. As much as I'd like to see this in its own repository for simplicity and let other projects use it, there are still a lot of changes and probably bug fixes to do here first. This is also to avoid having to update `Godep` each time we fix a bug in the external repository, which is not super practical for now :( 

Also we benefit from the integration tests and CI tools already in place in swarm to check for bugs and regressions with storage backends. I think we'll definitely consider moving this out of swarm, I'm not sure when exactly.

vieux:
@abronan can't you move it to `/pkg/store` ? We still get all the advantages of having it in swarm but it's easier to import. (it's just a best practice, it doesn't change anything)

abronan:
@vieux @mavenugo Oh I see now. Yes sure I'll move it to `pkg/store`, this way it's separated from the rest and we can move this out easily when we want to.

mavenugo:
@abronan My request is to move it to pkg/store internal to swarm :-) It just gives an indication that this is a candidate for future repository if there is a need.

mavenugo:
@abronan Thanks !

cluster/options.go
abronan: @vieux yes that would be cool if we could feed the etcd `http.Transport` with a `tls.Config` directly. As of now I think properties and methods doing this are private and the only external point of setup for TLS are path to cert files.

abronan:
@mavenugo @vieux done! `image affinity` and `zookeeper` tests in Jenkins keep failing for unrelated reasons. So I guess we'll wait for those to be fixed first ;-)

aluzzardi:
@abronan Do you mind to do the cleanup (using an Options struct or something like that) directly into this PR?

@vieux Once that is done, could you do a review?

Thanks

abronan:
@aluzzardi No that was planned, I also need to change a few other things in this PR directly:

- using `time.Duration` instead of raw `uint64` for the `Watch()` call.
- returning an event channel for the client to check instead of giving the callback directly when calling `Watch`
- discovered an issue with Consul which tries to register entries twice at first call of `swarm manage` thus creating a lot of fuss with the output (although it's working fine and the machine are registered). Might be related to the default timeout value.

I'll ping you when it's ready to be reviewed more thoroughly.

cluster/options.go
abronan: @vieux there is a `SetTransport()` method in the client to override the default `http.Transport`, so we good :)

aluzzardi:
Awesome! Would you mind rebasing? We just merged #763 with etcd integration tests, that would be useful.

abronan:
@aluzzardi rebased and still have to fix zookeeper with Jenkins somehow (although testing zookeeper on a real cluster works..). Don't worry about the amount of commits, I'll squash everything when I figure it out.

pkg/store/README.md
```diff
@@ -0,0 +1,91 @@
...
+type Store interface {
+	Put(key string, value []byte) error
+	Get(key string) (value []byte, lastIndex uint64, err error)
+	Delete(key string) error
+	Exists(key string) (bool, error)
+	Watch(key string, ttl uint64, callback WatchCallback) error
+	CancelWatch(key string) error
+	Acquire(key string, value []byte) (string, error)
+	Release(session string) error
+	GetRange(prefix string) (value [][]byte, err error)
```
mavenugo: In addition to the [] of values, shouldn't `Range` APIs also return [] of actual `keys` and [] of `lastindex` for each of those keys ?
This will bring the GetRange more usable. Without the corresponding key and lastIndex, the application cannot make much use of just the [] of values.

pkg/store/store.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	"time"
+
+	log "github.com/Sirupsen/logrus"
+)
+
+// WatchCallback is used for watch methods on keys
+// and is triggered on key change
+type WatchCallback func(value [][]byte)
```
mavenugo: Shouldnt the WatchCallback carry [] of `Keys` and [] of `Index` ?

pkg/store/store.go
```diff
@@ -0,0 +1,85 @@
...
+	CancelWatch(key string) error
+
+	// Acquire the lock at key
+	Acquire(key string, value []byte) (string, error)
+
+	// Release the lock at key
+	Release(session string) error
+
+	// Get range of keys based on prefix
+	GetRange(prefix string) (value [][]byte, err error)
```
mavenugo: In addition to the [] of values, shouldn't Range APIs also return [] of actual keys and [] of lastindex for each of those keys ?
This will bring the GetRange more usable. Without the corresponding key and lastIndex, the application cannot make much use of just the [] of values.

pkg/store/consul.go
```diff
@@ -0,0 +1,283 @@
...
+			option := &api.QueryOptions{
+				WaitIndex: watch.LastIndex,
+				WaitTime:  watch.Interval}
+			_, meta, err := s.client.KV().Get(key, option)
+			if err != nil {
+				log.WithField("name", "consul").Errorf("Discovery error: %v", err)
+				break
+			}
+			watch.LastIndex = meta.LastIndex
+			ch <- watch.LastIndex
```
mavenugo: I think it is better to add a `if watch.LastIndex != meta.LastIndex` around the above 2 lines, to avoid triggering a watch event if the index has not changed.

pkg/store/consul.go
```diff
@@ -0,0 +1,283 @@
...
+
+	for _ = range eventChan {
+		log.WithField("name", "consul").Debug("Key watch triggered")
+		values, err := s.GetRange(prefix)
+		if err != nil {
+			log.Error("Cannot refresh keys with prefix: ", prefix, ", cancelling watch")
+			s.watches[prefix] = nil
+			return err
+		}
+		callback(values)
```
mavenugo: Is there a way we can avoid updating the values in bulk and callback only on those values that have actually undergone some change ?
By providing a `callback(values)`, every single KV pair that falls under the `prefix`/`filter` will be part of the callback even though they haven't gone through any change.

mavenugo:
@abronan @aluzzardi BTW, The above comments are NOT blockers for this PR. The sooner we merge, the better it is for everyone to collaborate and raise PRs.
(I have a few patches myself).

pkg/store/store.go
abronan: +1

pkg/store/consul.go
abronan: At first I thought that was the issue with Consul refreshing the key like crazy at first startup of the manager with Consul, but it happens that I was not doing a Get to retrieve the LastIndex first. So the LastIndex in my watch was initialized to 0 and during the key watch, Consul would catch up with all the past updates and trigger a watch for each index that it missed. The check is not really necessary here, as the two values will almost always be different. But you helped me pinpoint the  root cause of the issue! :smiley:

abronan:
@aluzzardi @mavenugo I think we are good for now, there will always be ways to improve this code anyway as I was mostly trying to make the 3 backends play nicely behind a common interface. All store should be working as expected now (I corrected the issue with Consul `Watch` and rolled back using simple `Config` structs).

pkg/store/store.go
abronan: I think there is only one index for the whole tree/key so in the case of the WatchCallback returning this index could be sufficient?

abronan:
@aluzzardi Also, a good idea would be to add a test for multiple levels of directories: one for `/test` and another one for `/discovery/test`. This will ensure that we don't break things in key formatting. What do you think?

mavenugo:
@abronan Thanks for the fixes. We can take care of other API changes after this is merged.

discovery/kv/kv.go
```diff
@@ -0,0 +1,93 @@
...
+	s.prefix = parts[1]
+
+	// Creates a new store, will ignore options given
+	// if not supported by the chosen store
+	s.store, err = store.CreateStore(
+		s.name, // name of the store
+		addrs,
+		store.Config{
+			TLS:     tls.TLSConfig,
+			Timeout: s.heartbeat,
```
aluzzardi: I think this is wrong

pkg/store/README.md
```diff
@@ -0,0 +1,91 @@
+---
```
aluzzardi: I don't think we need this - metadata is meant for Swarm documentation page (user guide), but `pkg/` won't be included in it (I think?)

pkg/store/README.md
```diff
@@ -0,0 +1,91 @@
...
+	AtomicDelete(key string, oldValue []byte, index uint64) (bool, error)
+}
+```
+
+To be elligible as a **discovery backend** only, a K/V store implementation should at least offer `Get`, `Put`, `WatchRange`, `GetRange`.
+
+You can get inspiration from existing backends to create a new one. This interface could be subject to changes to improve the experience of using the library and contributing to a new backend.
+
+
+## Docker Swarm documentation index
```
aluzzardi: Same here - no need to mention Swarm

discovery/kv/kv.go
abronan: +1 We should have two flags: 
* 1 for the timeout of the client connection
* 1 for the heartbeat/wait time value of the range key

pkg/store/consul.go
```diff
@@ -0,0 +1,301 @@
...
+		log.Errorf("Couldn't initialize consul client..")
+		return nil, err
+	}
+	s.client = client
+
+	return s, nil
+}
+
+// SetTLS sets Consul TLS options
+func (s *Consul) SetTLS(tls *tls.Config) {
```
aluzzardi: This should be private right?

pkg/store/consul.go
```diff
@@ -0,0 +1,301 @@
...
+	s.config.HttpClient.Transport = &http.Transport{
+		TLSClientConfig: tls,
+	}
+	s.config.Scheme = "https"
+}
+
+// SetTimeout sets the timout for connecting to Consul
+func (s *Consul) SetTimeout(time time.Duration) {
```
aluzzardi: Same for this one

pkg/store/consul.go
aluzzardi: Same comment for other stores - if it's not in the interface, it should be private unless good reasons not to

cluster/options.go
vieux: do we still need the `type TLSConfig struct` then ? I think we can just use `Config *tls.Config`

cli/join.go
```diff
@@ -26,7 +26,9 @@ func join(c *cli.Context) {
 		log.Fatal("--heartbeat should be an unsigned integer and greater than 0")
 	}
 
-	d, err := discovery.New(dflag, hb)
+	// FIXME Add TLS
+	tls := &discovery.TLS{TLSConfig: nil}
```
vieux: when we remove the `type TLSConfig struct` we will be able to use 

`d, err := discovery.New(dflag, hb, nil)`

cluster/swarm/cluster.go
```diff
@@ -54,6 +55,9 @@ func NewCluster(scheduler *scheduler.Scheduler, store *state.Store, options *clu
 		}
 		cluster.newEntries(entries)
 
+		// FIXME remove
+		// time.Sleep(5 * time.Second)
```
vieux: it's already commented ? should we remove it ?

discovery/discovery.go
```diff
@@ -15,6 +16,11 @@ type Entry struct {
 	Port string
 }
 
+// TLS is exported
+type TLS struct {
```
vieux: do we still need the type TLS struct ? I think we can just use Config *tls.Config

discovery/discovery.go
```diff
@@ -73,12 +79,12 @@ func parse(rawurl string) (string, string) {
 }
 
 // New is exported
-func New(rawurl string, heartbeat uint64) (Discovery, error) {
+func New(rawurl string, heartbeat uint64, tls *TLS) (Discovery, error) {
```
vieux: Since currently none of ou discovery service supports TLS, can we not change the prototype of this func but change it when we will add TLS ?

discovery/discovery.go
```diff
@@ -33,7 +39,7 @@ type WatchCallback func(entries []*Entry)
 
 // Discovery is exported
 type Discovery interface {
-	Initialize(string, uint64) error
+	Initialize(string, uint64, *TLS) error
```
vieux: Since currently none of ou discovery service supports TLS, can we not change the prototype of this func but change it when we will add TLS ?

cli/cli.go
```diff
@@ -82,7 +82,7 @@ func Run() {
 					log.Fatalf("discovery required to list a cluster. See '%s list --help'.", c.App.Name)
 				}
 
-				d, err := discovery.New(dflag, 0)
+				d, err := discovery.New(dflag, 10)
```
aluzzardi: Please add a `FIXME` here

cli/manage.go
```diff
@@ -125,7 +125,7 @@ func manage(c *cli.Context) {
 		log.Fatal("--heartbeat should be an unsigned integer and greater than 0")
 	}
 	options := &cluster.Options{
-		TLSConfig:       tlsConfig,
```
vieux: why change the name ?

abronan: Will revert to its initial name, this a the result of tinkering with TLS in discovery.

cli/cli.go
abronan: OK

cli/manage.go
vieux: @abronan thanks

aluzzardi:
LGTM

vieux:
LGTM

vieux:
:fireworks:  :fireworks:  :fireworks:

abronan:
Thanks everyone for the thorough review :+1: :balloon:

mavenugo:
:+1:  Thanks @abronan  @aluzzardi @vieux

PR #763, opened 2015-05-11
etcd Integration test

snrism:
Similar to other discovery service, this PR addresses the integration test for etcd discovery please refer to https://github.com/docker/swarm/issues/644

### DIFF
--- a/test/integration/discovery-etcd.bats
+++ b/test/integration/discovery-etcd.bats
@@ -1,0 +1,52 @@
+#!/usr/bin/env bats
+
+load helpers
+
+# Address on which Etcd will listen (random port between 9000 and 10,000).
+ETCD_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 9000 ))
+
+# Container name for integration test
+CONTAINER_NAME=swarm_etcd
+
+function check_leader() {
+	# Confirm Cluster leader election
+	docker_host logs $CONTAINER_NAME | grep "state changed from 'follower' to 'leader'"
+	# Check leader event
+	docker_host logs $CONTAINER_NAME | grep "leader changed from '' to"
+}
+
+function start_etcd() {
+	docker_host run -p $ETCD_HOST:4001 --name=$CONTAINER_NAME -d coreos/etcd
+	# Check if etcd cluster leader is elected
+	retry 30 1 check_leader
+}
+
+function stop_etcd() {
+	docker_host rm -f -v $CONTAINER_NAME
+}
+
+function setup() {
+	start_etcd
+}
+
+function teardown() {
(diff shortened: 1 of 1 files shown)


snrism:
@aluzzardi  In the `check_leader` function, I wanted to check if leader changed to swarm_etcd in the logs, but when I look into the logs, the $CONTAINER NAME is not picked up in the logs, rather only displays the container_id. Please refer to the last line in the following etcd log.

[etcd] May 11 18:44:37.652 WARNING   | Using the directory d50bb34a126b.etcd as the etcd curation directory because a directory was not specified.
[etcd] May 11 18:44:37.654 INFO      | d50bb34a126b is starting a new cluster
[etcd] May 11 18:44:37.662 INFO      | etcd server [name d50bb34a126b, listen on :4001, advertised url http://127.0.0.1:4001]
[etcd] May 11 18:44:37.663 INFO      | peer server [name d50bb34a126b, listen on :7001, advertised url http://127.0.0.1:7001]
[etcd] May 11 18:44:37.663 INFO      | d50bb34a126b starting in peer mode
[etcd] May 11 18:44:37.664 INFO      | d50bb34a126b: state changed from 'initialized' to 'follower'.
[etcd] May 11 18:44:37.665 INFO      | d50bb34a126b: state changed from 'follower' to 'leader'.
[etcd] May 11 18:44:37.665 INFO      | d50bb34a126b: leader changed from '' to 'd50bb34a126b'.

If we need to check for the leader change as well, then one way to do this is either use `docker_host ps` or `inspect` to get the Container ID and then use that to check as follows: 

docker_host logs $CONTAINER_NAME | grep "leader changed from '' to 'd50bb34a126b'"

Any thoughts?

aluzzardi:
@snrism Thanks for your work.

I think just checking `leader changed from '' to` should be enough, right? We only have one leader.

aluzzardi:
If on the other hand we really need the ID, then this should work:

```bash
function start_etcd() {
	id=$(docker_host run -p $ETCD_HOST:4001 --name=$CONTAINER_NAME -d coreos/etcd)
	# Check if etcd cluster leader is elected
	retry 30 1 check_leader $id
}

function check_leader {
    # "$1" is our leader container ID
}
```

aluzzardi:
Let me know if you think it's required - if it works without that, then it's alright.

snrism:
@aluzzardi thanks for the code snippet. I did think about that, but the container ID returned is the entire ID, while the etcd log gives the truncated one. I guess checking for `leader changed from '' to` should be fine since we only have one leader. Updated the code in the latest version.

aluzzardi:
LGTM

Thanks a lot!

Issue #631, opened 2015-04-16
Add support "docker import" api


jimmyxian:
I will implement the "docker import" api by the following steps .
+ Submit a PR in docker client : Implement the import api
+ Submit a PR in swarm: Send pull request on every machine(same as Load and Pull)

any advice?

HisenWu:
LGTM

vieux:
sound good to me as well

aluzzardi:
Merged

Issue #672, opened 2015-04-25
Cluster needs better unit testing.


aluzzardi:
In order to improve the code coverage of `Cluster`, we might need to create a mock Engine.

Related to #502

jimmyxian:
Great! I think creating a mock Engine help us a lot to write unit test in cluster

amitshukla:
@aluzzardi is this still relevant?

abronan:
@amitshukla yes still relevant cluster needs better unit testing.

nishanttotla:
Closing due to lack of activity. Please reopen if you wish to continue discussing it.

PR #768, opened 2015-05-12
Add support import

jimmyxian:
https://github.com/docker/swarm/issues/631

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -378,7 +378,7 @@ func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)
 }
 
 // Import image
-func (c *Cluster) Import(source string, repository string, tar string, imageReader io.Reader, callback func(what, status string)) {
+func (c *Cluster) Import(source string, repository string, tag string, imageReader io.Reader, callback func(what, status string)) {
 	var wg sync.WaitGroup
 	c.RLock()
 	pipeWriters := []*io.PipeWriter{}
@@ -395,7 +395,7 @@ func (c *Cluster) Import(source string, repository string, tar string, imageRead
 			defer reader.Close()
 
 			// call engine import
-			err := nn.Import(source, repository, tar, reader)
+			err := nn.Import(source, repository, tag, reader)
 			if callback != nil {
 				if err != nil {
 					callback(nn.Name, err.Error())
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -377,6 +377,60 @@ func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)
 	wg.Wait()
 }
 
+// Import image
+func (c *Cluster) Import(source string, repository string, tar string, imageReader io.Reader, callback func(what, status string)) {
+	var wg sync.WaitGroup
+	c.RLock()
+	pipeWriters := []*io.PipeWriter{}
+	pipeReaders := []*io.PipeReader{}
+	for _, n := range c.engines {
+		wg.Add(1)
+
+		pipeReader, pipeWriter := io.Pipe()
+		pipeReaders = append(pipeReaders, pipeReader)
+		pipeWriters = append(pipeWriters, pipeWriter)
+
+		go func(reader *io.PipeReader, nn *cluster.Engine) {
+			defer wg.Done()
+			defer reader.Close()
+
+			// call engine import
+			err := nn.Import(source, repository, tar, reader)
+			if callback != nil {
+				if err != nil {
+					callback(nn.Name, err.Error())
+				} else {
+					callback(nn.Name, "Import success")
+				}
+			}
+
+		}(pipeReader, n)
+	}
+
+	// create multi-writer
+	listWriter := []io.Writer{}
+	for _, pipeW := range pipeWriters {
+		listWriter = append(listWriter, pipeW)
+	}
+	mutiWriter := io.MultiWriter(listWriter...)
+
+	// copy image-reader to muti-writer
+	_, err := io.Copy(mutiWriter, imageReader)
+	if err != nil {
+		log.Error(err)
+	}
+
+	// close pipe writers
+	for _, pipeW := range pipeWriters {
+		pipeW.Close()
+	}
+
+	c.RUnlock()
+
+	wg.Wait()
+
+}
+
 // Containers returns all the containers in the cluster.
 func (c *Cluster) Containers() cluster.Containers {
 	c.RLock()
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -382,12 +382,11 @@ func (c *Cluster) Import(source string, repository string, tag string, imageRead
 	var wg sync.WaitGroup
 	c.RLock()
 	pipeWriters := []*io.PipeWriter{}
-	pipeReaders := []*io.PipeReader{}
+
 	for _, n := range c.engines {
 		wg.Add(1)
 
 		pipeReader, pipeWriter := io.Pipe()
-		pipeReaders = append(pipeReaders, pipeReader)
 		pipeWriters = append(pipeWriters, pipeWriter)
 
 		go func(reader *io.PipeReader, nn *cluster.Engine) {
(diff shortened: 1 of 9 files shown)


jimmyxian:

vieux:
@jimmyxian can you add some unit tests ? I't a pretty big drop in our coverage right now, thanks

vieux:
@jimmyxian can you also update `api/README.md` to say import is done ? Thanks

jimmyxian:
@vieux Done to update README.md
Unit test need to mock engine. https://github.com/docker/swarm/issues/672 track the missing test

aluzzardi:
@jimmyxian I 100% agree with mock engine. However, in the meantime, do you think you could add tests using the real Engine with a mock dockerclient?

It's harder to do but we really try not to drop coverage by that much

jimmyxian:
@aluzzardi 
I have tried to use the real Engine with a mock dockerclient in `load` one month ago.
There is one problem. the `Engine.client` is an unexported field.
Two ways to solve:
* Change `Engine.client` to a exported field
* Change `Engine.connectClient(dockerclient.Client)` to a public function

Any thoughts ?

aluzzardi:
Oh, right! I remember.

Well, yeah, we can make `ConnectClient` public I guess and name it `ConnectWithClient`

jimmyxian:
@aluzzardi 
Greate! I will implement this unit test. Also, implement the `load` unit test as well.

vieux:
@jimmyxian  is it ready to be reviewed ? Thanks.

jimmyxian:
@vieux Not ready. I'm writing the unit test. :)

jimmyxian:
ping @aluzzardi
Unit test failed because data race, I check N times, but not found why. Can you help me take a look? thanks.
```
WARNING: DATA RACE
```

aluzzardi:
I believe we have to lock the mocks call to fix this

jimmyxian:
@aluzzardi Thanks
I have locked mock call, but also 'data race'.
Maybe the mock module reads something racing with `io.Copy(pipeWrtier, file)` / `pipeWriter.Close()`.
The following code is my simple test: I locked both of `mock call` and `io.Copy(pipeWrtier, file)` / `pipeWriter.Close()`. It works.
```
package yours

import (
    "github.com/stretchr/testify/assert"
    "github.com/stretchr/testify/mock"
    "sync"
    "testing"
    "io"
    "bytes"
)

type MyMockedObject struct {
    mock.Mock
}

func (m *MyMockedObject) DoSomething(reader io.Reader) (bool, error) {

    args := m.Called(reader)
    return args.Bool(0), args.Error(1)

}

func TestSomething(t *testing.T) {
    var l sync.Mutex
    var wg sync.WaitGroup
    testObj := new(MyMockedObject)

    pipeReader, pipeWriter := io.Pipe()
    testObj.On("DoSomething", pipeReader).Return(true, nil)
    go func() {
        wg.Add(1)
        defer wg.Done()
        l.Lock()
        result, _ := testObj.DoSomething(pipeReader)
        l.Unlock()
        assert.True(t, result)
    }()

    l.Lock()
    io.Copy(pipeWriter, bytes.NewReader(nil))
    pipeWriter.Close()
    l.Unlock()

    wg.Wait()
}
```

aluzzardi:
I think you only need to lock this:

```go
type MyMockedObject {
    mock.Mock
    sync.Mutex
}

func (m *MyMockedObject) DoSomething(reader io.Reader) (bool, error) {
    m.Lock()
    defer m.Unlock()

    args := m.Called(reader)
    return args.Bool(0), args.Error(1)

}
```

jimmyxian:
@aluzzardi I tried this just now, also 'data race'.

aluzzardi:
@jimmyxian Okay so I managed to fix it.

```go
package yours

import (
    "github.com/stretchr/testify/assert"
    "github.com/stretchr/testify/mock"
    "testing"
    "io"
    "bytes"
)

type MyMockedObject struct {
    mock.Mock
}

func (m *MyMockedObject) DoSomething(reader io.Reader) (bool, error) {
    args := m.Called(reader)
    return args.Bool(0), args.Error(1)

}

func TestSomething(t *testing.T) {
    // FIX: NO NEED TO LOCK
    var wg sync.WaitGroup
    testObj := new(MyMockedObject)

    pipeReader, pipeWriter := io.Pipe()
    testObj.On("DoSomething", pipeReader).Return(true, nil)
    
   // FIX: Add() must be called BEFORE entering the go-routine
    wg.Add(1)
    go func() {
        defer wg.Done()
        l.Lock()
        result, _ := testObj.DoSomething(pipeReader)
        l.Unlock()
        assert.True(t, result)
    }()

    io.Copy(pipeWriter, bytes.NewReader(nil))
    wg.Wait()
    // FIX: Close() must be called AFTER wg.Wait()
    pipeWriter.Close()
}
```

cluster/swarm/cluster.go
```diff
@@ -347,6 +347,60 @@ func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)
...
+
+		}(pipeReader, n)
+	}
+
+	// create multi-writer
+	listWriter := []io.Writer{}
+	for _, pipeW := range pipeWriters {
+		listWriter = append(listWriter, pipeW)
+	}
+	mutiWriter := io.MultiWriter(listWriter...)
```
aluzzardi: no need to create listWriter, this should work:

`multiWriter := io.MultiWriter(pipeWriters...)`

(also, there was a typo in "multi")

jimmyxian: @aluzzardi Yeah. I will modify this.

cluster/swarm/cluster.go
```diff
@@ -347,6 +347,60 @@ func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)
...
+	if err != nil {
+		log.Error(err)
+	}
+
+	// close pipe writers
+	for _, pipeW := range pipeWriters {
+		pipeW.Close()
+	}
+
+	c.RUnlock()
```
aluzzardi: `RUnlock` must happen **immediately** after `for _, n := range c.engines {`.

Otherwise, this will lock the entire cluster down until the entire `io.Copy` is finished

cluster/swarm/cluster.go
```diff
@@ -347,6 +347,60 @@ func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)
...
+
+	wg.Wait()
```
aluzzardi: This needs to happen **before** closing the pipe writers

jimmyxian:
@aluzzardi 
```
    io.Copy(pipeWriter, bytes.NewReader(nil))
    wg.Wait()
    // FIX: Close() must be called AFTER wg.Wait()
    pipeWriter.Close()
```
Actually, I can't follow this in real logic. 
`pipeWriter.Close()` here is used to tell pipeReader EOF, then `import` will be done and exit.
If follow this,  the code will lock forever befcause of `wg.Wait()`

cluster/swarm/cluster.go
jimmyxian: Yeah...Right....

aluzzardi:
Shouldn't `import` be done and exit whenever the import on the engine is actually done?

aluzzardi:
Nevermind!

jimmyxian:
@aluzzardi 
My bad, not say clearly.
`pipeWriter.Close()` here is used to tell pipeReader EOF, then `engine.import`(`dockerclient.import`) will be done and return.

cluster/swarm/cluster.go
jimmyxian: I try modify this, but
`Cannot use pipeWriters (type []*io.PipeWriter) as type []io.Writer in function argument`

jimmyxian:
@aluzzardi 
I try to modify this
FROM
```
 	// create multi-writer
	listWriter := []io.Writer{}
	for _, pipeW := range pipeWriters {
		listWriter = append(listWriter, pipeW)
	}
	mutiWriter := io.MultiWriter(listWriter...)
```
TO
```
	multiWriter := io.MultiWriter(pipeWriters...)
```
But
```
Cannot use pipeWriters (type []*io.PipeWriter) as type []io.Writer in function argument
```

Also, I have fixed the unit test, need code-review again, thanks
cc /@vieux

aluzzardi:
LGTM

@vieux ?

aluzzardi:
Thank you @jimmyxian !

PR #772, opened 2015-05-12
change [][]byte to []store.Entry

vieux:
Also, tiny optimisation, now the `WatchCallback` takes varargs to we don't extra allocate variables.


vieux:
@abronan what do you think ?

abronan:
I think it's better this way :)

We'll probably see if we can turn this into a struct and include the `lastIndex` or useful informations about lock sessions, etc. But it's cleaner this way.

LGTM

/cc @aluzzardi

Issue #633, opened 2015-04-16
Cluster driver specific flags


vieux:
We need each cluster driver to install specific flags in swarm and I have 2 questions: 

CLI
----

* Should we do something simple like docker-machine did:
`swarm --debug manage --tlsverify 1.2.3.4:2375`
`swarm --debug manage --tlsverify -c mesos --mesos-user=root 1.2.3.4:5050`

 * Pros: easy to implement
 * Cons: `swarm manage --help` will display all the flags installed by all the drivers

* Or more complicated like using subcommands:
`swarm --debug manage --tlsverify swarm 1.2.3.4:2375`
`swarm --debug manage --tlsverify mesos -user=root 1.2.3.4:5050`

 * Pros: the help will be cleaner we will have a help per driver.
 * Cons: Not sure `swarm --debug manage --tlsverify 1.2.3.4:2375` will be possible (no swarm) so we might have to change the cli

Swarm only flag
----

I want to test this by moving a manage flag to be swarm only, I was thinking about `--overcommit` but @tnachen told me it could be useful with `mesos` so I am not sure. Maybe `--heartbeat` ?

vieux:
@tnachen @jimenez @abronan @aluzzardi @ehazlett @chanwit @jimmyxian I'd love your feedback!

jimmyxian:
I prefer the subcommands, because the help information is cleaner if we have a lot of cluster drivers.

ehazlett:
+1 for the subcommands.  we (machine) are most likely going to switch to them and in hindsight would have been better from the start.  once you get several drivers, the help text gets almost unusable without `grep` :)

abronan:
+1 for the subcommands and a cleaner help output.

chanwit:
I am currently a docker-machine user with `grep`. So I prefer subcommands too :):)

tnachen:
subcommands is much more ideal as well, from a user standpoint and feels more flexible from a cluster impl to add help and commands.

vieux:
Awesome everyone seems to agree! @aluzzardi ?

aluzzardi:
Any idea of what kind of cluster driver specific flags we'll have?

vieux:
@aluzzardi at least `--mesos-user` `--mesos-port` `--mesos-addr`

vieux:
Here is what I managed to do: `swarm manage <cluster> <flags> <discovery>`
Unfortunately there is no way to no break the current command line if we decide to use subcommands.

So it would give:

[...]

PR #774, opened 2015-05-13
add --cluster-driver and --cluster-opt

vieux:
See #633

### DIFF
--- a/cli/help.go
+++ b/cli/help.go
@@ -25,7 +25,7 @@ ARGUMENTS:
             {{printf "\t"}} * <ip1>,<ip2>{{end}}{{if .Flags}}
 OPTIONS:
    {{range .Flags}}{{.}}
-   {{end}}{{ end }}
+   {{end}}{{if (eq .Name "manage")}}{{printf "\t * swarm.overcommit=0.05\tovercommit to apply on resources\n"}}{{end}}{{ end }}
 `
 
 }
--- a/cli/help.go
+++ b/cli/help.go
@@ -25,7 +25,8 @@ ARGUMENTS:
             {{printf "\t"}} * <ip1>,<ip2>{{end}}{{if .Flags}}
 OPTIONS:
    {{range .Flags}}{{.}}
-   {{end}}{{if (eq .Name "manage")}}{{printf "\t * swarm.overcommit=0.05\tovercommit to apply on resources\n"}}{{end}}{{ end }}
+   {{end}}{{if (eq .Name "manage")}}{{printf "\t * swarm.overcommit=0.05\tovercommit to apply on resources"}}
+                                    {{printf "\t * swarm.heartbeat=25\ttime in second between each heartbeat"}}{{end}}{{ end }}
 `
 
 }
--- a/cli/help.go
+++ b/cli/help.go
@@ -26,7 +26,7 @@ ARGUMENTS:
 OPTIONS:
    {{range .Flags}}{{.}}
    {{end}}{{if (eq .Name "manage")}}{{printf "\t * swarm.overcommit=0.05\tovercommit to apply on resources"}}
-                                    {{printf "\t * swarm.heartbeat=25\ttime in second between each heartbeat"}}{{end}}{{ end }}
+                                    {{printf "\t * swarm.discovery.heartbeat=25\ttime in second between each heartbeat"}}{{end}}{{ end }}
 `
 
 }
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -125,13 +125,16 @@ func manage(c *cli.Context) {
 		log.Fatal("--heartbeat should be an unsigned integer and greater than 0")
 	}
 	options := &cluster.Options{
-		TLSConfig:       tlsConfig,
-		OvercommitRatio: c.Float64("overcommit"),
-		Discovery:       dflag,
-		Heartbeat:       hb,
+		TLSConfig: tlsConfig,
+		Opts:      c.StringSlice("cluster-opt"),
+		Discovery: dflag,
+		Heartbeat: hb,
 	}
 
-	cluster := swarm.NewCluster(sched, store, options)
+	cluster, err := swarm.NewCluster(sched, store, options)
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	// see https://github.com/codegangsta/cli/issues/160
 	hosts := c.StringSlice("host")
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
-	"strconv"
 
 	log "github.com/Sirupsen/logrus"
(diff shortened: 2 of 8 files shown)


vieux:
I opted for less changes as possible, we could refactor the code later

vieux:
ping @aluzzardi

vieux:
ping @jimenez @tnachen does this looks good to you ?

tnachen:
@vieux I see, wrapping everything in cluster-opts is fine, I think this change makes sense, +1

vieux:
I pushed another commit with a couple of changes:

* DriverOpts is now a type with a few getters as suggested by @aluzzardi 
* I figured that the heartbeat is also swarm specific
* I removed the option struct because I hated it since I introduced it
 * TLSConfig is now a parameter
 * discovery is now injected on the fly (it's good because for mesos it will be like `mesos.master`)

@aluzzardi let me know what you think

jimenez:
LGTM

abronan:
LGTM

vieux:
@aluzzardi I added some test, the CI is green now :)

cli/manage.go
```diff
@@ -120,19 +119,11 @@ func manage(c *cli.Context) {
...
-	hb, err := strconv.ParseUint(c.String("heartbeat"), 0, 32)
-	if hb < 1 || err != nil {
-		log.Fatal("--heartbeat should be an unsigned integer and greater than 0")
-	}
-	options := &cluster.Options{
-		TLSConfig:       tlsConfig,
-		OvercommitRatio: c.Float64("overcommit"),
-		Discovery:       dflag,
-		Heartbeat:       hb,
+	cluster, err := swarm.NewCluster(sched, store, tlsConfig, append(c.StringSlice("cluster-opt"), "swarm.discovery="+dflag))
```
aluzzardi: Is discovery `swarm` only? It's kind of weird to pass it like that

cluster/options.go
```diff
@@ -1,11 +1,47 @@
...
+import (
+	"strconv"
+	"strings"
+)
+
+// DriverOpts are key=values options
+type DriverOpts []string
+
+// String returns a string from the driver options
+func (do DriverOpts) String(key string) (bool, string) {
```
aluzzardi: nit: To be consistent with `map`, `bool` should be the last argument:

```go
str, ok := opts.String("foo")
```

cluster/options.go
```diff
@@ -1,11 +1,47 @@
...
+		kv := strings.SplitN(opt, "=", 2)
+		if kv[0] == key {
+			return true, kv[1]
+		}
+	}
+	return false, ""
+}
+
+// Int returns an int64 from the driver options
+func (do DriverOpts) Int(key string) (bool, int64) {
```
aluzzardi: Same

cluster/options.go
```diff
@@ -1,11 +1,47 @@
...
+	if ok, value := do.String(key); ok {
+		v, _ := strconv.ParseInt(value, 0, 64)
+		return true, v
+	}
+	return false, 0
+}
+
+// Uint returns an int64 from the driver options
+func (do DriverOpts) Uint(key string) (bool, uint64) {
```
aluzzardi: Same

cluster/options.go
```diff
@@ -1,11 +1,47 @@
...
+	if ok, value := do.String(key); ok {
+		v, _ := strconv.ParseUint(value, 0, 64)
+		return true, v
+	}
+	return false, 0
+}
+
+// Float returns a float64 from the driver options
+func (do DriverOpts) Float(key string) (bool, float64) {
```
aluzzardi: Same

cluster/options_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+	ok, val = opts.String("foo3")
+	assert.True(t, ok)
+	assert.Equal(t, val, "7")
+
+	ok, val = opts.String("foo4")
+	assert.True(t, ok)
+	assert.Equal(t, val, "0.6")
+
+	ok, val = opts.String("foo5")
```
aluzzardi: tiny little nit, feel free to ignore: Instead of `foo5` I'd use `invalid` or `non-existent`. It's not obvious that `foo5` doesn't exist

cli/manage.go
vieux: yes, we could add another argument to NewCluster to pass it if you think it's better.

cluster/swarm/cluster.go
```diff
@@ -26,24 +27,37 @@ type Cluster struct {
...
+	}
+	_, cluster.discovery = options.String("swarm.discovery")
+	if _, cluster.heartbeat = options.Uint("swarm.heartbeat"); cluster.heartbeat < 1 {
+		return nil, errors.New("heartbeat should be an unsigned integer and greater than 0")
 	}
 
 	// get the list of entries from the discovery service
 	go func() {
-		d, err := discovery.New(options.Discovery, options.Heartbeat)
+		d, err := discovery.New(cluster.discovery, cluster.heartbeat)
```
aluzzardi: Entirely unrelated to this PR: `heartbeat` should be a discovery flag, not a cluster flag.

It only makes sense for `token` (and `file` but it should be called something like `refresh` for that backend)

cli/help.go
```diff
@@ -25,7 +25,8 @@ ARGUMENTS:
             {{printf "\t"}} * <ip1>,<ip2>{{end}}{{if .Flags}}
 OPTIONS:
    {{range .Flags}}{{.}}
-   {{end}}{{ end }}
+   {{end}}{{if (eq .Name "manage")}}{{printf "\t * swarm.overcommit=0.05\tovercommit to apply on resources"}}
```
aluzzardi: We don't have an option currently, but it's a bit bad to have the defaults duplicated. Out of scope with this PR.

cluster/swarm/cluster.go
aluzzardi: On a side note:

This is related to `swarm.discovery` and `swarm.heartbeat`: Does it make sense for cluster to only take a `Discovery`, not actually create it?

Those two options are only useful for discovery not for the cluster

aluzzardi:
Minor nits and comments.

Overall LGTM - some questions regarding the relationship between Discovery and Cluster but those should probably be addressed by other PRs.

cli/help.go
vieux: I agree

cluster/swarm/cluster.go
vieux: I agree, the thing is,  you need `swarm.heartbeat` to create the discovery, so it feels weird to parse a `swarm.` opt outside of swarm, to create the discovery and give it to swarm.

vieux:
@aluzzardi I know it doesn't fix the core issue, but I renamed `swarm.heartbeat` to `swarm.discovery.heartbeat` in the last commit, let me know what you think.

aluzzardi:
LGTM

PR #777, opened 2015-05-13
store: Improved Lock support.

aluzzardi:
- New interface: `store.CreateLock` returns a `Locker` interface with
  `Lock` and `Unlock` functions.
- Consul: New locking implementation leveraging helper lock
  implementation that comes with the go library.
- ZK: Lock support using the library helpers.

### DIFF
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -36,11 +36,10 @@ type Store interface {
 	// Cancel watch key
 	CancelWatch(key string) error
 
-	// Acquire the lock at key
-	Acquire(key string, value []byte) (string, error)
-
-	// Release the lock at key
-	Release(session string) error
+	// CreateLock for a given key.
+	// The returned Locker is not held and must be acquired with `.Lock`.
+	// value is optional.
+	CreateLock(key string, value []byte) (Locker, error)
 
 	// Get range of keys based on prefix
 	GetRange(prefix string) ([]KVEntry, error)
@@ -66,6 +65,13 @@ type KVEntry interface {
 	Key() string
 	Value() []byte
 	LastIndex() uint64
+}
+
+// Locker provides locking mechanism on top of the store.
+// Similar to `sync.Lock` except it may return errors.
+type Locker interface {
+	Lock() error
+	Unlock() error
 }
 
 var (
--- a/pkg/store/zookeeper.go
+++ b/pkg/store/zookeeper.go
@@ -16,6 +16,10 @@ type Zookeeper struct {
 	watches map[string]<-chan zk.Event
 }
 
+type zookeeperLock struct {
+	lock *zk.Lock
+}
+
 // InitializeZookeeper creates a new Zookeeper client
 // given a list of endpoints and optional tls config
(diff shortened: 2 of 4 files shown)


aluzzardi:
There are no tests and I frankly don't even know if this works. This is the basis of leader election and will be tested alongside that in the near future.

I don't like `CreateLock()` as a name but I wanted to emphasize that it only creates the lock it doesn't actually acquire it. Any better name? Perhaps `Mutex()`.

The real issue here is etcd: I don't have a real solution.

- There are a couple of libraries to help do that, however they are not very popular:
  - https://github.com/Scalingo/go-etcd-lock
  - https://github.com/hayesgm/go-etcd-lock
- etcd 0.4 has a lock module, although it doesn't seem to be part of etcd's Go API.

/cc @abronan @vieux @mavenugo

pkg/store/zookeeper.go
```diff
@@ -199,15 +203,22 @@ func (s *Zookeeper) AtomicDelete(key string, oldValue []byte, index uint64) (boo
...
-func (s *Zookeeper) Acquire(path string, value []byte) (string, error) {
-	// lock := zk.NewLock(s.client, path, nil)
-	// locks[path] = lock
-	// lock.Lock()
-	return "", ErrNotImplemented
+// CreateLock returns a handle to a lock struct which can be used
+// to acquire and release the mutex.
+func (s *Zookeeper) CreateLock(key string, value []byte) (Locker, error) {
+	// FIXME: `value` is not being used since there is no support in zk.NewLock().
+	return &zookeeperLock{
```
aluzzardi: Self-comment: This is useless for now, I could just return the result of `NewLock` since it adheres with our `Locker` interface.

Initially, `Locker.Lock()` returned a channel as well to notify of lost lock (this is supported by Consul).
Removed for now for simplicity, will need to add back soon.

pkg/store/store.go
```diff
@@ -36,11 +36,10 @@ type Store interface {
...
 
-	// Acquire the lock at key
-	Acquire(key string, value []byte) (string, error)
-
-	// Release the lock at key
-	Release(session string) error
+	// CreateLock for a given key.
+	// The returned Locker is not held and must be acquired with `.Lock`.
+	// value is optional.
+	CreateLock(key string, value []byte) (Locker, error)
```
tnachen: Does this means each key will have a seperate lock, and while the lock is held no one else can CreateLock again on the same key?

vieux: ping @aluzzardi

aluzzardi:
```go
// Anyone can create the same lock even if the lock is held
l1, _ := store.CreateLock("/foo)
l1.Lock()
l2, _ := store.CreateLock("/foo")
l3, _ := store.CreateLock("/foo")
l2.Lock() // this will block until l1 is released
```

aluzzardi:
Rebased.

PTAL

abronan:
LGTM

aluzzardi:
@abronan Can you please hit `delete branch` next time you merge a PR? Thanks :)

abronan:
@aluzzardi :+1:

PR #778, opened 2015-05-13
GetRange and WatchRange to return {key,value,index} tuple

mavenugo:
I tested the watchRange and GetRange for consul backend. Hoping for the integration tests to cover etcd and zookeeper

Signed-off-by: Madhu Venugopal <madhu@docker.com>

### DIFF
--- a/pkg/store/consul.go
+++ b/pkg/store/consul.go
@@ -115,7 +115,7 @@ func (s *Consul) Exists(key string) (bool, error) {
 }
 
 // GetRange gets a range of values at "directory"
-func (s *Consul) GetRange(prefix string) (values [][]byte, err error) {
+func (s *Consul) GetRange(prefix string) (kvi []KVEntry, err error) {
 	pairs, _, err := s.client.KV().List(partialFormat(prefix), nil)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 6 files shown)


pkg/store/zookeeper.go
```diff
@@ -137,17 +137,17 @@ func (s *Zookeeper) CancelWatch(key string) error {
 }
 
 // GetRange gets a range of values at "directory"
-func (s *Zookeeper) GetRange(prefix string) (values [][]byte, err error) {
+func (s *Zookeeper) GetRange(prefix string) (kvi []KVEntry, err error) {
 	prefix = format(prefix)
-	entries, _, err := s.client.Children(prefix)
+	entries, stat, err := s.client.Children(prefix)
```
abronan: Here we can probably rename `entries` to something else like `results` and change the return array with:
`GetRange(prefix string) (entries []KVEntry, err error)`

It'll be more meaningful with the now renamed struct.

pkg/store/consul.go
```diff
@@ -127,9 +127,9 @@ func (s *Consul) GetRange(prefix string) (values [][]byte, err error) {
 		if pair.Key == prefix {
 			continue
 		}
-		values = append(values, pair.Value)
+		kvi = append(kvi, &kviTuple{pair.Key, pair.Value, pair.ModifyIndex})
```
abronan: I'd put a more meaningful name to the array of `kviTuple` like `entries`/`tuples` or `values` instead of `kvi`. Same thing about other methods.

abronan:
replaces #772

abronan:
My comments above are not blocking as we can make those in another PR related to code quality (using the same blueprint for variable names everywhere).

Overall LGTM

ping @aluzzardi @vieux

vieux:
LGTM

PR #780, opened 2015-05-13
Refactor helpers for storage key formatting

abronan:
Refactor helpers to have more meaningful names and be more flexible with future usage.

I kept the approach with a boolean to omit the first `/` or not. Because we need this for the `getDirectory` function and I'd like to avoid having a private `normalize` and `getDirectory` functions for each store that needs to omit the first `/` in the future.

### DIFF
--- a/pkg/store/helpers.go
+++ b/pkg/store/helpers.go
@@ -12,21 +12,22 @@ func createEndpoints(addrs []string, scheme string) (entries []string) {
 	return entries
 }
 
-// Formats the key
-func format(key string) string {
-	return fullpath(splitKey(key))
+// Normalize the key for each store to the form:
+//
+//     /path/to/key
+//
+func normalize(key string) string {
+	return "/" + join(splitKey(key))
 }
 
-// Formats the key partially (omits the first '/')
-func partialFormat(key string) string {
-	return partialpath(splitKey(key))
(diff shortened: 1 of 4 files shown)


pkg/store/helpers.go
```diff
@@ -12,21 +12,38 @@ func createEndpoints(addrs []string, scheme string) (entries []string) {
...
+// Normalize the key for each store to the form:
+//
+//     /path/to/key
+//
+// If omit is set to 'true', this will omit
+// the first '/' character:
+//
+//     path/to/key/
+//
+func normalize(key string, omit bool) string {
```
aluzzardi: I wouldn't put omit in there probably, it's not useful to anyone but consul.

abronan: I'm worried that some K/V (only if we consider adding more backends in the future) may forbid using special characters like `.`, `-`, and especially `/` as the first character of a Key. Either way this is a cosmetic change so I don't mind changing it to have a private method in Consul for now :)

aluzzardi: We'll do that once we get there :)

So far, no new backends are planned

vieux:
@abronan please rebase :P

aluzzardi:
LGTM

PR #782, opened 2015-05-13
Call SyncCluster at etcd client initialization

abronan:

### DIFF
--- a/pkg/store/etcd.go
+++ b/pkg/store/etcd.go
@@ -34,6 +34,8 @@ func InitializeEtcd(addrs []string, options Config) (Store, error) {
 		s.setTimeout(options.Timeout)
 	}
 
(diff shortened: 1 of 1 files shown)


abronan:
ping @aluzzardi

aluzzardi:
We'll have to create integration tests that set up a replicated environment :/

I'm pretty sure if you kill the etcd leader our discovery becomes fubar.

Please add a FIXME on top of it

LGTM

abronan:
@aluzzardi Yes either we refresh on each operation (overkill but we are pretty much sure we have an up to date state) or we sync periodically in a goroutine. Not sure what is the good practice with etcd.

Added the FIXME.

aluzzardi:
LGTM

PR #785, opened 2015-05-14
APIVersion -> ApiVersion to match docker API

### LINKED ISSUE #687: Docker and Swarm ReST APIs differ
Hello guys,

while we (@holgerkoser and me) were trying to make https://github.com/cf-platform-eng/docker-boshrelease work with Swarm instead of Docker (by swapping it under the hood), we noticed one compatibility issue between the two ReST APIs (Swarm vs. Docker):
* Docker returns "ApiVersion" (https://docs.docker.com/reference/api/docker_remote_api_v1.18/, GET /version)
* Swarm returns "APIVersion" (https://github.com/docker/swarm/blob/master/api/handlers.go, lines 23+47+54)

According to https://docs.docker.com/swarm/#swarm-api, Swarm should be compatible with Docker, except for the points mentioned in https://docs.docker.com/swarm/API/.

We worked-around the issue by locally patching the project, but that's effectively now a fork. If possible, it would be great if in the above Swarm handlers.go, "APIVersion" could simply be replaced with "ApiVersion". That would solve the issue and make case-sensitive property access work.

Other than that, swapping Docker with Swarm worked like a charme. Thank you for these great tools. They are awesome!

Cheers, Vedran

vieux:
Close #687

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -44,7 +44,7 @@ func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
 func getVersion(c *context, w http.ResponseWriter, r *http.Request) {
 	version := struct {
 		Version    string
-		APIVersion string
+		APIVersion string `json:"ApiVersion"`
 		GoVersion  string
 		GitCommit  string
 		Os         string


aluzzardi:
LGTM

Merge after test pass

abronan:
LGTM

PR #788, opened 2015-05-14
DO NOT MERGE - integration  Add rmi without tag integration test

jimmyxian:
Signed-off-by: Xian Chaobo <xianchaobo@huawei.com>

PR #791, opened 2015-05-14
remove images by name or ID, not always ID

vieux:
Replace #788

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -152,8 +152,8 @@ func (e *Engine) updateSpecs() error {
 }
 
 // RemoveImage deletes an image from the engine.
-func (e *Engine) RemoveImage(image *Image) ([]*dockerclient.ImageDelete, error) {
-	return e.client.RemoveImage(image.Id)
+func (e *Engine) RemoveImage(image *Image, name string) ([]*dockerclient.ImageDelete, error) {
+	return e.client.RemoveImage(name)
 }
 
 // RefreshImages refreshes the list of images on the engine.
--- a/cluster/image.go
+++ b/cluster/image.go
@@ -21,8 +21,10 @@ func (image *Image) Match(IDOrName string) bool {
 		return true
 	}
 	for _, repoTag := range image.RepoTags {
-		parts := strings.SplitN(repoTag, ":", 2)
-		if repoTag == IDOrName || parts[0] == IDOrName {
+		if len(strings.SplitN(repoTag, ":", 2)) == 1 {
+			repoTag = repoTag + ":latest"
+		}
+		if repoTag == IDOrName {
 			return true
 		}
 	}
--- a/cluster/image_test.go
+++ b/cluster/image_test.go
@@ -18,7 +18,7 @@ func TestMatch(t *testing.T) {
 	assert.False(t, img.Match("37"))
 
 	assert.True(t, img.Match("name:latest"))
-	assert.True(t, img.Match("name"))
+	assert.False(t, img.Match("name"))
 	assert.False(t, img.Match("nam"))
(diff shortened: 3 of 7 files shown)


cluster/swarm/cluster.go
```diff
@@ -242,11 +242,32 @@ func (c *Cluster) Image(IDOrName string) *cluster.Image {
...
+	for _, n := range c.engines {
+		for _, image := range n.Images() {
+			if image.Match(name) {
+				content, err := image.Engine.RemoveImage(image, name)
+				if err != nil {
+					errs = append(errs, fmt.Sprintf("%s: %s", image.Engine.Name, err.Error()))
+					continue
+				}
+				out = append(out, content...)
+			}
```
jimmyxian: This case:
Node1: testimage:ok
Node2:testimage:latest
`docker_swarm rmi testimage`
 I think the ouput will contain "No such image: testimage:latest" respond by Node1.

cluster/engine.go
```diff
@@ -152,8 +152,8 @@ func (e *Engine) updateSpecs() error {
 }
 
 // RemoveImage deletes an image from the engine.
-func (e *Engine) RemoveImage(image *Image) ([]*dockerclient.ImageDelete, error) {
-	return e.client.RemoveImage(image.Id)
+func (e *Engine) RemoveImage(image *Image, name string) ([]*dockerclient.ImageDelete, error) {
```
aluzzardi: Maybe this should live in `Image`?

`Image.Remove()`

cluster/image_test.go
```diff
@@ -18,7 +18,7 @@ func TestMatch(t *testing.T) {
 	assert.False(t, img.Match("37"))
 
 	assert.True(t, img.Match("name:latest"))
-	assert.True(t, img.Match("name"))
+	assert.False(t, img.Match("name"))
```
aluzzardi: Why is this not matching?

aluzzardi:
This should maybe get the same treatment as #803?

jimmyxian:
@vieux This two interface have different behavior.
`docker rmi`: If remove one image with no tag, it will remove latest.
`docker save`: If save one image with no tag, it will save all the images with repo name.

This two interface in swarm both use image.Match(). I think it's not proper.
In `rmi`, match image with default latest tag if rmi with no tag.
In `docker save`, If save with no tag, match image without tag, just match repo name.

So, two functions maybe OK. `Match` and `MatchWithoutTag()`

cluster/image.go
```diff
@@ -21,8 +21,10 @@ func (image *Image) Match(IDOrName string) bool {
 		return true
 	}
 	for _, repoTag := range image.RepoTags {
-		parts := strings.SplitN(repoTag, ":", 2)
-		if repoTag == IDOrName || parts[0] == IDOrName {
+		if len(strings.SplitN(repoTag, ":", 2)) == 1 {
```
jimmyxian: `repoTag` maybe always have ':'

jimmyxian:
@vieux This change may influence `docker save`
The following two interfaces have different behavior.
`docker rmi`: If remove one image with no tag, it will remove latest.
`docker save`: If save one image with no tag, it will save all the images with repo name.

This two interface in swarm both use image.Match(). I think it's not proper.
In `rmi`, match image with default latest tag if rmi with no tag.
In `docker save`, If save with no tag, match image without tag, just match repo name.

So, two functions maybe OK. `Match` and `MatchWithoutTag()`

aluzzardi:
ping @vieux :)

Again, not familiar with image matching - I'll just review the code itself but I'll let you two figure out if the logic is good

vieux:
replaced by #815

PR #793, opened 2015-05-15
store: Lock returns a channel that is closed if the lock is lost

aluzzardi:


abronan:
LGTM

PR #795, opened 2015-05-15
Store cleanup

aluzzardi:
Some refactoring of the store.

The diff looks overwhelming but if you look at individual commits its actually a bunch of trivial changes.


pkg/store/etcd.go
```diff
@@ -176,48 +175,39 @@ func (s *Etcd) CancelWatch(key string) error {
 
 // AtomicPut put a value at "key" if the key has not been
 // modified in the meantime, throws an error if this is the case
-func (s *Etcd) AtomicPut(key string, oldValue []byte, newValue []byte, index uint64) (bool, error) {
-	resp, err := s.client.CompareAndSwap(normalize(key), string(newValue), 5, string(oldValue), 0)
+func (s *Etcd) AtomicPut(key string, value []byte, previous *KVPair) (bool, error) {
```
aluzzardi: @abronan Need your review of this. I think atomic operations were broken in etcd

aluzzardi:
CI is currently broken but:
```
DOCKER_IMAGE=dockerswarm/dind DOCKER_VERSION=1.6.0 ./test/integration/run.sh
✓ consul discovery
 ✓ etcd discovery
 ✓ file discovery
 ✓ token discovery
 ✓ zookeeper discovery
```

pkg/store/etcd.go
abronan: @aluzzardi I think that you are not required to give each conditions but just one that it is going to be checked for the Compare and Swap. In theory the `lastIndex` is enough in our case so your call is OK (and unsure if mine was but at least giving only the `oldValue` should also be ok in `etcd`). Still we need more fine grained checks on the error:
- Is it an `etcd` related error?
- Is it an `error` related to the valued that cannot being set because the check failed? (In this case returning `store.ErrKeyModified` can help the client code take a decision like retrying the operation)

So I would just find a way to add the `store.ErrKeyModified` error back to the call because this information is useful imho.

abronan:
LGTM

Just a comment on the `AtomicPut`, but as we are not using it yet it's not a blocking point.

/cc @vieux

vieux:
ljy2w

vieux:
LGTM

PR #800, opened 2015-05-15
store: Use channels for Watch/WatchTree

aluzzardi:


aluzzardi:
ping @abronan @vieux @mavenugo

mavenugo:
@aluzzardi did you read my mind :-), I was about to request for this. Not just for the Store, but also for the `discovery.Watch`. Now that libnetwork is making use of discovery package, there will be cases where we would like to stop the discovery. Hence, I was thinking of checking with you for an ability to pass a `stop` channel to the Watch() API, which gives the control to the user to stop the discovery when it prefers. Do you have any opinion on this ?

aluzzardi:
@mavenugo: Yes, this PR is the foundations for that :)

I will change discovery so that:
1/ `Watch` will take a stop channel and return a channel which can be stopped at any time
2/ `discovery.Fetch` will disappear. The only call will be `Watch` which will immediately fire the current entries to the channel (so no need for `List`+`Watch`, just `Watch).
3/ `Watch` will be resilient to errors. Right now, if the store goes down, Watch aborts. It won't be the case anymore and it will keep retrying forever and ever, until `stopCh` gets called.

vieux:
LGTM

abronan:
We should probably wait for integration tests in this case? Just to be sure that any future change on the `Watch` won't break the whole discovery part?

aluzzardi:
We should but I'm actually working on resilient `Discovery` which will alter a lot integration tests so I'll do it in a single pass

aluzzardi:
Merging this since I have your LGTMs.

More tests are coming in another PR.

Issue #794, opened 2015-05-15
Broken TTLs


aluzzardi:
It turns out that:

- zk/etcd/consul were not using any TTLs (entries never expired)
- token discovery "never" expires (one week)

Once we fix **1** (@abronan) and **2** (@vieux) we should probably alter the `discovery-*.bats` tests to, at the end, stop one join, wait `heartbeat` seconds and make sure `swarm list` (and possibly `docker info`) doesn't show up the node.

vieux:
At least etcd used to handle TTLs : https://github.com/docker/swarm/blob/48fd99368feacb2929563ea7ec74931d8101b166/discovery/etcd/etcd.go#L87

abronan:
Actually it is fairly easy to implement the concept of Ephemeral for Consul and Zookeeper. For Zookeeper I create a `Key` with the `Ephemeral` flag while in Consul I create a global session with the attribute `Behavior` to `delete` (which will trigger the node deletion at Session invalidate, when the node is considered down by the failure detector)

The open question is: how to do a similar thing with `etcd`? We can emulate this behavior using TTLs but not sure it's the best idea.

So we need to decide whether we:
- Sacrifice TTLs for Ephemeral behavior (supported by Consul and Zookeeper)
- Use TTLs in place of Ephemeral (supported by Etcd, and Consul through sessions), thus ignoring the fact that Zookeeper does not handle TTLs at all.

PR #801, opened 2015-05-15
store: Add node Ephemeral support

abronan:
Adds an `Ephemeral` flag to Consul, Zookeeper and Etcd for usage with Discovery.

This PR is mostly made to discuss about what is the best approach: rely on TTLs or use an Ephemeral flag. In both cases there are trade-offs to be made on the semantics of both operations:

- If we use an Ephemeral flag, we hide etcd's lack of support for Ephemeral by the use of a periodic `Set` that renews the TTL (What this PR does)
- If we use a TTL, we must hide zookeeper's lack of support for TTL behind the use of an Ephemeral flag.

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -49,13 +49,13 @@ var (
 		EnvVar: "SWARM_HOST",
 	}
 	flHeartBeat = cli.StringFlag{
-		Name:  "heartbeat, hb",
-		Value: "25s",
+		Name:  "heartbeat",
+		Value: "20s",
 		Usage: "period between each heartbeat",
 	}
 	flTTL = cli.StringFlag{
-		Name:  "time-to-live, ttl",
-		Value: "75s",
+		Name:  "ttl",
+		Value: "60s",
 		Usage: "sets the expiration of an ephemeral node",
 	}
 	flTimeout = cli.StringFlag{
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -53,6 +53,11 @@ var (
 		Value: "25s",
 		Usage: "period between each heartbeat",
 	}
+	flTTL = cli.StringFlag{
+		Name:  "time-to-live, ttl",
+		Value: "75s",
+		Usage: "sets the expiration of an ephemeral node",
+	}
 	flTimeout = cli.StringFlag{
 		Name:  "timeout",
 		Value: "10s",
--- a/cli/join.go
+++ b/cli/join.go
@@ -27,7 +27,14 @@ func join(c *cli.Context) {
 	if hb < 1*time.Second {
 		log.Fatal("--heartbeat should be at least one second")
 	}
-	d, err := discovery.New(dflag, hb)
+	ttl, err := time.ParseDuration(c.String("ttl"))
+	if err != nil {
+		log.Fatalf("invalid --ttl: %v", err)
+	}
+	if ttl <= hb {
+		log.Fatal("--ttl must be strictly superior to the heartbeat value")
+	}
+	d, err := discovery.New(dflag, hb, ttl)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -56,7 +56,6 @@ func NewCluster(scheduler *scheduler.Scheduler, store *state.Store, TLSConfig *t
 	}
 
 	heartbeat := 25 * time.Second
-	ttl := 75 * time.Second
 	if opt, ok := options.String("swarm.discovery.heartbeat", ""); ok {
 		h, err := time.ParseDuration(opt)
 		if err != nil {
@@ -68,19 +67,8 @@ func NewCluster(scheduler *scheduler.Scheduler, store *state.Store, TLSConfig *t
 		heartbeat = h
 	}
 
-	if opt, ok := options.String("swarm.discovery.ttl", ""); ok {
-		t, err := time.ParseDuration(opt)
-		if err != nil {
-			return nil, err
-		}
-		if t <= heartbeat {
-			return nil, fmt.Errorf("invalid ttl %s: must be strictly superior to heartbeat value", opt)
-		}
-		ttl = t
-	}
-
 	// Set up discovery.
-	cluster.discovery, err = discovery.New(dflag, heartbeat, ttl)
+	cluster.discovery, err = discovery.New(dflag, heartbeat, 0)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -56,6 +56,7 @@ func NewCluster(scheduler *scheduler.Scheduler, store *state.Store, TLSConfig *t
 	}
 
 	heartbeat := 25 * time.Second
+	ttl := 75 * time.Second
 	if opt, ok := options.String("swarm.discovery.heartbeat", ""); ok {
 		h, err := time.ParseDuration(opt)
 		if err != nil {
@@ -67,8 +68,19 @@ func NewCluster(scheduler *scheduler.Scheduler, store *state.Store, TLSConfig *t
 		heartbeat = h
 	}
 
+	if opt, ok := options.String("swarm.discovery.ttl", ""); ok {
+		t, err := time.ParseDuration(opt)
+		if err != nil {
+			return nil, err
+		}
+		if t <= heartbeat {
+			return nil, fmt.Errorf("invalid ttl %s: must be strictly superior to heartbeat value", opt)
+		}
+		ttl = t
+	}
+
 	// Set up discovery.
-	cluster.discovery, err = discovery.New(dflag, heartbeat)
+	cluster.discovery, err = discovery.New(dflag, heartbeat, ttl)
 	if err != nil {
 		log.Fatal(err)
 	}
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -84,7 +84,7 @@ func (e Entries) Diff(cmp Entries) (Entries, Entries) {
 // The Discovery interface is implemented by Discovery backends which
 // manage swarm host entries.
 type Discovery interface {
-	// Initialize the discovery with URIs and a heartbeat.
+	// Initialize the discovery with URIs, a heartbeat and a ttl.
 	Initialize(string, time.Duration, time.Duration) error
 
 	// Watch the discovery for entry changes.
@@ -131,7 +131,7 @@ func parse(rawurl string) (string, string) {
 	return parts[0], parts[1]
 }
 
-// New returns a new Discovery given a URL and heartbeat settings.
+// New returns a new Discovery given a URL, heartbeat and ttl settings.
 // Returns an error if the URL scheme is not supported.
 func New(rawurl string, heartbeat time.Duration, ttl time.Duration) (Discovery, error) {
 	scheme, uri := parse(rawurl)
--- a/discovery/discovery.go
+++ b/discovery/discovery.go
@@ -85,7 +85,7 @@ func (e Entries) Diff(cmp Entries) (Entries, Entries) {
 // manage swarm host entries.
 type Discovery interface {
 	// Initialize the discovery with URIs and a heartbeat.
-	Initialize(string, time.Duration) error
+	Initialize(string, time.Duration, time.Duration) error
 
 	// Watch the discovery for entry changes.
 	// Returns a channel that will receive changes or an error.
@@ -133,12 +133,12 @@ func parse(rawurl string) (string, string) {
 
 // New returns a new Discovery given a URL and heartbeat settings.
 // Returns an error if the URL scheme is not supported.
-func New(rawurl string, heartbeat time.Duration) (Discovery, error) {
+func New(rawurl string, heartbeat time.Duration, ttl time.Duration) (Discovery, error) {
 	scheme, uri := parse(rawurl)
 
 	if discovery, exists := discoveries[scheme]; exists {
 		log.WithFields(log.Fields{"name": scheme, "uri": uri}).Debug("Initializing discovery service")
-		err := discovery.Initialize(uri, heartbeat)
+		err := discovery.Initialize(uri, heartbeat, ttl)
 		return discovery, err
 	}
 
--- a/pkg/store/consul.go
+++ b/pkg/store/consul.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"net/http"
 	"strings"
+	"sync"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -19,11 +20,11 @@ const (
 
 // Consul embeds the client and watches
 type Consul struct {
+	sync.Mutex
 	config           *api.Config
 	client           *api.Client
 	ephemeralTTL     time.Duration
 	ephemeralSession string
-	sessions         map[string]string
 }
 
 type consulLock struct {
@@ -34,7 +35,6 @@ type consulLock struct {
 // a list of endpoints and optional tls config
 func InitializeConsul(endpoints []string, options *Config) (Store, error) {
 	s := &Consul{}
-	s.sessions = make(map[string]string)
 
 	// Create Consul client
 	config := api.DefaultConfig()
@@ -85,6 +85,26 @@ func (s *Consul) setEphemeralTTL(time time.Duration) {
 	s.ephemeralTTL = time
 }
 
+// CreateEphemeralSession creates the a global session
+// once that is used to delete keys at node failure
+func (s *Consul) createEphemeralSession() error {
+	s.Lock()
+	defer s.Unlock()
+	if s.ephemeralSession == "" {
+		entry := &api.SessionEntry{
+			Behavior: api.SessionBehaviorDelete,
+			TTL:      s.ephemeralTTL.String(),
+		}
+		// Create global ephemeral keys session
+		session, _, err := s.client.Session().Create(entry, nil)
+		if err != nil {
+			return err
+		}
+		s.ephemeralSession = session
+	}
+	return nil
+}
+
 // Normalize the key for usage in Consul
 func (s *Consul) normalize(key string) string {
 	key = normalize(key)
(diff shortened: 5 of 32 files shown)


abronan:
partially fixes #794

cli/join.go
```diff
@@ -42,11 +42,10 @@ func join(c *cli.Context) {
 	}
 
 	hbval := time.Duration(hb)
+	log.WithFields(log.Fields{"addr": addr, "discovery": dflag}).Infof("The entry in the discovery will disappear if the node goes down..", hbval)
+	if err := d.Register(addr); err != nil {
+		log.Error(err)
+	}
 	for {
```
aluzzardi: `select{}`

pkg/store/store.go
```diff
@@ -97,6 +97,11 @@ type KVPair struct {
 	LastIndex uint64
 }
 
+// RequestOptions contains optional request parameters
+type RequestOptions struct {
```
aluzzardi: `WriteOptions`?

aluzzardi: `AtomicPut` should take a `WriteOptions` as well, to be consistent.

aluzzardi: Also, ephemeral should probably be implemented in AtomicPut as well I guess

cli/join.go
```diff
@@ -42,11 +42,10 @@ func join(c *cli.Context) {
 	}
 
 	hbval := time.Duration(hb)
+	log.WithFields(log.Fields{"addr": addr, "discovery": dflag}).Infof("The entry in the discovery will disappear if the node goes down..", hbval)
```
aluzzardi: Probably not worth mentioning

pkg/store/etcd.go
```diff
@@ -99,13 +86,21 @@ func (s *Etcd) Get(key string) (*KVPair, error) {
 }
 
 // Put a value at "key"
-func (s *Etcd) Put(key string, value []byte) error {
-	if _, err := s.client.Set(key, string(value), 0); err != nil {
+func (s *Etcd) Put(key string, value []byte, opts *RequestOptions) error {
+
+	// Default TTL = 0 means no expiration
+	ttl := 0 * time.Second
```
aluzzardi: `ttl := 0` should be enough?

pkg/store/etcd.go
```diff
@@ -99,13 +86,21 @@ func (s *Etcd) Get(key string) (*KVPair, error) {
...
+	if opts != nil && opts.Ephemeral {
+		ttl = 60 * time.Second
```
aluzzardi: Make `60` either:

- a `const`
- a `WriteOption`
- a `Config`

pkg/store/etcd.go
```diff
@@ -115,6 +110,30 @@ func (s *Etcd) Put(key string, value []byte) error {
...
+		} else {
+			return err
+		}
+	}
+	return nil
+}
+
+// Periodically renew a key with an Ephemeral flag set
+func (s *Etcd) periodicRenew(key string, value []byte, ttl time.Duration) {
+	renew := time.NewTicker(40 * time.Second)
```
aluzzardi: Same here:

Make `40` either:
- a `const`
- a `WriteOption`
- a `Config`

This could be named `heartbeat` or something like that (this is exactly what the heartbeat was supposed to be like)

cli/join.go
abronan: Ok

pkg/store/etcd.go
```diff
@@ -115,6 +110,30 @@ func (s *Etcd) Put(key string, value []byte) error {
...
+	for _ = range renew.C {
+		if _, err := s.client.Set(key, string(value), uint64(ttl)); err != nil {
```
aluzzardi: This will override the key, meaning that:

```
c.Put("foo", "bar", &WriteOptions{Ephemeral: true})
c.Put("foo", "hello world")
c.Get("foo") <- after 40 seconds this is "bar" again
```

You can solve this by using an `CompareAndSwap` and if it fails, at that point do a get. 
Probably no need to get the `value` as a parameter: just go a get first, grab the lastIndex and value, then compare and swap with that value and last index. When it fails, get the new value and lastIndex and loop on

pkg/store/etcd.go
aluzzardi: Instead of `NewTicket`, you can use:

```
renewCh := time.Tick(...)
for _ = range renew {
}
```

pkg/store/etcd.go
abronan: I think the good practice would be to use the `Update` method of the go-client, so I need to dig this.

abronan: Well cannot seem to avoid the CAS call here: there is no way with the client to only update the key's TTL without specifying a value, this is only do-able for a directory. So.. going back to the CAS solution..

pkg/store/etcd.go
abronan: If I put this into `WriteOption` that would defeat the purpose of hiding it behind the `Ephemeral` flag. Same thing if I put this in Config. Best bet for now would be to put those as `const` (unless we can specify `heartbeat` + `refresh` time in the `cli`).

pkg/store/etcd.go
```diff
@@ -99,13 +86,21 @@ func (s *Etcd) Get(key string) (*KVPair, error) {
...
 // Put a value at "key"
-func (s *Etcd) Put(key string, value []byte) error {
-	if _, err := s.client.Set(key, string(value), 0); err != nil {
+func (s *Etcd) Put(key string, value []byte, opts *RequestOptions) error {
+
+	// Default TTL = 0 means no expiration
+	ttl := 0 * time.Second
+	if opts != nil && opts.Ephemeral {
+		ttl = 60 * time.Second
+		go s.periodicRenew(key, value, ttl)
```
aluzzardi: Unfortunately, I think we have to keep track of our goroutines :/

This means that every time we do an ephemeral put of the same key, we end up with a new goroutine.

We could use a kind of "set" (`map[string]struct{}`) and conditionally spawn goroutines

cli/join.go
aluzzardi: Actually you were right. I would keep this the way it was and continue issuing periodic `Register()` in case something goes wrong (instead of doing just one Register)

abronan:
Rebased and updated Godeps (there was an issue with `go-etcd` and the `ttl` parameter of `Set`, setting it to a value different than 0 would cause an index out of bound exception).

It now works consistently across the 3 stores (at least locally, Jenkins fails on `etcd` for some reasons). I kept it simple, for now the TTL is set as a const: `60s`, and it is hidden behind the `ephemeral` behavior. I'm waiting for #805 to update this and place the `EphemeralTTL` to be higher than the heartbeat (something like 2 or 3 times the heartbeat set on `join`).

abronan:
Should we update the `etcd` image on integration tests to use at least `v2.0.0` (and require everyone to use etcd >= `v2.0.0`)? I think that updating Godeps with the latest version of `go-etcd` is probably the reason integration tests are failing now.. (we still use `v0.4.6`)

I did all my tests on the latest version of etcd: `v2.0.11`

/cc @aluzzardi

aluzzardi:
@abronan I actually tried that - for some reason, updating etcd or go-etcd breaks our tests :(

I didn't try to updating both though.

We could try once we get #805 in since that PR is changing the tests a lot (they test way more), as it would be unsafe to upgrade without getting those tests in first.

abronan:
@aluzzardi Yes that was what I was thinking. I'll try updating both once we have your PR merged and see how we can fix this.

aluzzardi:
We should test that the TTL is working correctly with a test similar to #811

aluzzardi:
>  I kept it simple, for now the TTL is set as a const: 60s, and it is hidden behind the ephemeral behavior. I'm waiting for #805 to update this and place the EphemeralTTL to be higher than the heartbeat (something like 2 or 3 times the heartbeat set on join).

In order to be able to do integration tests, we'll need to be able to configure the TTL.

How about adding `EphemeralTTL` and `EphemeralHeartbeat` to the store `Config`?

We'd use the cli `heartbeat` flag for the heartbeat, and we could probably add a `swarm.discovery.ttl` flag for the TTL - it makes sense for other stores as well (`token` /cc @vieux)

abronan:
@aluzzardi Rebased once again. Tests are now passing. I changed the etcd image for the latest version `v2.0.11`. Still a few changes to make but we are slowly getting there :ghost: 

Also I think I'd probably like to add the integration tests at this point.

abronan:
ping @aluzzardi 

Rebased on master (with Node removal). It works fine, I added a single integration test for each backend.

It is not ideal at the moment but we need to figure out what to do with etcd first if we want a refresh happening in the background (because of the lack of a `TTL Update` function on key without specifying a value).

Also Consul does not accepts TTLs inferior to `10s`, thus the total integration tests time is increasing a little bit :-(

discovery/discovery.go
```diff
@@ -85,7 +85,7 @@ func (e Entries) Diff(cmp Entries) (Entries, Entries) {
 // manage swarm host entries.
 type Discovery interface {
 	// Initialize the discovery with URIs and a heartbeat.
-	Initialize(string, time.Duration) error
+	Initialize(string, time.Duration, time.Duration) error
```
aluzzardi: Please update the comment above

abronan: +1

cli/flags.go
```diff
@@ -53,6 +53,11 @@ var (
 		Value: "25s",
 		Usage: "period between each heartbeat",
 	}
+	flTTL = cli.StringFlag{
+		Name:  "time-to-live, ttl",
```
aluzzardi: Should we call this just `ttl`?

Also, what about removing `hb` - just `heartbeat`?

I think shortcut helpers don't make much sense for advanced options

aluzzardi: Also, how about rounding out?

What do you think of those defaults:
- `heartbeat` = 20s
- `ttl` = 60s

cluster/swarm/cluster.go
```diff
@@ -68,7 +68,7 @@ func NewCluster(scheduler *scheduler.Scheduler, store *state.Store, TLSConfig *t
 	}
 
 	// Set up discovery.
-	cluster.discovery, err = discovery.New(dflag, heartbeat)
+	cluster.discovery, err = discovery.New(dflag, heartbeat, 0)
```
aluzzardi: Shouldn't we pass a `ttl` here?

aluzzardi: Oh, I see - it's useless.

cli/flags.go
abronan: +1

cluster/swarm/cluster.go
aluzzardi: How about building the discovery object in manage directly? Dependency injection would be better.

Also, `swarm.discovery.heartbeat`: we could simply reuse the flHeartbeat we already have in manage

discovery/discovery.go
```diff
@@ -133,12 +133,12 @@ func parse(rawurl string) (string, string) {
 
 // New returns a new Discovery given a URL and heartbeat settings.
 // Returns an error if the URL scheme is not supported.
-func New(rawurl string, heartbeat time.Duration) (Discovery, error) {
+func New(rawurl string, heartbeat time.Duration, ttl time.Duration) (Discovery, error) {
```
aluzzardi: Comment above

pkg/store/consul.go
```diff
@@ -91,11 +105,55 @@ func (s *Consul) Get(key string) (*KVPair, error) {
...
+			// EphemeralSession
+			lockOpts := &api.LockOptions{
+				Key:     key,
+				Session: session,
+			}
+
+			// Lock and ignore if lock is held
+			// It's just a placeholder for the
+			// ephemeral behavior
+			lock, _ := s.client.LockOpts(lockOpts)
```
aluzzardi: I don't understand this?

pkg/store/zookeeper.go
```diff
@@ -22,10 +22,13 @@ type zookeeperLock struct {
 // given a list of endpoints and optional tls config
 func InitializeZookeeper(endpoints []string, options *Config) (Store, error) {
 	s := &Zookeeper{}
-	s.timeout = 5 * time.Second // default timeout
+	s.timeout = 10 * time.Second // default timeout
```
aluzzardi: please put a const in `zookeeper.go` (private, doesn't need to be exported)

pkg/store/consul.go
```diff
@@ -91,11 +105,55 @@ func (s *Consul) Get(key string) (*KVPair, error) {
...
+func (s *Consul) Put(key string, value []byte, opts *WriteOptions) error {
+
+	key = s.normalize(key)
+
+	p := &api.KVPair{
+		Key:   key,
+		Value: value,
+	}
+
+	if opts != nil && opts.Ephemeral {
```
aluzzardi: Since the session shares the same `ttl` as everyone else, I think you can create just one single session for all ephemeral keys.

Also, I would put this code in something like `createEphemeralSession` so you can remove it from the put.

Maybe you can call that function if `opts.Epheneral` and let it check if it needs or not to create a session:
```
if opts != nil && opts.Ephemeral {
   s.createEphemeralSession()
   [...]
```

createEphemeralSession could use a mutex to avoid creating multiple sessions for nothing.

pkg/store/consul.go
```diff
@@ -91,11 +105,55 @@ func (s *Consul) Get(key string) (*KVPair, error) {
...
+		if _, ok := s.sessions[key]; !ok {
+			entry := &api.SessionEntry{
+				Behavior: api.SessionBehaviorDelete,
+				TTL:      s.ephemeralTTL.String(),
```
aluzzardi: maybe: `"kv_ephemeral_" + s.ephemeralTTL.String()`? Just to avoid clashing with user sessions

test/integration/discovery/discovery_helpers.bash
```diff
@@ -10,6 +10,11 @@ function discovery_check_swarm_info() {
 	docker_swarm info | grep -q "Nodes: $count"
 }
 
+# Returns true if swarm info outputs is empty (0 nodes).
```
aluzzardi: Actually, I changed the above (`discovery_check_swarm_info`) to take a parameter so you could do this with:

`discovery_check_swarm_info "$DISCOVERY" 0`

test/integration/discovery/consul.bats
```diff
@@ -59,6 +59,31 @@ function teardown() {
 	retry 5 1 discovery_check_swarm_info
 }
 
+@test "consul discovery: check for engines departure" {
```
aluzzardi: Could the name be consistent with `file`?
https://github.com/abronan/swarm/blob/add_ephemeral_support/test/integration/discovery/file.bats#L56

Either call this "node removal" or change file to "check for engines departure"

test/integration/discovery/etcd.bats
```diff
@@ -60,6 +64,31 @@ function teardown() {
 	retry 5 1 discovery_check_swarm_info
 }
 
+@test "etcd discovery: check for engines departure" {
```
aluzzardi: Same comments than consul

test/integration/discovery/zk.bats
```diff
@@ -59,6 +59,31 @@ function teardown() {
 	retry 10 1 discovery_check_swarm_info
 }
 
+@test "zk discovery: check for engines departure" {
```
aluzzardi: Same

test/integration/discovery/consul.bats
```diff
@@ -59,6 +59,31 @@ function teardown() {
...
+	retry 10 1 discovery_check_swarm_info
+
+	# Add Engines to the cluster and make sure it's picked by swarm
+	start_docker 2
+	swarm_join "$DISCOVERY"
+	retry 5 1 discovery_check_swarm_list "$DISCOVERY"
+	retry 5 1 discovery_check_swarm_info
+
+	# Removes all the swarm agents
+	swarm_join_cleanup
```
aluzzardi: So, with `file` I removed only one engine.

It's up to you what you prefer to do (remove one or all).

If you wish to remove only one, you could do `kill ${SWARM_JOIN_PIDS[0]}`

aluzzardi:
Few minor comments otherwise looks good

test/integration/discovery/discovery_helpers.bash
abronan: Oh yeah I see that's new :)

pkg/store/consul.go
abronan: We need a lock as the holder of the session with `Behavior: Delete` to mimic Zookeeper's ephemeral node behavior :)

pkg/store/consul.go
abronan: Session ID is a generated uuid and the name to give to the session is optional. So we can add a name but I'm not sure it will serve any purpose.

abronan:
@aluzzardi PTAL

/cc @vieux

pkg/store/consul.go
```diff
@@ -91,11 +125,47 @@ func (s *Consul) Get(key string) (*KVPair, error) {
...
+	if opts != nil && opts.Ephemeral {
+		// Creates the global ephemeral session
+		// if it does not exist
+		if s.ephemeralSession == "" {
+			s.createEphemeralSession()
+		}
+
+		// Create lock option with the
+		// EphemeralSession
+		lockOpts := &api.LockOptions{
```
aluzzardi: Should we do this in `createEphemeralSession`?

pkg/store/consul.go
```diff
@@ -91,11 +125,47 @@ func (s *Consul) Get(key string) (*KVPair, error) {
...
+
+	p := &api.KVPair{
+		Key:   key,
+		Value: value,
 	}
+
+	if opts != nil && opts.Ephemeral {
+		// Creates the global ephemeral session
+		// if it does not exist
+		if s.ephemeralSession == "" {
```
aluzzardi: We should probably not do the check here since it's already done by `createEphemeralSession`

pkg/store/consul.go
abronan: After tinkering a little bit with Consul and the lock mechanism, if we give the same session to the `api.KVPair` and `api.LockOptions`, the call to `lock.Lock()` is non blocking because of those lines in `consul/lock.go`:

```
if pair != nil && pair.Session == l.lockSession {
		goto HELD
}
```
I don't think we can put it in `createEphemeralSession` either, we need to go through the lock call each time just in case the session expired and we want to come back to the discovery service if we recover (but I need to add one line during an error of refresh if the session does not exists anymore and put it back to `""`).

Also, I thought that this might leak goroutines by just calling `lock.Lock()` each time because of this line in `consul/lock.go`:

```
go l.monitorLock(l.lockSession, leaderCh)
```
But it appears that no :D

It's just a placeholder lock that does nothing else but hold the ephemeral session :)

abronan: Errr didn't answered the real question :D I think it's good here because we associate the key in `LockOptions` with the same global session. `LockOptions` should be different for each key though (to be associated with this global session).

aluzzardi:
LGTM

PR #803, opened 2015-05-15
fix container matching algorithm

### LINKED ISSUE #463: Cluster-wide container name conflict
When using swarm, the `NAMES` column of `docker ps` shows names using `hostname/container` format. But the conflict detection mechanism only check the `container` name...

```bash
$ docker run -d --name nginx.remote.container -e constraint:type==app nginx
79af90719474340437933d5129a59e4e592fb2bd908a665e7c854d9c87f87567

$ docker run -d --name nginx.remote.container -e constraint:type==db nginx
FATA[0000] Error response from daemon: Conflict, The name nginx.remote.container is already assigned to 79af90719474340437933d5129a59e4e592fb2bd908a665e7c854d9c87f87567. You have to delete (or rename) that container to be able to assign host.swarm to a container again.
```

Name conflicts shouldn't be checked cluster-wide, but local to each remote nodes using the full `hostname/container` format.

---

```bash
$ docker version
Client version: 1.4.1
Client API version: 1.16
Go version (client): go1.3.3
Git commit (client): 5bc2ff8
OS/Arch (client): linux/amd64
Server version: swarm/0.1.0
Server API version: 1.16
Go version (server): go1.4.2
Git commit (server): HEAD

$ swarm -v
swarm version 0.1.0 (HEAD)
```

### LINKED ISSUE #802: Container matching algorithm is wrong
If we name a container `c1` and another one has an ID that starts with `c1`, `cluster.Container()` will return the one whose ID starts with `c1`, which is wrong.

We should revisit our matching policy has follows:
1/ Exact ID
2/ Short ID (maybe? the one 12 chars long ... there is a helper in `docker/pkg/stringid`)
3/ Exact Name
4/ Prefix ID matching

vieux:
fix #802
fix #463

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -371,8 +371,42 @@ func (c *Cluster) Container(IDOrName string) *cluster.Container {
 
 	c.RLock()
 	defer c.RUnlock()
-	for _, n := range c.engines {
-		if container := n.Container(IDOrName); container != nil {
+
+	containers := c.Containers()
+
+	// Match exact or short Container ID.
+	for _, container := range containers {
+		if container.Id == IDOrName || stringid.TruncateID(container.Id) == IDOrName {
+			return container
+		}
+	}
+
+	// Match exact Swarm ID.
+	for _, container := range containers {
+		if swarmID := container.Config.SwarmID(); swarmID == IDOrName || stringid.TruncateID(swarmID) == IDOrName {
+			return container
+		}
+	}
+
+	// Match name, /name or engine/name.
+	for _, container := range containers {
+		for _, name := range container.Names {
+			if name == IDOrName || name == "/"+IDOrName || container.Engine.ID+name == IDOrName || container.Engine.Name+name == IDOrName {
+				return container
+			}
+		}
+	}
+
+	// Match Container ID prefix.
+	for _, container := range containers {
+		if strings.HasPrefix(container.Id, IDOrName) {
+			return container
+		}
+	}
+
+	// Match Swarm ID prefix.
+	for _, container := range containers {
+		if strings.HasPrefix(container.Config.SwarmID(), IDOrName) {
 			return container
 		}
 	}
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -388,6 +388,8 @@ func (c *Cluster) Container(IDOrName string) *cluster.Container {
 		}
 	}
 
+	candidates := []*cluster.Container{}
+
 	// Match name, /name or engine/name.
 	for _, container := range containers {
 		for _, name := range container.Names {
@@ -397,18 +399,28 @@ func (c *Cluster) Container(IDOrName string) *cluster.Container {
 		}
 	}
 
+	if size := len(candidates); size == 1 {
+		return candidates[0]
+	} else if size > 1 {
+		return nil
+	}
+
 	// Match Container ID prefix.
 	for _, container := range containers {
 		if strings.HasPrefix(container.Id, IDOrName) {
-			return container
+			candidates = append(candidates, container)
 		}
 	}
 
 	// Match Swarm ID prefix.
 	for _, container := range containers {
 		if strings.HasPrefix(container.Config.SwarmID(), IDOrName) {
-			return container
+			candidates = append(candidates, container)
 		}
+	}
+
+	if len(candidates) == 1 {
+		return candidates[0]
 	}
 
 	return nil
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -455,35 +455,6 @@ func (e *Engine) Containers() []*Container {
 	return containers
 }
 
-// Container returns the container with IDOrName in the engine.
-func (e *Engine) Container(IDOrName string) *Container {
-	// Abort immediately if the name is empty.
-	if len(IDOrName) == 0 {
-		return nil
-	}
-
-	for _, container := range e.Containers() {
-		// Match Container ID prefix.
-		if strings.HasPrefix(container.Id, IDOrName) {
-			return container
-		}
-
-		// Match Swarm ID prefix.
-		if strings.HasPrefix(container.Config.SwarmID(), IDOrName) {
-			return container
-		}
-
-		// Match name, /name or engine/name.
-		for _, name := range container.Names {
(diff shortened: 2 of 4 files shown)


vieux:
ping @aluzzardi

thaJeztah:
Does this match the behaviour in Docker? I know in Docker this exact issue was fixed a number of times, so would be good if they behaved the same.

vieux:
@thaJeztah it should with the `SwarmID` stuff added

thaJeztah:
@vieux alright then! Just checking, I can try to find the issues in docker, if needed.

vieux:
@thaJeztah the PR would be nice, just to check :D

thaJeztah:
I'll try and find them (picks memory for useful keywords) :)

thaJeztah:
First results; https://github.com/docker/docker/pull/7443, https://github.com/docker/docker/issues/8069, https://github.com/docker/docker/pull/9705.

Slightly related; https://github.com/docker/docker/issues/8098 and https://github.com/docker/docker/pull/8941

And some background https://github.com/docker/docker/pull/8087and https://groups.google.com/forum/m/#!topic/docker-user/jyf9_mYcMI8

aluzzardi:
Docker seems to handle this differently (docker/docker#9705). If I'm not mistaken:

1/ Exact name
2/ Exact ID
3/ Prefix ID

aluzzardi:
It also handles ambiguous IDs which we should probably do as well and test for.

container1: c1234
container2: c1999

`Container("c1")` should return an error

thaJeztah:
Hm, strange, especially in light of https://github.com/docker/docker/pull/8087#issuecomment-55969917.

Perhaps that was changed later on? Preferring full ID match over full name match looks to me like the preferred approach. (IMO, a name should **never** be able to shadow an ID). If that's the current case in Docker, that looks like a bug to me.

@aluzzardi could you check if that's indeed the current behavior in Docker?

/cc @ewindisch

aluzzardi:
@thaJeztah Yeah, docker/docker#9705 replaced the behavior of docker/docker#8087.

It's the current behavior in master:

https://github.com/docker/docker/blob/master/daemon/daemon.go#L128
```go
	// GetByName will match only an exact name provided; we ignore errors
	containerByName, _ := daemon.GetByName(prefixOrName)
	containerId, indexError := daemon.idIndex.Get(prefixOrName)

	if containerByName != nil {
		// prefix is an exact match to a full container Name
		return containerByName, nil
	}
```

thaJeztah:
Looks like that PR change wasn't reviewed by a security maintainer/member.

(Sorry, I'm going to hijack this PR to try and include them in this discussion. Please let me know if you would like that discussion somewhere else)

thaJeztah:
ping @ewindisch @diogomonica your input would be appreciated on this. Perhaps I'm on the wrong foot here, but it looks like the matching in Docker is currently incorrect. Please read the discussion above

diogomonica:
@thaJeztah this does seems concerning. The default behavior should always be matching on full Id, and the name/prefix only be used as a convenience (more specific to less specific). One would expect that Docker does resolution of potential collisions when outputting prefixes (which from @aluzzardi's comments seems like it does).

Can this be supporting something along the lines of aliases? To be honest, I'm not familiar with this area of the code. Who is the best person to ping about this behavior?

aluzzardi:
@thaJeztah @diogomonica I've opened docker/docker#13286 and I'll merge this PR since it looks like we are doing the right thing here in Swarm.

aluzzardi:
Actually: @vieux we need to handle collision checking in this PR (as commented above). Right now we are taking the first match.

vieux:
@aluzzardi can you please take another look ?

cluster/swarm/cluster.go
```diff
@@ -371,12 +371,52 @@ func (c *Cluster) Container(IDOrName string) *cluster.Container {
...
+	}
+
+	// Match Swarm ID prefix.
+	for _, container := range containers {
+		if strings.HasPrefix(container.Config.SwarmID(), IDOrName) {
+			candidates = append(candidates, container)
+		}
+	}
+
+	if len(candidates) == 1 {
```
aluzzardi: Maybe:

if len(candidates) > 1 {
  return nil, fmt.Errorf("whatever docker's message is when there is a conflict")
}

aluzzardi: Maybe:

if len(candidates) > 1 {
  return nil, fmt.Errorf("whatever docker's message is when there is a conflict")
}

cluster/swarm/cluster.go
```diff
@@ -371,12 +371,52 @@ func (c *Cluster) Container(IDOrName string) *cluster.Container {
...
+	}
+
+	// Match exact Swarm ID.
+	for _, container := range containers {
+		if swarmID := container.Config.SwarmID(); swarmID == IDOrName || stringid.TruncateID(swarmID) == IDOrName {
 			return container
 		}
 	}
 
+	// Match name, /name or engine/name.
```
aluzzardi: Maybe do the same here?

Since people can create containers outside of swarm, we may end up with a name collision

aluzzardi:
LGTM

estesp:
For integrity of the history for those who might come behind or end up here before seeing the comments in docker/docker#13286, it is important to note that the code snippet posted above related to name/ID lookup in the Docker daemon ignores the lines preceding it which **does** retain full ID match prior to checking names:
https://github.com/docker/docker/blob/b2a43baf2e2cc68c83383a7524441f81bc4c4725/daemon/daemon.go#L125-L128

```
func (daemon *Daemon) Get(prefixOrName string) (*Container, error) {
	if containerByID := daemon.containers.Get(prefixOrName); containerByID != nil {
		// prefix is an exact match to a full container ID
		return containerByID, nil
	}
```

You can also click the link to the code above and read the algorithm details in the comment preceding the function which clearly shows that the order of matching is:
 1. full ID trumps all other matching
 2. **exact** name matches trump partial ID matches (e.g. *db* supersedes *db019a3..*)
 3. singular match on a short/partial ID

Non-singular matches (e.g. ambiguous hex partial ID prefixes that match multiple objects) return an error from the truncindex package.  This detail may help your noted concern @diogomonica as I know you had less history on this issue from the prior PRs/issues.

PR #811, opened 2015-05-18
Handle Node Removal from Discovery correctly

### LINKED ISSUE #197: Unable to remove a node from the cluster
The ```Watch``` method of the discovery backend calls ```cluster.UpdateNodes``` method to update the  node list according the return value of the ```Fetch``` method of the discovery service.
But it only inserts the nodes that aren't exist in the node list and doesn't touch the nodes that are "removed" from the discovery service.

If a node is removed from the discovery service, the ```swarm list``` command returns the correct result because it calls the ```Fetch``` method of the discovery service directly. But in fact the removed node is still in the node list of the cluster.

aluzzardi:
Fixes #197

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -91,7 +91,7 @@ func createRouter(c *context, enableCors bool) *mux.Router {
 			localRoute := route
 			localFct := fct
 			wrap := func(w http.ResponseWriter, r *http.Request) {
-				log.WithFields(log.Fields{"method": r.Method, "uri": r.RequestURI}).Info("HTTP request received")
+				log.WithFields(log.Fields{"method": r.Method, "uri": r.RequestURI}).Debug("HTTP request received")
 				if enableCors {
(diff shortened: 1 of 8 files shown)


aluzzardi:
ping @vieux @abronan

api/router.go
```diff
@@ -91,7 +91,7 @@ func createRouter(c *context, enableCors bool) *mux.Router {
 			localRoute := route
 			localFct := fct
 			wrap := func(w http.ResponseWriter, r *http.Request) {
-				log.WithFields(log.Fields{"method": r.Method, "uri": r.RequestURI}).Info("HTTP request received")
```
vieux: ok

vieux:
LGTM

vieux:
ping @abronan

abronan:
LGTM

PR #812, opened 2015-05-18
adding environment variable option when treating flags

jimenez:
Signed-off-by: Isabel Jimenez <contact.isabeljimenez@gmail.com>

### DIFF
--- a/cluster/options.go
+++ b/cluster/options.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"os"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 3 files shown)


vieux:
ping @aluzzardi @abronan this is needed from mesos.

abronan:
LGTM

cluster/options.go
```diff
@@ -9,19 +11,22 @@ import (
...
 // String returns a string from the driver options
-func (do DriverOpts) String(key string) (string, bool) {
+func (do DriverOpts) String(key, env string) (string, bool) {
 	for _, opt := range do {
 		kv := strings.SplitN(opt, "=", 2)
 		if kv[0] == key {
 			return kv[1], true
 		}
 	}
+	if env := os.Getenv(env); env != "" {
```
aluzzardi: Why not `os.Getenv("SWARM_CLUSTER_OPT_" + strings.ToUpper(key))` so you don't need to pass an extra parameter?

jimenez: Do we want all the flags to be accessible via the environment? For example in this PR swarm.overcommit could only be set via flag.

aluzzardi: Uhm, you are right. We could replace `.` with `_` so it'd be `SWARM_CLUSTER_OPT_SWARM_OVERCOMMIT` but maybe that's overkill?

aluzzardi:
LGTM

vieux:
LGTM

PR #813, opened 2015-05-19
Add cluster.kill and cluster.stop

vieux:
This will be useful in mesos


vieux:
ping @aluzzardi I know the cluster interface get's bigger, but it's useful

api/handlers.go
```diff
@@ -334,6 +334,30 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 	return
 }
 
+// POST /containers/{name:.*}/stop
```
aluzzardi: why only for stop and not for kill?

aluzzardi:
Is it really necessary? :/

We'll have to maintain API changes and so on and so on

api/handlers.go
vieux: I forgot

vieux:
@aluzzardi I know :/ but mesos handle stop/kill through the master

aluzzardi:
Mmmmkay I guess we don't have a choice :/

I'd rather have those be `Container.Kill()` and `Container.Stop()` but it'd require changing `Container` to an interface and that's probably too big of a change right now.

vieux:
After a good night of sleep, I'm not convinced we need this @tnachen 

Using task kill in mesos will only shell out to the docker cli do to a stop right ?
We already have the direct connection to the engine so we could call stop directly.
I don't think it would make any difference, in any case mesos does a docker wait on the container so you will detect the end of the task.

And the experience is much better from someone who know the docker cli

aluzzardi:
\o/

PR #814, opened 2015-05-19
Fix #810 bool value

### LINKED ISSUE #810: Getting all the containers using the flag true instead of 1
To get all the containers running on a swarm (including the ones stopped) through the Remote API we have to use the flag `1` as we can read in [api/handlers.go](https://github.com/docker/swarm/blob/master/api/handlers.go): 
```go
// GET /containers/ps
// GET /containers/json
func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
	if err := r.ParseForm(); err != nil {
		return
	}

	// Parse flags.
	all := r.Form.Get("all") == "1"
    ...
}
```
`1` is the only flag allowed to get all the containers, not `true` or `True`. As the [Docker Remote API](https://github.com/docker/docker/blob/master/docs/sources/reference/api/docker_remote_api_v1.7.md#list-containers) states that the query parameter `all` accept the flags 1/True/true or 0/False/false when listing containers, there is a minor issue.

jimmyxian:
Fix #810

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -145,8 +145,8 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Parse flags.
-	all := r.Form.Get("all") == "1"
-	limit, _ := strconv.Atoi(r.Form.Get("limit"))
+	all := boolValue(r, "all")
+	limit := intValueOrZero(r, "limit")
 
 	// Parse filters.
 	filters, err := dockerfilters.FromParam(r.Form.Get("filters"))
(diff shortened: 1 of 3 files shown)


vieux:
@jimmyxian can you please add a few tests for `boolValue` and `intValueOrZero` ?

Thanks.

jimmyxian:
@vieux aha, my bad.....forget unit test........I will add some tests....

jimmyxian:
ping @vieux 
Done to add unit test. Check again, thanks

vieux:
LGTM

Issue #740, opened 2015-05-07
Bug: Image.Match() matching image using prefix influence some api


jimmyxian:
### BUG reproduce
one manage: 127.0.0.1:2222
one node with busybox image:127.0.0.1:2375
* docker rmi API
```
[root@localhost swarm]# docker -H 127.0.0.1:2222 rmi bus
Error response from daemon: localhost.localdomain: 409 Conflict: Conflict, cannot delete image 4986bf8c15363d1c5d15512d5266f8777bfba4974ac56e3270e7760f6f0a8125 because it is tagged in multiple repositories, use -f to force
FATA[0000] Error: failed to remove one or more images 

[root@localhost swarm]# docker -H 127.0.0.1:2222 rmi not_exists
Error response from daemon: No such image not_exists
FATA[0000] Error: failed to remove one or more images

[root@localhost swarm]# docker -H 127.0.0.1:2375 rmi bus
Error response from daemon: No such image: bus
FATA[0000] Error: failed to remove one or more images  
```
We find that "bus image" and "not_exists image"  are not in node, but respond differently. Because the busybox'prefix is bus, in our swarm, will send the request the node with busybox.

* docker history API
```
[root@localhost swarm]# docker -H 127.0.0.1:2222 history bus
FATA[0000] Error response from daemon: could not find image: no such id: bus 
[root@localhost swarm]# docker -H 127.0.0.1:2222 history not_exist
FATA[0000] Error response from daemon: No such image: not_exist 
[root@localhost swarm]# docker -H 127.0.0.1:2375 history bus
FATA[0000] Error response from daemon: could not find image: no such id: bus
```
Also respond differently

* docker save API
```
[root@localhost swarm]# docker -H 127.0.0.1:2222 save bus > /home/aaa.tar
time="2015-05-07T06:14:01-04:00" level="fatal" msg="Error: could not find image: no such id: bus" 
[root@localhost swarm]# 
[root@localhost swarm]# docker -H 127.0.0.1:2222 save not_exists > /home/aaa.tar
time="2015-05-07T06:14:11-04:00" level="fatal" msg="Error: No such image: not_exists" 
```
Also respond differently

* Other API which use proxyImage and Image.Match() maybe have the save problem

### Problem
Image.Match() : match image name using prefix

### How to Solve? Need help
Fix matching image name behavior. 
Sometimes name contains tag, sometimes do not have. How to match, does not have a clear way?
In delete image, if name do not have tag, docker will delete latest.
[...]

PR #815, opened 2015-05-19
Fix image match

jimmyxian:
Origin Issue:
https://github.com/docker/swarm/issues/740

Replace:
https://github.com/docker/swarm/pull/791

### DIFF
--- a/api/router.go
+++ b/api/router.go
@@ -29,7 +29,7 @@ var routes = map[string]map[string]handler{
 		"/images/viz":                     notImplementedHandler,
 		"/images/search":                  proxyRandom,
 		"/images/get":                     getImages,
-		"/images/{name:.*}/get":           proxyImage,
+		"/images/{name:.*}/get":           getImage,
 		"/images/{name:.*}/history":       proxyImage,
 		"/images/{name:.*}/json":          proxyImage,
 		"/containers/ps":                  getContainersJSON,
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -477,36 +477,18 @@ func deleteImages(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 	var name = mux.Vars(r)["name"]
 
-	matchedImages := []*cluster.Image{}
-	for _, image := range c.cluster.Images() {
-		if image.Match(name) {
-			matchedImages = append(matchedImages, image)
-		}
-	}
-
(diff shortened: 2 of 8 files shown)


jimmyxian:
ping @vieux 
This PR is what i'm thinking about.
https://github.com/docker/swarm/pull/791#issuecomment-103011312
```
The following two interfaces have different behavior.
docker rmi: If remove one image with no tag, it will remove latest.
docker save: If save one image with no tag, it will save all the images with repo name.

This two interface in swarm both use image.Match(). I think it's not proper.
In rmi, match image with default latest tag if rmi with no tag.
In docker save, If save with no tag, match image without tag, just match repo name.

So, two functions maybe OK. Match and MatchWithoutTag()
```

api/router.go
```diff
@@ -29,7 +29,7 @@ var routes = map[string]map[string]handler{
 		"/images/viz":                     notImplementedHandler,
 		"/images/search":                  proxyRandom,
 		"/images/get":                     getImages,
-		"/images/{name:.*}/get":           proxyImage,
+		"/images/{name:.*}/get":           getImage,
```
vieux: @jimmyxian I don't get why do you need a custom handler here, using proxy ensure we have the same behaviour as the docker daemon.

jimmyxian: @vieux Actually, not a custom handler. The difference is the `image match`
```
+// GET /images/{name:.*}/get
+func getImage(c *context, w http.ResponseWriter, r *http.Request) {
+ name := mux.Vars(r)["name"]
+
+     for _, image := range c.cluster.Images() {
+         if len(strings.SplitN(name, ":", 2)) == 2 && image.Match(name) ||
+           len(strings.SplitN(name, ":", 2)) == 1 && image.MatchWithoutTag(name) {
+                proxy(c.tlsConfig, image.Engine.Addr, w, r)
+                return
+          }
+     }
+     httpError(w, fmt.Sprintf("No such image: %s", name), http.StatusNotFound)
+}
```

jimmyxian:
ping @vieux 
Done to merge two functions. Check again, thanks

vieux:
LGTM

vieux:
ping @abronan

aluzzardi:
@vieux The code looks good to me but I'm not familiar at all with the logic so if you say it's good let's merge it

PR #819, opened 2015-05-20
Leader Election

aluzzardi:
This PR contains the Leader Election algorithm built on top of `pkg/store`.

It is designed to work on top of Consul, etcd and Zookeeper but is currently restricted to Consul due to limitations in `pkg/store` which we are going to address soon.

Follow-up PRs will actually integrate this into the Swarm manager to perform leader election (using `leadership.Candidate`) and automatically forward API requests to the current leader (using `leadership.Follower`).

### DIFF
--- a/leadership/README.md
+++ b/leadership/README.md
@@ -19,7 +19,7 @@ if err != nil {
 underwood := leadership.NewCandidate(client, "service/swarm/leader", "underwood")
 underwood.RunForElection()
 
-for elected := range candidate.ElectedCh {
+for elected := range underwood.ElectedCh {
 	// This loop will run every time there is a change in our leadership
 	// status.
 
--- a/leadership/README.md
+++ b/leadership/README.md
@@ -1,0 +1,58 @@
+# Leadership: Distributed Leader Election for Clustered Environments.
+
+Leadership is a library for a cluster leader election on top of a distributed
+Key/Value store.
+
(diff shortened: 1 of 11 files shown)


leadership/README.md
```diff
@@ -0,0 +1,58 @@
...
+// Create a store using pkg/store.
+client, err := store.NewStore("consul", []string{"127.0.0.1:8500"}, &store.Config{})
+if err != nil {
+	panic(err)
+}
+
+underwood := leadership.NewCandidate(client, "service/swarm/leader", "underwood")
+underwood.RunForElection()
+
+for elected := range candidate.ElectedCh {
```
abronan: `candidate.ElectedCh` -> `underwood.ElectedCh` ;)

abronan:
LGTM! just this tiny issue on the doc.

/cc @vieux

vieux:
LGTM after the doc change

aluzzardi:
Docs updated. PTAL

aluzzardi:
@kelseyhightower @xiang90: We are building leader election for Swarm on top of `etcd`/`zk`/`consul`.

Both `consul` and `zk` provide a `lock` mechanism in their client library, however we couldn't find a good way to do this in `etcd`.

The etcd `leader election` module is deprecated and I couldn't find any helpers in `go-etcd` to talk to the `lock` module.

Could you point us in the right direction? Thanks!

PR #823, opened 2015-05-21
create containers.go

vieux:
Today the exact same code is copy pasted into the mesos branch.
With this pr you can share the code.

### DIFF
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -24,9 +24,11 @@ type Cluster interface {
 	RemoveImage(image *Image) ([]*dockerclient.ImageDelete, error)
 
 	// Return all containers
-	Containers() []*Container
+	Containers() Containers
 
 	// Return container the matching `IDOrName`
+	// TODO: remove this method from the interface as we can use
+	// cluster.Containers().Get(IDOrName)
 	Container(IDOrName string) *Container
 
 	// Pull images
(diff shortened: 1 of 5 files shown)


vieux:
ping @aluzzardi

cluster/containers.go
```diff
@@ -0,0 +1,64 @@
...
+	"strings"
+
+	"github.com/docker/docker/pkg/stringid"
+)
+
+// Containers represents a list a containers
+type Containers []*Container
+
+// Match returns a container using it's ID or Name
+func (containers Containers) Match(IDOrName string) *Container {
```
aluzzardi: I'd use something else than `Match` since this doesn't match (a match is typically a boolean operation).

`Find` or `Get` might work

cluster/containers.go
```diff
@@ -0,0 +1,64 @@
+package cluster
```
aluzzardi: I'd move this into `container.go` instead of having a separate file

aluzzardi: Also, please grep for `[]*Container` in the project (probably `Engine` and so on) and replace it with `Containers`

vieux:
@aluzzardi please take another look
@abronan please take a a look

aluzzardi:
LGTM

PR #824, opened 2015-05-21
Adding export Init method for the exported discovery components

mavenugo:
Since Swarm discovery is used by libnetwork without the swarm main(), it
is forced to use annonymous import in order to force-trigger the
unexported init() methods. Using annonymous import is highly discouraged
in a Non-Main and Non-Test packages.

Introducing exported Init() methods on these files so that libnetwork
can use Discovery as a library without having a main package.

Signed-off-by: Madhu Venugopal <madhu@docker.com>

### DIFF
--- a/discovery/file/file.go
+++ b/discovery/file/file.go
@@ -16,6 +16,11 @@ type Discovery struct {
 }
 
 func init() {
+	Init()
+}
+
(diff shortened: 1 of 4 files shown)


aluzzardi:
LGTM

PR #825, opened 2015-05-21
store: Implement locking mechanisms with Etcd

abronan:
Its seems that F.Underwood can now run for the `Etcd` State ;)

Tested using the `leadership` package documentation example.

### DIFF
--- a/pkg/store/etcd.go
+++ b/pkg/store/etcd.go
@@ -21,6 +21,7 @@ type etcdLock struct {
 	stopLock chan struct{}
 	key      string
 	value    string
+	last     *etcd.Response
 	ttl      uint64
 }
 
@@ -322,11 +323,12 @@ func (s *Etcd) NewLock(key string, options *LockOptions) (Locker, error) {
 func (l *etcdLock) Lock() (<-chan struct{}, error) {
 
 	key := normalize(l.key)
-	var lastIndex uint64
 
 	// Lock holder channels
 	lockHeld := make(chan struct{})
 	stopLocking := make(chan struct{})
+
+	var lastIndex uint64
 
 	for {
 		resp, err := l.client.Create(key, l.value, l.ttl)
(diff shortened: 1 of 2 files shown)


abronan:
There was two choices of implementation:

- *A la Zookeeper*: With directory sequential child keys and by watching the minimum child index
- Using the trick of `Create/Delete` which only let one process pass successfully, we let other processes face the CAS wall with an incorrect index and watch the key to wait for `delete`/`expire` events to occur

Overall the second solution seems cleaner and we use a single key for the lock (compared to `X` child keys for the first option)

abronan:
ping @aluzzardi

pkg/store/etcd.go
```diff
@@ -280,5 +293,117 @@ func (s *Etcd) DeleteTree(prefix string) error {
...
+
+// Hold the lock as long as we can
+// Updates the key ttl periodically until we receive
+// an explicit stop signal from the Unlock method
+func (l *etcdLock) holdLock(key string, lockHeld chan struct{}, stopLocking chan struct{}) {
+	defer close(lockHeld)
+
+	update := time.NewTicker(defaultUpdateTime)
+
+	for {
```
aluzzardi: Paranoia level 12: At some point, we should also watch the lock and make sure no-one stole it from us (similar to what Consul does). The reason is that anyone can `Unlock` our lock, but also, as a result of a network partition conflict resolution, our lock may be gone.

Not worth doing it now, but perhaps add a FIXME and/or create an issue on GH

pkg/store/etcd.go
```diff
@@ -280,5 +293,117 @@ func (s *Etcd) DeleteTree(prefix string) error {
...
+		select {
+		case <-update.C:
+			_, err := l.client.Update(key, l.value, l.ttl)
+			if err != nil {
+				return
+			}
+
+		case <-stopLocking:
+			update.Stop()
```
aluzzardi: Should probably `defer` this after creatiion?

aluzzardi:
LGTM with a few minor comments

pkg/store/etcd.go
abronan: Will try to correct this directly in the PR by:
- Updating the index at each refresh and use `CompareAndDelete` to avoid anyone `Unlocking` our key.
- Watch for unsolicited modifications on the key by watching `update` and `set` events and restoring the leader with another `Update`

abronan: @aluzzardi Well I attempted to do this but this is actually trickier because a race can occur between the `Update` and `CompareAndDelete` (with a wrong index) which will leave the cluster with no master until the key expires.

If we attempt the CAS multiple time to mitigate the race, processes not holding the lock are going to loop indefinitely because there is no way to succeed the CAS call and go out of the loop (unless saying we try the call 5 times and otherwise we break out of the loop, but this is a nasty workaround).

Same about watching the key, we are going to take actions on something that we probably did (let's say we deleted the key, we need to check wether it's us or another process that issued the action).

I'll leave this as it is with a FIXME, I already have a branch with those changes but I need to find a better solution to this issue :(

abronan: Ignore my last comment, I'll try something else..

pkg/store/etcd.go
abronan: +1

abronan:
@aluzzardi PTAL Fixed, now the key cannot be deleted by a process who does not own the lock.

PR #827, opened 2015-05-21
store: Enhance zookeeper lock with value support

abronan:
Add support for setting values in zookeeper lock.

This is pretty much all we need for leader election as `go-zookeeper` already has support for locks.

### DIFF
--- a/pkg/store/zookeeper.go
+++ b/pkg/store/zookeeper.go
@@ -252,6 +252,7 @@ func (l *zookeeperLock) Lock() (<-chan struct{}, error) {
 
 	if err == nil {
 		// We hold the lock, we can set our value
+		// FIXME: When the last leader leaves the election, this value will be left behind
 		_, err = l.client.Set(l.key, l.value, -1)
 	}
 
--- a/pkg/store/zookeeper.go
+++ b/pkg/store/zookeeper.go
@@ -17,7 +17,10 @@ type Zookeeper struct {
 }
 
 type zookeeperLock struct {
-	lock *zk.Lock
+	client *zk.Conn
+	lock   *zk.Lock
+	key    string
+	value  []byte
 }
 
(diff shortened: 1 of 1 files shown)


pkg/store/zookeeper.go
```diff
@@ -225,20 +228,34 @@ func (s *Zookeeper) AtomicDelete(key string, previous *KVPair) (bool, error) {
...
-		return nil, err
+	err := l.lock.Lock()
+
+	if err == nil {
+		// We hold the lock, we can set our value
+		_, err = l.client.Set(l.key, l.value, -1)
 	}
-	return make(<-chan struct{}), nil
+
+	return nil, err
```
abronan: @aluzzardi Do we need to initialize that chan? Seems that it is working without it being initialized with the `leadership` package.

aluzzardi: `nil` channels behave magically and never block during a `select` so it's all good

pkg/store/zookeeper.go
```diff
@@ -225,20 +228,34 @@ func (s *Zookeeper) AtomicDelete(key string, previous *KVPair) (bool, error) {
...
 // Lock attempts to acquire the lock and blocks while doing so.
 // Returns a channel that is closed if our lock is lost or an error.
 func (l *zookeeperLock) Lock() (<-chan struct{}, error) {
-	if err := l.lock.Lock(); err != nil {
-		return nil, err
+	err := l.lock.Lock()
+
+	if err == nil {
+		// We hold the lock, we can set our value
+		_, err = l.client.Set(l.key, l.value, -1)
```
aluzzardi: `// FIXME: When the last leader leaves the election, this value will be left behind`

aluzzardi:
LGTM with minor comment

abronan:
@aluzzardi PTAL

PR #828, opened 2015-05-21
Godeps for Mesos

jimenez:
Signed-off-by: Isabel Jimenez <contact.isabeljimenez@gmail.com>


jimenez:
I'm know this is huge, but once compiled, the binary will be only 2 mb bigger. Also, as soon as we have out of process plugins for swarm, this code will go away, so this is only "temporary"

abronan:
@jimenez :+1:

vieux:
LGTM

abronan:
LGTM

PR #829, opened 2015-05-21
Reconfigure for documentation tooling

moxiegirl:
Consolidate documentation under a new docs directory

I haven't removed the old readmes. Over time, the content in your READMEs should diverge from the documentation. 


moxiegirl:
ping @aluzzardi I'd like to merge this before you cut the release branch.  Let me know if you have any questions.

vieux:
I wouldn't include "docs/token.md", in my opinion it's focused for devs, it should stay as a "README.md" only.

moxiegirl:
@vieux Ok, I can pull that pretty easy.  You'll note I didn't touch y'alls README's :-D anticipating just the statement.

moxiegirl:
@vieux done

vieux:
thanks, LGTM

ping @aluzzardi

Makefile
```diff
@@ -0,0 +1,15 @@
+# Currently this just contains the existing docs build cmds from docker/docker
```
aluzzardi: If possible, I'd like to avoid a root `Makefile` (and having a `docs/Makefile` instead).

This is because it goes against `golang` conventions: a proper go project should be buildable with `go build` and testable with `go test` out of the box, no Makefile required.

As a reference, none of the official `golang` repositories have a `Makefile`:
https://github.com/golang

docs/discovery.md
```diff
@@ -0,0 +1,274 @@
...
+swarm manage -H tcp://<swarm_ip:swarm_port> file:///tmp/my_cluster
+```
+
+And for the `nodes` discovery method.
+
+```bash
+swarm manage -H <swarm_ip:swarm_port> "nodes://10.0.0.[10:200]:2375,10.0.1.[2:250]:2375"
+```
+
+## Contributing a new discovery backend
```
aluzzardi: Maybe we should keep this part in the `README.md` file as it's intended for Swarm developers rather than users.

aluzzardi:
Just a few minor comments otherwise LGTM

Makefile
moxiegirl: I can do that.

docs/discovery.md
moxiegirl: @aluzzardi  Well, we do have developer docs elsewhere in the tree.  Meaning, API and how to develop in Docker.  Keeping this in here informs people that they can extend Swarm --- though eventually we will want to separate user from developer material and not mash it into one file.

What do you think? Ok to keep for reason stated above or pull it --- you have the final call.

aluzzardi: I think the API is still part of the user docs (as in, how do I use this thing with the API as an app developer), not sure about the "how to develop in Docker" though (which file is that?).

I agree that it's nice to inform people that they can extend Swarm, what about changing:
> Contributing a new discovery backend is easy, simply implement this interface:

To:
> Contributing a new discovery backend is easy, please refer to [link to README]

And end the documentation there?

Beside keeping the documentation simpler/shorter, it also means that people will always see our latest (master) README rather than the one from the docs branch and changing the API won't require a docs push.

For instance, the stated interface here is wrong as it got changed a week or so ago and we forgot to update the README :) (`Fetch` is gone, the signature fo `Watch` is now different).

moxiegirl: Sounds good to me.  I'll make this fix and push.  I can't merge on your project --- if you want to give me access, I promise to be really careful.

moxiegirl:
ping @vieux @aluzzardi ready to go! Thanks for your help getting this in place. I really appreciate it.

PR #832, opened 2015-05-21
add getter/setter for namespaced labels

vieux:
This is useful for mesos.

@aluzzardi @abronan I'm not sure about the function names though, any advice ?


aluzzardi:
LGTM

Issue #817, opened 2015-05-19
Feedback: `--addr` is very confusing


aluzzardi:
One of the first steps to set up a cluster is to run `swarm join --addr=<address> <discovery>`.

Many people mistake `--addr` for the manager address instead of the docker address.

We should find an alternative name, perhaps:
- `--cluster-addr`
- `--peer-addr`
- `--my-addr`
- `--engine-addr`
- `--docker-addr`

Additionally, perhaps it would help if we'd pre-fill that option. In order to do that, we'd need to run `swarm join` with `--net=host` and grab the default interface's IP address.

Thoughts?

aluzzardi:
See also #741

squillace:
See also [#437](https://github.com/docker/swarm/issues/437). :-)

aluzzardi:
@squillace Thanks! Re-opening that, we should fix the docs.

thaJeztah:
My first idea was `-H`, like the regular flag to specify the docker host, but that's confusing, because `swarm manage` uses that. I like `--docker-addr`, `--docker-address` or perhaps `--docker-host`.

UserTaken:
I like `--node-addr` since isn't that the offical term when running `info` and all in the docs?

aluzzardi:
@UserTaken That's another thing - I'm unsure about the term `node`, wondering if it should be `engine`. Thoughts?

thaJeztah:
Engine, without the "docker" context ("Docker Engine") ... not sure. Daemon might ring more bells on the command line.

UserTaken:
I think `node` is fitting as it is a de facto term used in explaining a distributed systems such as Consul, Etcd, Fleet, Kubernetes, cryptocurrency, etc.

thaJeztah:
While I agree that `node` is used as the "official" term; the example, as mentioned in https://github.com/docker/swarm/issues/741 shows `$ swarm join --addr=node_ip:2375 consul://consul_addr/path`

The `node_ip` apparently wasn't enough to clarify what had to be entered;

> what is the "node_ip" and what is the 2375 port? and can it be changed?

For consistency, `--node-addr` / `--node-address` sounds good, but it still needs a good explanation around that in the docs. I don't think this can be solved by only renaming the option.

abronan:
I agree with @thaJeztah and @UserTaken for the `node` term.

In comparison:
[...]

PR #833, opened 2015-05-22
Add example of secured setup using docker-machine, marks the manual setup as insecure

abronan:
Change the setup explanations to use  `docker-machine` instead which is secured by default.

Marked the manual setup as **insecure**.

It's not ideal but this is a placeholder until the doc overall with more consistent informations on how to setup a secured cluster manually.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -69,7 +69,116 @@ The easiest way to get started with Swarm is to use the
 $ docker pull swarm
 ```
 
-## Set up Swarm nodes
+## Set up a secured Swarm cluster using docker-machine
+
+`docker-machine` is a tool to provision docker nodes locally or within a Cloud provider infrastructure.
+
+You can setup a Swarm cluster (**secured by default** using TLS) with this tool.
+
+### Installation
+
+See the steps in the [Docker Machine documentation](https://docs.docker.com/machine/) to install `docker-machine`
+
+Make sure you have **Virtualbox** installed locally, this will be the driver used to create our Swarm Virtual Machines
+
+### Create a token
+
+First, create a Swarm token.  Optionally, you can use another discovery service.
+See the Swarm documentation on alternative solutions in [Discovery Documentation](https://github.com/docker/swarm/blob/master/discovery/README.md)
+
+To create the token, first create a Machine. This example will use VirtualBox.
+
+```
+$ docker-machine create -d virtualbox local
+```
+
+Load the Machine configuration into your shell:
+
+```
+$ eval "$(docker-machine env local)"
+```
+
+Then run generate the token using the Swarm Docker image:
+
+```
+$ docker run swarm create
+1257e0f0bbb499b5cd04b4c9bdb2dab3
+```
+Once you have the token, you can create the cluster.
+
+### Launch the swarm manager
+
+Use this command to launch the *Swarm Manager*:
+
+```
+docker-machine create \
+    -d virtualbox \
+    --swarm \
+    --swarm-master \
+    --swarm-discovery token://<TOKEN-FROM-ABOVE> \
+    swarm-master
+```
+
+The *Swarm Manager* is the machine in charge of orchestrating and scheduling containers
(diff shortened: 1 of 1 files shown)


UserTaken:
Using both the term `node` and `agent` could cause unnecessary confusion when other documentation I've seen use just the term `node` and `join --addr` will be changed to `join --node-addr`  soon.

abronan:
Thanks @UserTaken I added a quick fix on the usage. Still confusing but hopefully this will evolve after the discussions on #817 and the use of `--node-addr` instead.

abronan:
ping @aluzzardi @vieux to merge after #829 I guess

vieux:
LGTM

thaJeztah:
:heart: thanks!

squillace:
:+1:

aluzzardi:
NOT LGTM

We should have @moxiegirl review this.

We've just coupled Swarm with Machine - something we didn't want to do from the get go.

Also, if I follow the documentation step by step, I expect to get a working Swarm cluster. Instead, I'll likely get a `docker-machine: command not found`.

abronan:
@aluzzardi @moxiegirl is actually improving this doc. We know it's not ideal but we need to let people get started with Swarm with a secured setup even though this means going through `docker-machine` and using 2 tools instead of one. We'll go back on this when we'll have a story to tell to people on how to setup the cluster manually and securely using TLS.

moxiegirl:
@aluzzardi  My intent is to position this as a way to get up running with a test.  I was able to get up a swarm up running very quickly with `docker-machine` which was impressive.  I thought it was a powerful demonstration of Docker's tool in general.  What I can do is move people quickly into using some Docker Swarm commands.  

I do think it is important to show the "manual" way of installing as well. I want to keep that in a separate document.  @abronan says that this is a recommended way to do it for development. Do you @aluzzardi agree or disagree with that notion. Just want to make sure everyone is on the same page. @vieux you too! Everyone agree the manual way is likely only in a dev env?

aluzzardi:
I do agree that `docker-machine` is the easy way and we should document that, probably the default to get started and experiment with swarm.

However, I disagree that manual is likely only for dev. Quite the opposite: Swarm is a tool meant for production environments, therefore the primary target is **manual**.

Unlike other docker tools which are meant for dev (compose) or a mix of dev&prod (machine/engine), swarm is meant to be used in production and is mostly useless for developers.

In the *getting started with docker-machine* example, we let people experiment with swarm using virtualbox in a matter of seconds which is great. However, nobody will actually run their production environments out of their laptops, therefore if they like what they see and want to use it for real, they will have to go through the manual process.

Granted, machine will allow you to provision cloud providers (EC2, Azure, ...) which **can** be used for production. However, this assumes that they are just getting started and have no production machines yet. If you already have a production environment and want to use Swarm, you will have to go through the manual setup as well.

Ping machine maintainers for feedback /cc @ehazlett @nathanleclaire 
Ping contributors who use swarm in non-dev environments /cc @chanwit @smothiki @tnachen @jhamrick @jmmills

abronan:
Agreed with @aluzzardi. It is recommended for dev but this is not likely *only* as a dev environment. Manual should be the default path for everyone as we target production setups.

I disagree though with the fact that this is mostly useless for developers. Swarm is a great playground for anyone who wants to get introduced to clustering/orchestration tools and/or test a cluster deployment on a local setup, so we target developers as well.

Having `docker-machine` as the setup process is good for now, but we definitely want to work on the story regarding manual setups and how to handle certs creation, etc.

moxiegirl:
@aluzzardi  My understanding is that Swarm is still in Beta. We don't recommend it for anyone's production regardless of how it is installed. So, I left notes about Swarm not being production ready in the documentation.

Is 1.7 going to be production ready?

chanwit:
What our team have done is that tweaking docker-machine to allow it to create a none-driver machine with both server and client certs.

It allows us to trivially setup Swarm with TLS and play nice with the current workflow.

moxiegirl:
Hey gang, this is already merged. Can we move comments and discussion to 839 which revised everything including this piece?

PR #840, opened 2015-05-25
Modified command to run all the tests

rajdeepd:
Signed-off-by: Rajdeep Dua <dua_rajdeep@yahoo.com>


vieux:
@rajdeepd the `$GOBIN/godep restore` will take all the godeps and put them in the gopath, so there is no need to use gobin after that.

PR #841, opened 2015-05-25
Additional check for ttl < 10s and Consul

### LINKED ISSUE #838: Discovery broken with low TTL
On master, this works:

```
go run main.go join --ttl 10s --heartbeat 1s --addr 127.0.0.1:4242 consul://localhost:8500/test
INFO[0000] Registering on the discovery service every 1s...  addr=127.0.0.1:4242 discovery=consul://localhost:8500/test
```

However this fails mysteriously:

```
go run main.go join --ttl 2s --heartbeat 1s --addr 127.0.0.1:4242 consul://localhost:8500/test
INFO[0000] Registering on the discovery service every 1s...  addr=127.0.0.1:4242 discovery=consul://localhost:8500/test
ERRO[0000] Unexpected response code: 400 (Missing session)
```

abronan:
Fixes #838 

### DIFF
--- a/discovery/kv/kv.go
+++ b/discovery/kv/kv.go
@@ -62,14 +62,7 @@ func (s *Discovery) Initialize(uris string, heartbeat time.Duration, ttl time.Du
 		},
 	)
 
-	if err != nil {
-		if err == store.ErrInvalidTTL {
-			log.Fatal(err)
-		}
-		return err
-	}
-
-	return nil
+	return err
 }
 
 // Watch the store until either there's a store error or we receive a stop request.
--- a/discovery/kv/kv.go
+++ b/discovery/kv/kv.go
@@ -61,7 +61,15 @@ func (s *Discovery) Initialize(uris string, heartbeat time.Duration, ttl time.Du
 			EphemeralTTL: s.ttl,
 		},
 	)
-	return err
+
+	if err != nil {
+		if err == store.ErrInvalidTTL {
+			log.Fatal(err)
+		}
+		return err
+	}
+
+	return nil
 }
 
 // Watch the store until either there's a store error or we receive a stop request.
--- a/pkg/store/consul.go
+++ b/pkg/store/consul.go
@@ -16,6 +16,10 @@ const (
 	// watched key has changed.  This affects the minimum time it takes to
 	// cancel a watch.
 	DefaultWatchWaitTime = 15 * time.Second
+
+	// MinimumTimeToLive is the minimum TTL value allowed by Consul for
+	// Ephemeral entries
+	MinimumTimeToLive = 10 * time.Second
 )
 
 // Consul embeds the client and watches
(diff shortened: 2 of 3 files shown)


abronan:
ping @aluzzardi

pkg/store/consul.go
```diff
@@ -81,8 +85,11 @@ func (s *Consul) setTimeout(time time.Duration) {
 }
 
 // SetEphemeralTTL sets the ttl for ephemeral nodes
-func (s *Consul) setEphemeralTTL(time time.Duration) {
-	s.ephemeralTTL = time
+func (s *Consul) setEphemeralTTL(ttl time.Duration) {
+	if ttl < MinimumTimeToLive {
+		log.Fatal("Consul does not allow a ttl < 10s, please specify a ttl >= 10s")
```
aluzzardi: I believe a library should never Fatal - we should return an error in initialize

abronan: Ok, almost forgot that store is in its own package now..

pkg/store/consul.go
```diff
@@ -52,7 +56,10 @@ func InitializeConsul(endpoints []string, options *Config) (Store, error) {
 			s.setTimeout(options.ConnectionTimeout)
 		}
 		if options.EphemeralTTL != 0 {
-			s.setEphemeralTTL(options.EphemeralTTL)
+			err := s.setEphemeralTTL(options.EphemeralTTL)
```
vieux: nit: can be merged into a single line

pkg/store/consul.go
```diff
@@ -81,8 +88,12 @@ func (s *Consul) setTimeout(time time.Duration) {
 }
 
 // SetEphemeralTTL sets the ttl for ephemeral nodes
-func (s *Consul) setEphemeralTTL(time time.Duration) {
-	s.ephemeralTTL = time
+func (s *Consul) setEphemeralTTL(ttl time.Duration) error {
```
vieux: can we add a test for `setEphemeralTTL`

discovery/kv/kv.go
```diff
@@ -61,7 +61,15 @@ func (s *Discovery) Initialize(uris string, heartbeat time.Duration, ttl time.Du
 			EphemeralTTL: s.ttl,
 		},
 	)
-	return err
+
+	if err != nil {
+		if err == store.ErrInvalidTTL {
```
aluzzardi: Is it really necessary to check this?

AFAIK `CreateStore` never fails unless:
- The backend doesn't exist
- The TTL is invalid

Both should just fail

pkg/store/consul.go
abronan: Don't know where to put the test for now to be honest as we only rely on mocking with `pkg/store`.

vieux: ok

vieux:
LGTM

aluzzardi:
LGTM

PR #845, opened 2015-05-26
[EXPERIMENTAL] Adding Mesos support

jimenez:
This PR adds all the code for supporting Mesos, although all the flags are commented as it requires changes in Mesos for the support to work. 
The main code base is barely touched, most of the added code is on `cluster/mesos/ `, the only exception being in `engine.go` for `RefreshContainers` that needed to be exported for this first version.
ping @aluzzardi  @abronan @tnachen @vieux


shykes:
Nice :) It's great to see this making progress. Thanks @jimenez !

jimenez:
@aluzzardi  I expected the coverage to decrease, but 7.46 seems odd and a bit lot to me. Should we exclude tests in `cluster/mesos` for now? `slave.go` and `task.go` have proper unit tests, like for `swarm/cluster.go` we didn't do much tests for `mesos/cluster.go` (it required mocking)

aluzzardi:
I'll make sure to review and merge this tomorrow.

I find the coverage drop concerning, although given the time constraint I'm okay to close my eyes for now as long as we make it a priority to raise it back to normal levels before the final release :)

Regarding `flags.go`, `help.go` and `manage.go`: How about excluding those files for the time being? I'd rather avoid having commented code (it will get broken as we have competing PRs on those files, for instance leader election)

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	// Register the container immediately while waiting for a state refresh.
+	// Force a state refresh to pick up the newly created container.
+	// FIXME: unexport this method, see FIXME in engine.go
+	s.engine.RefreshContainers(true)
+
+	// TODO: We have to return the right container that was just created.
+	// Once we receive the ContainerID from the executor.
+	for _, container := range s.engine.Containers() {
```
aluzzardi: I don't get this - we are iterating over `Containers` and when the first arrives, we push it to a channel and return immediately?

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+			out = append(out, container)
+		}
+	}
+
+	return out
+}
+
+// Container returns the container with IdOrName in the cluster
+func (c *Cluster) Container(IDOrName string) *cluster.Container {
+	// Abort immediately if the name is empty.
```
aluzzardi: // FIXME: Duplicated code.

Not sure how to un-duplicate this

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// Load images
+func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)) {
+
+}
+
+// RenameContainer Rename a container
+func (c *Cluster) RenameContainer(container *cluster.Container, newName string) error {
+	//FIXME this probably doesn't work as the next refreshcontainer will erase this change
```
aluzzardi: Yeah, this change is in-memory only

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+	scheduler    *scheduler.Scheduler
+	options      *cluster.DriverOpts
+	store        *state.Store
+	TLSConfig    *tls.Config
+	master       string
+	pendingTasks *queue.Queue
+	offerTimeout time.Duration
+}
+
+var (
```
aluzzardi: Those should probably be `const`?

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+	log.WithFields(log.Fields{"name": "mesos"}).Debug("Initializing cluster")
+
+	cluster := &Cluster{
+		slaves:       make(map[string]*slave),
+		scheduler:    scheduler,
+		options:      &options,
+		store:        store,
+		master:       master,
+		TLSConfig:    TLSConfig,
+		offerTimeout: 10 * time.Minute,
```
aluzzardi: should probably be a `const defaultOfferTimeout` at the top of the file

cluster/mesos/cluster.go
tnachen: This is only a hack for now until we get the necessary mesos change in, so we know exactly what container we launched.

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+
+	driverConfig := mesosscheduler.DriverConfig{
+		Scheduler: cluster,
+		Framework: &mesosproto.FrameworkInfo{Name: &frameworkName, User: &user},
+		Master:    cluster.master,
+	}
+
+	// Changing port for https
+	if cluster.TLSConfig != nil {
+		dockerDaemonPort = "2376"
```
aluzzardi: Perhaps we should have two consts in the file: `defaultDockerEnginePort` and `defaultDockerEngineTLSPort` and have a member in the struct named `dockerEnginePort`.

This check right here should decide whether to set `cluster.dockerEnginePort = defaultDockerEnginePort` or `cluster.dockerEnginePort = defaultDockerEngineTLSPort`.

Otherwise, as is, this is a global variable and we try to stay clear from those

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+// Containers returns all the containers in the cluster.
+func (c *Cluster) Containers() []*cluster.Container {
+	c.RLock()
+	defer c.RUnlock()
+
+	out := []*cluster.Container{}
+	for _, s := range c.slaves {
+		for _, container := range s.engine.Containers() {
+			if name := container.Config.Labels[cluster.SwarmLabelNamespace+".mesos.name"]; name != "" && container.Names[0] != "/"+name {
+				container.Names = append([]string{"/" + name}, container.Names...)
```
aluzzardi: This is a bit strange. I see what it does, but container retrieval is probably not the best spot for rewriting names.

For instance, in `Container(IDOrName string)`:
```
+		if container.Id == IDOrName || stringid.TruncateID(container.Id) == IDOrName {
+			return container
+		}
```

Here we forgot to append the name, as well in multiple other places.

Maybe, as a temporary workaround (this is still ugly), you could add a helper function to `Cluster` (`formatContainer`?) which takes care of name cleanup and is called whenever `Cluster` returns a container (`Container()`, `Containers()`, and what about events actually?)

aluzzardi:
Thank you a lot for this, @jimenez @tnachen @vieux.

I made a preliminary review, will spend more time tomorrow.

cluster/mesos/cluster.go
aluzzardi: Oh, @vieux already found a way to dedup this in #823 :)

cluster/mesos/cluster.go
aluzzardi: I see!

Well, in case the Mesos change doesn't arrive on time (or actually, even to make Swarm work with the latest mesos release without requiring master), what if we injected a label into the container then use said label to find our container?

aluzzardi: We are already doing this:
```
config.Labels[cluster.SwarmLabelNamespace+".mesos.name"] = name
```

We could perhaps either inject something we would be able to retrieve (such as a task id), or if nothing else is available, we could make something out of thin air (just like `.swarm.id`) and grab it back.

aluzzardi: Thoughts? /cc @tnachen @jimenez @vieux

cluster/mesos/task.go
```diff
@@ -0,0 +1,162 @@
...
+
+		cluster:   c,
+		config:    config,
+		error:     make(chan error),
+		container: make(chan *cluster.Container),
+	}
+
+	ID := stringid.TruncateID(stringid.GenerateRandomID())
+	if name != "" {
+		ID = name + "." + ID
```
aluzzardi: why?

cluster/mesos/task.go
```diff
@@ -0,0 +1,162 @@
...
+	task := task{
+		updates: make(chan *mesosproto.TaskStatus),
+
+		cluster:   c,
+		config:    config,
+		error:     make(chan error),
+		container: make(chan *cluster.Container),
+	}
+
+	ID := stringid.TruncateID(stringid.GenerateRandomID())
```
aluzzardi: lowercase `id` should be good I believe

cluster/mesos/task.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+
+	t.SlaveId = &mesosproto.SlaveID{Value: &slaveID}
+}
+
+func newTask(c *Cluster, config *cluster.ContainerConfig, name string) (*task, error) {
+	// save the name in labels as the mesos containerizer will override it
+	config.Labels[cluster.SwarmLabelNamespace+".mesos.name"] = name
+
+	task := task{
```
aluzzardi: Should probably be `&task` here at creation and just `return task, nil` at return time

cluster/mesos/cluster.go
aluzzardi: For instance, `newTask` could do something like:
```
id := stringid.TruncateID(stringid.GenerateRandomID())
config.Labels[cluster.SwarmLabelNamespace+".mesos.task"] = id
config.Labels[cluster.SwarmLabelNamespace+".mesos.name"] = name
[...]
task.TaskId = &mesosproto.TaskID{Value: &id}
```

And here, in the loop, find the container where `config.Labels[cluster.SwarmLabelNamespace+".mesos.task"] == t.TaskId.Value`

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+	}
+
+	go c.pendingTasks.Add(task)
+
+	select {
+	case container := <-task.container:
+		return container, nil
+	case err := <-task.error:
+		return nil, err
+	case <-time.After(5 * time.Second):
```
aluzzardi: This should be at least a const, probably an option later on.

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,485 @@
...
+	TLSConfig    *tls.Config
+	master       string
+	pendingTasks *queue.Queue
+	offerTimeout time.Duration
+}
+
+var (
+	frameworkName    = "swarm"
+	dockerDaemonPort = "2375"
+	errNotSupported  = errors.New("not supported with mesos")
```
aluzzardi: Doesn't seem to be used

cluster/mesos/cluster.go
jimenez: Sounds good, let me try

cluster/mesos/task.go
vieux: It's a "common practice" in mesos, see some marathon task in the screenshot.
screen shot 2015-05-26 at 11 26 06

cluster/mesos/cluster.go
aluzzardi: Once #823 is merged (in a few minutes), could you change the `Container()` implementation?

jimenez: Sure :)

cluster/mesos/task.go
aluzzardi: Sounds good!

jimenez:
ping @aluzzardi , addressed most of the comments, should I un comment modifications in cli manage and help?

aluzzardi:
Yes! :)

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+		Framework: &mesosproto.FrameworkInfo{Name: proto.String(frameworkName), User: &user},
+		Master:    cluster.master,
+	}
+
+	// Changing port for https
+	if cluster.TLSConfig != nil {
+		cluster.dockerEnginePort = defaultDockerEngineTLSPort
+	}
+
+	if bindingPort, ok := options.Uint("mesos.port", "SWARM_MESOS_PORT"); ok {
```
aluzzardi: what happens if this is not defined?

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+		driverConfig.BindingPort = uint16(bindingPort)
+	}
+
+	if bindingAddress, ok := options.IP("mesos.address", "SWARM_MESOS_ADDRESS"); ok {
```
aluzzardi: same here

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+
+// Containers returns all the containers in the cluster.
+func (c *Cluster) Containers() cluster.Containers {
+	c.RLock()
+	defer c.RUnlock()
+
+	out := cluster.Containers{}
+	for _, s := range c.slaves {
+		for _, container := range s.engine.Containers() {
+			if name := container.Config.Labels[cluster.SwarmLabelNamespace+".mesos.name"]; name != "" && container.Names[0] != "/"+name {
```
aluzzardi: This should probably use `formatContainer`

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+			}
+			out = append(out, container)
+		}
+	}
+
+	return out
+}
+
+func formatContainer(container *cluster.Container) *cluster.Container {
+	if container != nil {
```
aluzzardi: nit: `if container == nil { return nil }` instead of having lots of code within a nested if

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+
+// Load images
+func (c *Cluster) Load(imageReader io.Reader, callback func(what, status string)) {
+
+}
+
+// RenameContainer Rename a container
+func (c *Cluster) RenameContainer(container *cluster.Container, newName string) error {
+	//FIXME this doesn't work as the next refreshcontainer will erase this change (this change is in-memory only)
+	container.Config.Labels[cluster.SwarmLabelNamespace+".mesos.name"] = newName
```
aluzzardi: Probably `return errNotSupported`?

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+}
+
+func (c *Cluster) addOffer(offer *mesosproto.Offer) {
+	s, ok := c.slaves[offer.SlaveId.GetValue()]
+	if !ok {
+		return
+	}
+	s.addOffer(offer)
+	go func(offer *mesosproto.Offer) {
+		<-time.After(c.offerTimeout)
```
aluzzardi: this basically is: `time.Sleep(c.offerTimeout)` :)

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+		if c.removeOffer(offer) {
```
aluzzardi: So, after a while we `removeOffer` the offer all the time?

What does it do exactly?

aluzzardi: I'm sure this is correct, but a comment might help people not familiar with mesos like me

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+			if _, err := c.driver.DeclineOffer(offer.Id, &mesosproto.Filters{}); err != nil {
+				log.WithFields(log.Fields{"name": "mesos"}).Errorf("Error while declining offer %q: %v", offer.Id.GetValue(), err)
+			} else {
+				log.WithFields(log.Fields{"name": "mesos"}).Debugf("Offer %q declined successfully", offer.Id.GetValue())
+			}
+		}
+	}(offer)
+}
+
+func (c *Cluster) removeOffer(offer *mesosproto.Offer) bool {
```
aluzzardi: Looks like this might need some locking?

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+	log.WithFields(log.Fields{"name": "mesos", "offerID": offer.Id.String()}).Debug("Removing offer")
+	s, ok := c.slaves[offer.SlaveId.GetValue()]
```
aluzzardi: Looks like this might need some locking?

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+	t.build(n.ID)
+
+	c.Lock()
+	// TODO: Only use the offer we need
+	offerIds := []*mesosproto.OfferID{}
+	for _, offer := range c.slaves[n.ID].offers {
+		offerIds = append(offerIds, offer.Id)
+	}
+
+	if _, err := c.driver.LaunchTasks(offerIds, []*mesosproto.TaskInfo{&t.TaskInfo}, &mesosproto.Filters{}); err != nil {
```
aluzzardi: I think in Go it should be `offerIDs` (@vieux ?)

cluster/mesos/cluster.go
vieux: we need this for docker-compose

cluster/mesos/cluster.go
```diff
@@ -0,0 +1,453 @@
...
+	finished, err := t.monitor()
+	taskID := t.TaskInfo.TaskId.GetValue()
+	if err != nil {
+		//remove task
+		s.removeTask(taskID)
+		t.error <- err
+		return true
+	}
+	if !finished {
+		go func() {
```
aluzzardi: What's the purpose of this?

It waits until its finished then does nothing, right? I assume it's preliminary code and sometime it's going to do something else. Would you mind adding a comment?

aluzzardi:
:+1: 

A few minor comments

For unit tests, could we exclude them from coverage (`script/coverage`) for now?

aluzzardi:
One last thing: With the last change (task label), does this code work with the latest mesos **release** (as in, not master)?

This would be huge.

cluster/mesos/cluster.go
vieux: yes

cluster/mesos/cluster.go
jimenez: It's auto-generated by mesos-go

cluster/mesos/cluster.go
jimenez: same

cluster/mesos/cluster.go
aluzzardi: :+1:

cluster/mesos/cluster.go
aluzzardi: :+1:

cluster/mesos/cluster.go
aluzzardi: :+1:

cluster/mesos/cluster.go
aluzzardi: :+1:

cluster/mesos/cluster.go
aluzzardi: Is it working? :)

cluster/mesos/cluster.go
jimenez: I changed the comment, and updated the TODO.

aluzzardi: :+1:

jimenez:
It's far from optimal but it should work with the latest mesos release :)

aluzzardi:
:fireworks: :fireworks: :fireworks: LGTM :fireworks: :fireworks: :fireworks:

abronan:
:balloon: :tada: :dancer:

vieux:
LGTM, waiting for all the tests to be green and then we merge.

jimenez:
:tada: :+1: Thanks @aluzzardi @vieux @abronan

PR #852, opened 2015-05-26
Fix zookeeper watch that was triggered only once

### LINKED ISSUE #842: Cluster entry is lost after leaving and rejoining the cluster with Zookeeper
Steps to reproduce using Zookeeper discovery:

- Create a single Manager
- Create 3 Agents and join the cluster
- Kill 1 agent
- Try to rejoin the cluster with that agent

**Expected**:
The entry is added back to the cluster and we see 3 nodes again

**What happens**:
Nothing, the entry is lost and the cluster always shows 2 entries with `docker infos`

abronan:
Fixes #842

### DIFF
--- a/pkg/store/zookeeper.go
+++ b/pkg/store/zookeeper.go
@@ -167,7 +167,7 @@ func (s *Zookeeper) WatchTree(prefix string, stopCh <-chan struct{}) (<-chan []*
 	go func() {
 		defer close(watchCh)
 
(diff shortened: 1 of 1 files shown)


abronan:
ping @aluzzardi @vieux

vieux:
LGTM

aluzzardi:
LGTM

PR #877, opened 2015-05-31
update dockerclient & stretchr & add 'docker run --add-host' param

lokikiller:
update the dependency "dockerclient and strrtchr" and supported "docker run --add-host param"


chanwit:
Hi @lokikiller I think you miss something :) When update the dependencies, the json file of Godeps should be changed and be a part of the commit too.

aluzzardi:
Indeed!

`godep` is not very straightforward to use.

Anyway: One should **never** touch `Godeps/` manually.

In order to update `dockerclient` for instance, you should do something like:

```
go get -u github.com/samalba/dockerclient
godep update github.com/samalba/dockerclient
```

This should take care of updating both `Godeps/Godeps.json` and `Godeps/_workspace`

Maybe we should write something about that in the contributing docs.

@chanwit @jimmyxian thoughts?

chanwit:
@aluzzardi it's partially there in:
https://github.com/docker/swarm/blob/master/CONTRIBUTING.md

I'll update the doc accordingly.

jimmyxian:
@aluzzardi @chanwit 
Yeah.. Also, I think we should add some guidelines in Godeps/Readme

vieux:
I'm closing this one, please ping us when you are ready and we will reopen.

Issue #876, opened 2015-05-31
Swarm doesn't check implicit :latest tag for private registry images on /image/{name}/* API calls


frol:
Reproduction:

```bash
$ curl --cacert ~/.docker/ca.pem --cert ~/.docker/cert.pem --key ~/.docker/key.pem \
    https://swarm.company.com:2377/v1.18/images/private.registry.com:5000/test/json
No such image: private.registry.com:5000/test

$ curl --cacert ~/.docker/ca.pem --cert ~/.docker/cert.pem --key ~/.docker/key.pem \
    https://swarm.company.com:2377/v1.18/images/private.registry.com:5000/test:latest/json
<valid JSON>
```

Docker daemon handles these requests well and returns valid JSON for both.

This bug affects docker-py `client.inspect_image()` and, as a result, docker-compose (where I have started my investigations).

```
Client version: 1.6.2
Client API version: 1.18
Go version (client): go1.4.2
Git commit (client): 7c8fca2
OS/Arch (client): linux/amd64
Server version: swarm/0.3.0
Server API version: 1.16
Go version (server): go1.3.3
Git commit (server): 84652ed
OS/Arch (server): linux/amd64
```

Workaround for docker-compose is to specify `:latest` tag explicitly.

frol:
You may even get such error with `docker-compose up`:

```
Traceback (most recent call last):
  File "<string>", line 3, in <module>
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.main", line 32, in main
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 21, in sys_dispatch
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 34, in dispatch 
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.docopt_command", line 24, in dispatch 
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.command", line 66, in perform_command
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.cli.main", line 470, in up
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.project", line 230, in up
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 333, in execute_convergence_plan 
  File "/code/build/docker-compose/out00-PYZ.pyz/compose.service", line 380, in recreate_container
[...]

PR #878, opened 2015-05-31
fix matching image name with private repo

chanwit:
This PR fixes (issue #876) image matching in case of image name containing a private repository path, like:
```
private.repo.com:5000/name:latest
```

### DIFF
--- a/cluster/image.go
+++ b/cluster/image.go
@@ -13,6 +13,35 @@ type Image struct {
 	Engine *Engine
 }
 
+func toImageName(repo string, name string, tag string) string {
+	fullname := name
+	if tag != "" {
+		fullname = name + ":" + tag
+	}
+	if repo != "" {
+		fullname = repo + "/" + fullname
+	}
+	return fullname
+}
+
+func parseImageName(fullname string) (repo string, name string, tag string) {
+	parts := strings.SplitN(fullname, "/", 2)
+
+	nameAndTag := parts[0]
+	if len(parts) == 2 {
+		repo = parts[0]
+		nameAndTag = parts[1]
+	}
+
+	parts = strings.SplitN(nameAndTag, ":", 2)
+	name = parts[0]
+	if len(parts) == 2 {
+		tag = parts[1]
+	}
+
+	return
+}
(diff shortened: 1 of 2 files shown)


chanwit:
@frol could you please test using codes from this branch? Thanks !!

cluster/image.go
```diff
@@ -21,18 +21,31 @@ func (image *Image) Match(IDOrName string, matchTag bool) bool {
 		return true
 	}
 
+	repo := ""
 	name := IDOrName
+	if strings.Contains(IDOrName, "/") {
+		parts := strings.SplitN(IDOrName, "/", 2)
+		repo = parts[0] + "/"
+		name = parts[1]
+	}
```
frol: This code is misleading and hard to read (I'm talking about this function overall after the fix). My suggestions are the following:

1. Don't put '/' to `repo` as it is not a part of repository name!
1. Finish parsing here - add `tag` var and parse `parts[1]` into real `name` and `tag` instead of this "maybe" `name`
1. We clearly have copy-paste here and 10 lines down, so move this parsing to a separate function!
1. It would be great if this thing could be synced with Docker Daemon code. (I couldn't find it there)

frol:
@chanwit Could you please refactor the code before we continue testing? Nevertheless, I can see from the code that it works fine.

frol:
@chanwit Extracting repo-name-tag parsing will also be useful for `Pull` func.

chanwit:
ping @frol for testing

ping @vieux @aluzzardi @abronan for code reviews
@aluzzardi do I need a function test for this PR?

frol:
Cool! This fixes the issue! (docker-compose now works fine also!)

jimmyxian:
@chanwit Thanks
LGTM

aluzzardi:
@chanwit Awesome!

Will do a code review tomorrow. In the meantime, I believe this does not require integration tests - unit tests are already covering everything.

vieux:
LGTM ping @aluzzardi

aluzzardi:
LGTM

PR #879, opened 2015-05-31
Fix reserved cpus computation

km4rcus:
This should fix the reserved cpus computation https://github.com/docker/swarm/issues/475#issuecomment-95858898

Signed-off-by: Marco Mancini <marcomancini72@gmail.com>

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


vieux:
@km4rcus can you add a test ? Thank you.

km4rcus:
@vieux I added test function and changed also the cpuShare computation when a container is created, otherwise usedCpu is wrong if NCPU is not a power of 2. Should I add to  test function several values for NCPU?

vieux:
@km4rcus yes please if you can.

km4rcus:
@vieux Added two arrays for testing different values for cpus.

vieux:
@aluzzardi @abronan not sure what to do about this one, I would prefer to wait after the RC to merge it.

aluzzardi:
@km4rcus So, actually, `-c 1` on a 1 CPU machine should give you 1024 shares by design.

The logic behind is explained in https://github.com/docker/swarm/issues/475#issuecomment-88258864

vieux:
replaced by #904

PR #881, opened 2015-06-01
store: Fix CAS returning ErrKeyModified because of Lock in Consul

### LINKED ISSUE #875: Newly introduced (#843) Store AtomicPut for Consul under Lock fails
#843 introduced the support for retuning the LastIndex for the AtomicPut operation.
But unfortunately for the Consul backend, the Lock() operation results in the subsequent CAS to fail with ErrKeyModified error in AtomicPut method.

As a quick test, commenting out solved the problem.
```
        /*
                lock, err := s.NewLock(key, nil)
                if err != nil {
                        return false, nil, err
                }

                lock.Lock()
                defer lock.Unlock()
        */
```

abronan:
Fixes #875 

I removed the lock and enforced the *consistency* parameter for the `Get` operation. This is to make sure that we get back the last value and index for the `Get` during a CAS operation. Now I'm not sure if we enforce it for every `Get` or just the one in the `CAS` (this is part of a larger question: should we ensure a strong consistency for every store supported in the *lib* if possible, or offer the choice of consistency in `Options`? etc.)

### DIFF
--- a/pkg/store/consul.go
+++ b/pkg/store/consul.go
@@ -125,6 +125,7 @@ func (s *Consul) normalize(key string) string {
 // to use in conjunction to CAS calls
 func (s *Consul) Get(key string) (*KVPair, error) {
 	options := &api.QueryOptions{
+		AllowStale:        false,
 		RequireConsistent: true,
 	}
 	pair, meta, err := s.client.KV().Get(s.normalize(key), options)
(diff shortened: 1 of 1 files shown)


abronan:
ping @mavenugo 

Would be cool if you can confirm that this is working for you until we have test coverage on this part of the lib :+1:

mavenugo:
+1 for options to choose consistency configuration. I will give it a spin and let you know.
Thanks for the quick turnaround.

aluzzardi:
LGTM

PR #884, opened 2015-06-01
store: Fix ephemeral behavior with Consul

### LINKED ISSUE #830: Default settings for Node removal are not optimal
Default settings for TTLs and Node removal on storage backends are not optimal and two annoying things could happen:

- It takes time using Consul and Etcd to detect a node failure, if the value of the TTL is high, swarm will see those nodes as alive until those entries are expired.
- This other scenario can happen:
  - Swarm agent registers onto discovery service
  - Manager sees the new entry: it is added to the list of known agents
  - Swarm agent fails
  - Manager removes the agent after the TTL expires
  - Swarm agent is back to life at the same time or directly after
  - It may take the Manager `heartbeat time` to register that node again although it's alive and working

One solution could be to reduce the default `heartbeat` and `refresh` time to a low value but at the risk of an increased pressure on the metadata storage cluster, especially if we use it for both discovery and metadata.

abronan:
Fixes #830 

This was an error with Consul: When we join a cluster with an agent, then kill it and re-join, the key is still there with the old unexpired session. But as we create a new one for ephemeral nodes each time we relaunch `swarm join`, the TTL for the old session expires and the Manager is removing the entry from the cluster because of the behavior set to `Delete`.

You then need to wait `heartbeat` time for this agent to join back the cluster with the new, correct session.

The solution is pretty primitive as we delete the key first to remove any attached session for that key. It's a quick fix but we need to find a better way to handle this gracefully.

### DIFF
--- a/pkg/store/consul.go
+++ b/pkg/store/consul.go
@@ -97,9 +97,21 @@ func (s *Consul) setEphemeralTTL(ttl time.Duration) error {
 
 // CreateEphemeralSession creates the a global session
 // once that is used to delete keys at node failure
-func (s *Consul) createEphemeralSession() error {
+func (s *Consul) createEphemeralSession(key string) error {
 	s.Lock()
 	defer s.Unlock()
+
+	// Recover existing global session
+	pair, _, err := s.client.KV().Get(key, nil)
+	if err != nil {
+		return err
+	}
+	if pair != nil && pair.Session != "" {
+		s.ephemeralSession = pair.Session
+		return nil
+	}
+
+	// Create new session
 	if s.ephemeralSession == "" {
 		entry := &api.SessionEntry{
 			Behavior: api.SessionBehaviorDelete,
@@ -152,7 +164,10 @@ func (s *Consul) Put(key string, value []byte, opts *WriteOptions) error {
 		// Creates the global ephemeral session
 		// if it does not exist
 		if s.ephemeralSession == "" {
-			s.createEphemeralSession()
+			err := s.createEphemeralSession(key)
+			if err != nil {
+				return err
+			}
 		}
 
 		// Create lock option with the
--- a/pkg/store/consul.go
+++ b/pkg/store/consul.go
@@ -95,21 +95,11 @@ func (s *Consul) setEphemeralTTL(ttl time.Duration) error {
 	return nil
 }
 
-// CreateEphemeralSession creates the a global session
+// createEphemeralSession creates the global session
 // once that is used to delete keys at node failure
-func (s *Consul) createEphemeralSession(key string) error {
+func (s *Consul) createEphemeralSession() error {
 	s.Lock()
 	defer s.Unlock()
-
-	// Recover existing global session
-	pair, _, err := s.client.KV().Get(key, nil)
-	if err != nil {
-		return err
-	}
-	if pair != nil && pair.Session != "" {
-		s.ephemeralSession = pair.Session
-		return nil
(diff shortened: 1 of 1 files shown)


vieux:
ping @aluzzardi

pkg/store/consul.go
```diff
@@ -97,9 +97,14 @@ func (s *Consul) setEphemeralTTL(ttl time.Duration) error {
...
 // CreateEphemeralSession creates the a global session
 // once that is used to delete keys at node failure
-func (s *Consul) createEphemeralSession() error {
+func (s *Consul) createEphemeralSession(key string) error {
 	s.Lock()
 	defer s.Unlock()
+
+	// Delete the key first (erase any attached session for that key)
+	// FIXME recover gracefully with the existing global session
+	s.Delete(key)
```
aluzzardi: uhm, this will delete **every time**, not just the first time we create the ephemeral session - not sure if we could do it differently

abronan: It deletes it a single time although it does not work as expected: each subsequent key going through the Put will stay with the old session (it thus only work truly with a single key, which is bad). A better solution is to get back the existing session but at the cost of the Manager not seeing the "change" of the Agent leaving and coming back to life.

I will keep this PR open and give it a shot.

abronan:
@aluzzardi PTAL now the session is recovered instead of destroyed first. The agent just keeps refreshing the existing session `ttl` instead of a new one. Even though now the Manager does not see the failure (but it will mark the node as unhealthy anyway so it's not a problem).

abronan:
Ok not good to review, still edge cases. **Do not merge** until this is fixed and there is proper integration test in #887.

pkg/store/consul.go
```diff
@@ -95,11 +95,13 @@ func (s *Consul) setEphemeralTTL(ttl time.Duration) error {
 	return nil
 }
 
-// CreateEphemeralSession creates the a global session
+// CreateEphemeralSession creates the global session
```
aluzzardi: `createEphemeralSession creates [...]`

pkg/store/consul.go
```diff
@@ -108,11 +110,23 @@ func (s *Consul) createEphemeralSession() error {
...
+			return "", err
 		}
-		s.ephemeralSession = session
+		return session, nil
 	}
-	return nil
+	return "", nil
+}
+
+// checkPreviousSession checks if the key already has a session attached
```
aluzzardi: `checkActiveSession checks...`

pkg/store/consul.go
```diff
@@ -145,17 +159,32 @@ func (s *Consul) Put(key string, value []byte, opts *WriteOptions) error {
...
 		if s.ephemeralSession == "" {
-			s.createEphemeralSession()
+			if s.ephemeralSession, err = s.createEphemeralSession(); err != nil {
+				return err
+			}
+		}
+
+		// If a previous session is still active for that key, use it
+		// else we use the global ephemeral session
+		if previous != "" && previous != s.ephemeralSession {
```
aluzzardi: there's probably no need to check `previous != s.ephemeralSession`

pkg/store/consul.go
vieux: @aluzzardi I agree for consistency, but this method is unexported. So golint won't complain

abronan:
@aluzzardi PTAL

aluzzardi:
LGTM

PR #885, opened 2015-06-01
use inspect sent by mesos 0.23

vieux:
ping @tnachen @jimenez

### DIFF
--- a/cluster/mesos/cluster.go
+++ b/cluster/mesos/cluster.go
@@ -2,6 +2,7 @@ package mesos
 
 import (
 	"crypto/tls"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -397,7 +398,7 @@ func (c *Cluster) scheduleTask(t *task) bool {
 	}
 	c.Unlock()
 	// block until we get the container
-	finished, err := t.monitor()
+	finished, data, err := t.monitor()
 	taskID := t.TaskInfo.TaskId.GetValue()
 	if err != nil {
 		//remove task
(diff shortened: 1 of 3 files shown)


vieux:
ping @aluzzardi

jimenez:
LGTM

cluster/mesos/cluster.go
```diff
@@ -423,18 +424,30 @@ func (c *Cluster) scheduleTask(t *task) bool {
...
 
 	// Register the container immediately while waiting for a state refresh.
-	// Force a state refresh to pick up the newly created container.
-	// FIXME: unexport this method, see FIXME in engine.go
+
+	// In mesos 0.23+ the docker inspect will be sent back in the taskStatus.Data
+	// We can use this to find the right container.
+	inspect := []dockerclient.ContainerInfo{}
+	if data != nil && json.Unmarshal(data, &inspect) != nil && len(inspect) == 1 {
+		s.engine.RefreshContainer(inspect[0].Id, true)
```
aluzzardi: How would this look like if we managed to build a `Container` from that info and then did a `container.Refresh()`?

In that case we would not need to change the function to public

vieux: I can try, but it will add a lot of code as creating a `dockerclient.Container` from a `dockerclient.ContainerInfo` wil be done field by field

aluzzardi: You only need to set the ID, right?

vieux: oh, right, only Id and engine (we have it)

vieux:
I would love to wait for https://github.com/docker/swarm/pull/894 to be merged before.

PR #886, opened 2015-06-01
fix resources in docker 1.7

vieux:


vieux:
@abronan @aluzzardi if you see a better way to do `consolidateResourceFields` please tell me

abronan:
Not really :-/ LGTM

cluster/config.go
```diff
@@ -24,6 +24,40 @@ func parseEnv(e string) (bool, string, string) {
 	return false, "", ""
 }
 
+func consolidateResourceFields(c *dockerclient.ContainerConfig) {
```
aluzzardi:
```
// FIXME: Temporary fix to handle forward/backward compatibility between Docker <1.6 and >=1.7
// ContainerConfig should be handling converting to/from different docker versions
```

aluzzardi:
The better way would be to handle converting one way using `BuildContainerConfig(<original docker config>, <api version>)` and the other way using `config.Serialize(<api version>)`, or something like that.

In the meantime, this does the trick but please add a FIXME as suggested.

LGTM

vieux:
@aluzzardi done I also added a test, so the coverage doesn't decrease

PR #887, opened 2015-06-02
store: Add unit testing for pkg/store on Travis

abronan:
First shot at unit testing `pkg/store` on Travis.

Include tests for `Put`/`Get`/`Delete`/`AtomicPut`/`Lock`/`Unlock` for the three backends.

The downside is that tests will now fail locally if you don't have the store available and running :-/


abronan:
ping @aluzzardi @mavenugo @vieux 

Should improve that coverage...

abronan:
Depends on: #896 #895 #884 #881

aluzzardi:
So the tests are currently replicated 3 times.

It would be best to use fixtures. This is an option to do it:

- In `store_test.go`:

1) Put all the test functions currently defined in `{consul,etcd,zookeeper}_test.go` so that:
`TestConsulPutGetDelete(*testing.T)` becomes `testStorePutGetDelete(*testing.T, store.Store)`

It should be exactly the same code, except that the function takes a store instead of creating one. Also, it's private rather than being public (so our fixtures don't get called by go test).

2) Add a `testStore(*testing.T, store.Store)` helper that invokes all the sub tests (`testStorePutGetDelete`, `testStoreWatch`, ...)

- In `{consul,etcd,zookeeper}_test.go`:

Have a single simple test, such as `TestConsul` that just creates the store and calls `testStore(t, kv)`

abronan:
@aluzzardi There are a few `FIXME`'s left, like on key formatting and watch management. But when those gets fixed we can basically merge all those tests into `store_test.go` and remove all the duplicated code. I did this mainly to pass those tests first and fix bugs as I encounter them on the three stores. The goal will be indeed to converge and have the same test file for each backend (to prove that the behavior stays the same no matter the backend you want to use).

aluzzardi:
LGTM

PR #892, opened 2015-06-02
add images to docker info

vieux:
docker1.7-rc1 info against a swarm cluster is kinda broken,
https://github.com/docker/docker/pull/13607 will fix most of the issue for docker1.7-rc2

Although, the `Images:` field isn't optional anymore so we need tu return it.


vieux:
ping @abronan

abronan:
LGTM

PR #894, opened 2015-06-03
Add mesos in the CI

vieux:

### DIFF
--- a/test/integration/helpers.bash
+++ b/test/integration/helpers.bash
@@ -165,7 +165,8 @@ function start_docker() {
 		# We have to manually call `hostname` since --hostname and --net cannot
 		# be used together.
 		DOCKER_CONTAINERS[$i]=$(
-			docker_host run -d --name node-$i --privileged -it --net=host \
+			# -v /usr/local/bin -v /var/run/docker.sock are specific to mesos, so the slave can do a --volumes-from and use the docker cli
+			docker_host run -d --name node-$i --privileged -v /usr/local/bin -v /var/run/docker.sock -it --net=host \
 			${DOCKER_IMAGE}:${DOCKER_VERSION} \
 			bash -c "\
 				hostname node-$i && \
--- a/test/integration/mesos/mesos_helpers.bash
+++ b/test/integration/mesos/mesos_helpers.bash
@@ -2,11 +2,40 @@
 
 load ../helpers
 
-MESOS_CLUSTER_ENTRYPOINT=${MESOS_CLUSTER_ENTRYPOINT:-0.0.0.0:5050}
+MESOS_IMAGE=jimenez/mesos-dev:clang
+MESOS_MASTER_PORT=$(( ( RANDOM % 1000 )  + 10000 ))
+BASE_PORT=2375
+
+# Start mesos master and slave.
+function start_mesos() {
+	MESOS_MASTER=$(
+	docker_host run -d --name mesos-master --net=host \
+	$MESOS_IMAGE /mesos/build/bin/mesos-master.sh --ip=127.0.0.1 --work_dir=/ --registry=in_memory --port=$MESOS_MASTER_PORT
+		)
+
+	retry 10 1 eval "docker_host ps | grep 'mesos-master'"
+
+	MESOS_SLAVE=$(
+	docker_host run --privileged -d --name mesos-slave --volumes-from node-0 -e DOCKER_HOST="${HOSTS[0]}" -v /sys/fs/cgroup:/sys/fs/cgroup --net=host \
+	$MESOS_IMAGE /mesos/build/bin/mesos-slave.sh --master=127.0.0.1:$MESOS_MASTER_PORT --containerizers=docker --hostname=127.0.0.1 --port=$(($MESOS_MASTER_PORT + 1))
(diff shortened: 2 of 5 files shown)


test/integration/helpers.bash
```diff
@@ -165,7 +165,7 @@ function start_docker() {
 		# We have to manually call `hostname` since --hostname and --net cannot
 		# be used together.
 		DOCKER_CONTAINERS[$i]=$(
-			docker_host run -d --name node-$i --privileged -it --net=host \
+			docker_host run -d --name node-$i --privileged -v /usr/local/bin -v /var/run/docker.sock -it --net=host \
```
vieux: we do a `--volume-from` in the mesos tests.

chanwit:
This is super!!!

vieux:
:fireworks: :cool: the CI is green

@tnachen @jimenez

jimenez:
:+1:

tnachen:
Nice! I'm watching the results to see if CI passes

vieux:
@tnachen it will :D it was green a minute ago, I had to do a force push

tnachen:
@vieux Cool, is the mesos-master/mesos-slave built with the latest master (0.23)?

vieux:
@tnachen yes, I used @jimenez's image that was build today with master (0.23)

test/integration/mesos/mesos_helpers.bash
```diff
@@ -2,11 +2,36 @@
 
 load ../helpers
 
-MESOS_CLUSTER_ENTRYPOINT=${MESOS_CLUSTER_ENTRYPOINT:-0.0.0.0:5050}
+MESOS_IMAGE=jimenez/mesos-dev:clang
+MESOS_MASTER_PORT=$(( ( RANDOM % 1000 )  + 7000 ))
```
aluzzardi: different port please (7000 is already taken)

test/integration/mesos/mesos_helpers.bash
```diff
@@ -2,11 +2,36 @@
...
+
+# Start mesos master and slave.
+function start_mesos() {
+	MESOS_MASTER=$(
+	docker_host run -d --name mesos-master --net=host \
+	$MESOS_IMAGE /mesos/build/bin/mesos-master.sh --ip=127.0.0.1 --work_dir=/ --registry=in_memory --port=$MESOS_MASTER_PORT
+		)
+
+	MESOS_SLAVE=$(
+	docker_host run -d --name mesos-slave --volumes-from node-0 -e DOCKER_HOST="${HOSTS[0]}" -v /sys/fs/cgroup:/sys/fs/cgroup --net=host \
```
tnachen: You should also run --pid=host, and containerizers=docker

tnachen: Also one thing I usually do is to mount in the sandbox volume from the host, so like -v /tmp/mesos:/tmp/mesos
Doing this we can also try out restarting the slave and it should recover all the docker tasks with swarm, etc.

test/integration/mesos/mesos_helpers.bash
```diff
@@ -2,11 +2,36 @@
...
+	$MESOS_IMAGE /mesos/build/bin/mesos-slave.sh --master=127.0.0.1:$MESOS_MASTER_PORT --containerizers=docker --hostname=127.0.0.1 --port=$(($MESOS_MASTER_PORT + 1))
```
tnachen: I just added some new flags that you should definitely use for running mesos-slave in a docker contianer.
You should at least use --docker-mesos-image=$MESOS_IMAGE
This will create all the executors in docker containers, and you can recover all the tasks on slave restart.

test/integration/mesos/mesos_helpers.bash
vieux: I agree on the sandbox, can you tell me more about the `--pid=host`, why is this needed ?

test/integration/mesos/mesos_helpers.bash
vieux: I don't think it will work, the mesos-slave can't access the image, here is the current setup:

```
jenkins 
 └ docker (here the image jimenez/mesos-dev:clang is pulled)
   └ docker (there the image is not)
   └ mesos-slave 
   └ mesos-master
```

test/integration/mesos/mesos_helpers.bash
tnachen: Mesos will try to find your cgroup settings based on the pid of the docker container you launched, so without --pid=host DockerContainerizer can't see the Docker container that you launched unless you're also using docker in docker where docker daemon is in the same image.

vieux: As `--pid=host` was added recently, I added`--privileged` so it will work with older version.

test/integration/mesos/mesos_helpers.bash
vieux: Let's add those new flags when we need them, it will require some changes in the tests: if a test does
```
docker run busybox ls
docker images
```

I'm expecting to see only one image (`busybox`), not two (`busybox`, `jimenez/mesos-dev:clang`)

PR #895, opened 2015-06-03
store: Fix zookeeper List call, implement DeleteTree

abronan:
`List` in Zookeeper was not returning the correct arguments for `KVEntry` because a call to `Children` returns a list of keys and not values.

This PR also implements `DeleteTree` (that was not working before) using `List`.

This is the first of a series of PRs to achieve 80%+ code coverage in `pkg/store`. (using unit tests on travis that is WIP)

### DIFF
--- a/pkg/store/zookeeper.go
+++ b/pkg/store/zookeeper.go
@@ -194,21 +194,36 @@ func (s *Zookeeper) WatchTree(prefix string, stopCh <-chan struct{}) (<-chan []*
 
 // List the content of a given prefix
 func (s *Zookeeper) List(prefix string) ([]*KVPair, error) {
-	prefix = normalize(prefix)
-	entries, stat, err := s.client.Children(prefix)
+	keys, stat, err := s.client.Children(normalize(prefix))
 	if err != nil {
 		return nil, err
 	}
 	kv := []*KVPair{}
-	for _, item := range entries {
-		kv = append(kv, &KVPair{prefix, []byte(item), uint64(stat.Mzxid)})
+	for _, key := range keys {
+		// FIXME Costly Get request for each child key..
+		pair, err := s.Get(prefix + normalize(key))
+		if err != nil {
+			return nil, err
+		}
+		kv = append(kv, &KVPair{key, []byte(pair.Value), uint64(stat.Version)})
 	}
-	return kv, err
+	return kv, nil
 }
 
 // DeleteTree deletes a range of keys based on prefix
 func (s *Zookeeper) DeleteTree(prefix string) error {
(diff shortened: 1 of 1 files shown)


abronan:
ping @aluzzardi @vieux

pkg/store/zookeeper.go
```diff
@@ -194,21 +194,37 @@ func (s *Zookeeper) WatchTree(prefix string, stopCh <-chan struct{}) (<-chan []*
 
 // List the content of a given prefix
 func (s *Zookeeper) List(prefix string) ([]*KVPair, error) {
-	prefix = normalize(prefix)
-	entries, stat, err := s.client.Children(prefix)
+	fprefix := normalize(prefix)
+	keys, stat, err := s.client.Children(fprefix)
```
chanwit: `fprefix` only used once so it could be inlined?

abronan: @chanwit +1 thanks!

pkg/store/zookeeper.go
```diff
@@ -194,21 +194,36 @@ func (s *Zookeeper) WatchTree(prefix string, stopCh <-chan struct{}) (<-chan []*
...
+		kv = append(kv, &KVPair{key, []byte(pair.Value), uint64(stat.Version)})
 	}
-	return kv, err
+	return kv, nil
 }
 
 // DeleteTree deletes a range of keys based on prefix
 func (s *Zookeeper) DeleteTree(prefix string) error {
-	err := s.client.Delete(normalize(prefix), -1)
+	pairs, err := s.List(prefix)
```
aluzzardi: Am I right to assume that zookeeper itself doesn't provide such a feature?

abronan: Zookeeper returns a list of keys with `Children` (and not pairs like etcd or consul) and then you can do whatever you want if those keys (get, delete, etc). I think the `List` is meant to be an upper level construct that you implement yourself.

Unfortunately I can't seem to find any way to avoid the `Get` operation for each key. I wanted to use the `Multi` (see below) but there is no `GetRequest` construct in go-zookeeper.

chanwit:
For `Multi` `GetDataRequest`, reported here in samuel\go-zookeeper#59. I think it's doable.

aluzzardi:
Awesome - thanks @chanwit.

How about we merge this (since it's plain broken right now) and open an issue to fix it once samuel/go-zookeeper#59 gets addressed?

aluzzardi:
LGTM

chanwit:
@abronan @aluzzardi I put a further tracking in #899.

PR #896, opened 2015-06-03
store: Fix Zookeeper single Key watch

abronan:
Fix Zookeeper single key `Watch` the same way than `WatchTree`.

ping @aluzzardi @vieux 

### DIFF
--- a/pkg/store/zookeeper.go
+++ b/pkg/store/zookeeper.go
@@ -121,7 +121,7 @@ func (s *Zookeeper) Exists(key string) (bool, error) {
 // Providing a non-nil stopCh can be used to stop watching.
 func (s *Zookeeper) Watch(key string, stopCh <-chan struct{}) (<-chan *KVPair, error) {
 	fkey := normalize(key)
-	resp, meta, eventCh, err := s.client.GetW(fkey)
+	pair, err := s.Get(key)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


vieux:
LGTM

pkg/store/zookeeper.go
```diff
@@ -131,9 +131,13 @@ func (s *Zookeeper) Watch(key string, stopCh <-chan struct{}) (<-chan *KVPair, e
 	go func() {
 		defer close(watchCh)
 
-		// GetW returns the current value before setting the watch.
-		watchCh <- &KVPair{key, resp, uint64(meta.Version)}
+		// Get returns the current value before setting the watch.
```
aluzzardi: This is not true anymore, right?

pkg/store/zookeeper.go
```diff
@@ -131,9 +131,13 @@ func (s *Zookeeper) Watch(key string, stopCh <-chan struct{}) (<-chan *KVPair, e
...
+		watchCh <- pair
 		for {
+			_, _, eventCh, err := s.client.GetW(fkey)
```
aluzzardi: How does `GetW` work?

aluzzardi: Meaning: it doesn't return the current value, only a watch right?

abronan: It's a one time watch thus the necessity to initialize it at each turn (If you want to keep watching events on that key).

pkg/store/zookeeper.go
abronan: It is, we still send back the first value to the channel but this time using a simple `Get` ;)

aluzzardi: Yep, but it doesn't:

> before setting the watch

abronan: Oh yeah I see ;) I will change the comment.

aluzzardi:
LGTM

PR #897, opened 2015-06-04
fix panic is disconnect

vieux:
this PR fix a bug in the disconnect that can lead to a panic (I can reproduce easily)

If we close the `stopCh` while the `refreshLoop` goroutine is not in the `select` it's going to panic, because we set the engine to nil and the refreshImages/refreshContainers uses the engine.

We need to wait on the goroutine the finish before setting the client to nil.

close #836

PR #898, opened 2015-06-04
disconnect from engines in mesos

vieux:
depends on #897


abronan:
@vieux Any status on this? :)

vieux:
Closing for now

Issue #888, opened 2015-06-02
Resource management integration tests


aluzzardi:
Docker 1.7 broke our resource management (`Memory` and `CPU` are now in `HostConfig` rather than `Config`).

This should not happen again - we need to add integration tests to verify everything is working correctly.

Basic tests:
- [x] **Resource Propagation**: `docker run -m X -c Y foo` and check if data is shares are propagated correctly (`docker inspect` again swarm and docker host). **Note**: We do weird stuff with CPU shares. We should verify it works correctly.

- [x] **Resource Limits**: `docker run -m BIG` and `docker run -c BIG` should fail to schedule (`not enough resources`)

Strategies: Perhaps, later, we could add some tests that actually try out the strategy, similar to `script/demo/misc/script.sh`:
- [ ] **binpack**
- [ ] **spread**

chanwit:
@aluzzardi could you please make them into a check list, so I can help fixing some of them. I see 4 items here.

vieux:
@aluzzardi I quickly created https://github.com/docker/swarm/pull/886 to track the issue in the engine.

aluzzardi:
@chanwit, sure!

chanwit:
I don't understand what resource propagation is, so I skip it :)
But I think I could handle `binpack` and `spread` tests.

PR #901, opened 2015-06-04
integration tests for resource limitation

chanwit:
Add resource limitation tests.
Subtask of #888 

### DIFF
--- a/test/integration/resource_management.bats
+++ b/test/integration/resource_management.bats
@@ -1,0 +1,44 @@
+#!/usr/bin/env bats
+
+load helpers
+
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+}
+
+@test "resource limitation: memory" {
+	start_docker_with_busybox 1
+	swarm_manage
+
+	run docker_swarm run -m 1000G busybox sh
+	[ "$status" -ne 0 ]
+	[[ "${lines[0]}" == *"no resources available to schedule container"* ]]
+
+	# The number of running containers should be still 0.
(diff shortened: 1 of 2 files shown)


chanwit:
@aluzzardi PTAL

test/integration/resource_management.bats
```diff
@@ -0,0 +1,34 @@
...
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+}
+
+@test "resource limitation: memory" {
+	start_docker_with_busybox 1
+	swarm_manage
+
+	run docker_swarm run -m 48G busybox sh
```
aluzzardi: Can you put something **crazy** ? :) Like 1000G?

aluzzardi:
I think @vieux added a similar test in `api/run.bats`.

This one is better - could you remove the other one?

vieux:
@chanwit can you please also move `@test "docker run with resources"` in this file and add a `docker info` after the test to make sure the docker info reports the right message.

chanwit:
@vieux @aluzzardi PTAL

It seems `janky` failed because its clean up process (`docker rmi` called without a parameter).

aluzzardi:
@chanwit I believe it's not because of the `rmi` :/

>    docker run87/87Build timed out (after 60 minutes). Marking the build as failed.

It looks like it's timing out. I'm relaunching this to see if it was transient

aluzzardi:
Okay it worked

LGTM

chanwit:
phew

PR #902, opened 2015-06-04
Improve docker info against 1.7

vieux:


vieux:
ping @aluzzardi @abronan i'd like to include this one in the rc2

api/handlers.go
```diff
@@ -44,16 +44,9 @@ func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
...
-		APIVersion string `json:"ApiVersion"`
-		GoVersion  string
-		GitCommit  string
-		Os         string
-		Arch       string
-	}{
+	version := dockerclient.Version{
 		Version:    "swarm/" + version.VERSION,
-		APIVersion: APIVERSION,
+		ApiVersion: APIVERSION,
```
aluzzardi: Are you sure about this one? I don't remember which one we should use

cluster/cluster.go
```diff
@@ -51,7 +51,7 @@ type Cluster interface {
 
 	// Return some info about the cluster, like nb or containers / images
 	// It is pretty open, so the implementation decides what to return.
-	Info() [][2]string
+	Info() ([][]string, int64, int64)
```
aluzzardi: Nit (feel totally free to ignore it):

It might be cleaner to have `cluster.Memory()`

WDYT?

vieux: I agree, and `cluster.CPUs()` it forces us to add 2 new method to the interface.

api/handlers.go
vieux: This one is good.

vieux:
@aluzzardi PTAL

aluzzardi:
LGTM

PR #916, opened 2015-06-09
cannot comment on one line command

affo:
You cannot do something like:

```
-v /sys/fs/cgroup:/sys/fs/cgroup:ro \ # this is specific to mesos
```

Because `\` will be taken as excape and so ` #` as image for the run command. So:

```
Invalid repository name ( #), only [a-z0-9-_.] are allowed
```

### DIFF
--- a/test/integration/run.sh
+++ b/test/integration/run.sh
@@ -14,7 +14,7 @@ INTEGRATION_IMAGE=${INTEGRATION_IMAGE:-dockerswarm/swarm-test-env}
 
 # Start the integration tests in a Docker container.
 ID=$(docker run -d -t --privileged \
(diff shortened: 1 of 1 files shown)


affo:
How is this change supposed to make the build fail?!

vieux:
@affo how od you run the tests ? I have not issue on my machine.

affo:
With a bash shell:

```
$ test/integration/run.sh
```
Il 09/giu/2015 21:00, "Victor Vieux" <notifications@github.com> ha scritto:

vieux:
@affo ok, can you put the comment somewhere else ?

affo:
I found only this way.  
It is surprising that it works for you!

vieux:
LGTM

vieux:
ping @aluzzardi

Issue #836, opened 2015-05-23
Node removal makes the manager crash


aluzzardi:
It looks like the events callback gets called anyway even though we unregister it when disconnecting the engine:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x60 pc=0x530f2b]

goroutine 54 [running]:
runtime.panic(0x829780, 0xad6b53)
	/usr/local/go/src/pkg/runtime/panic.c:279 +0xf5
github.com/docker/swarm/cluster.(*Engine).refreshContainer(0xc2087c91e0, 0xc2088c0440, 0x40, 0x0, 0x0, 0x0)
	/root/go/src/github.com/docker/swarm/cluster/engine.go:210 +0x1bb
github.com/docker/swarm/cluster.(*Engine).handler(0xc2087c91e0, 0xc2088c0400, 0x0, 0x0, 0x0, 0x0)
	/root/go/src/github.com/docker/swarm/cluster/engine.go:507 +0x227
github.com/docker/swarm/cluster.*Engine.(github.com/docker/swarm/cluster.handler)·fm(0xc2088c0400, 0x0, 0x0, 0x0, 0x0)
	/root/go/src/github.com/docker/swarm/cluster/engine.go:105 +0x58
github.com/samalba/dockerclient.(*DockerClient).getEvents(0xc2086b96c0, 0xc2087187c0, 0x0, 0x0, 0x0, 0x0)
	/root/go/src/github.com/samalba/dockerclient/dockerclient.go:295 +0x3e0
created by github.com/samalba/dockerclient.(*DockerClient).StartMonitorEvents
	/root/go/src/github.com/samalba/dockerclient/dockerclient.go:276 +0x83

goroutine 16 [chan receive, 39 minutes]:
github.com/docker/swarm/api.(*Server).ListenAndServe(0xc20872c120, 0x0, 0x0)
	/root/go/src/github.com/docker/swarm/api/server.go:113 +0x3c0
github.com/docker/swarm/cli.manage(0xc208044180)
	/root/go/src/github.com/docker/swarm/cli/manage.go:205 +0x1144
github.com/codegangsta/cli.Command.Run(0x889920, 0x6, 0x889880, 0x1, 0x8b7e50, 0x17, 0x0, 0x0, 0x0, 0x0, ...)
	/root/go/src/github.com/codegangsta/cli/command.go:113 +0xeb2
[...]

PR #932, opened 2015-06-11
fix the panic when docker event emitted after engine removal (#836)

mountkin:
I think #836  should be reopen because #897 didn't really fix the panic.
The problem can still be reproduced by
```
docker run -itd --name=test busybox sh
echo 127.0.0.1:2375 >/tmp/discovery.txt
./swarm m -H :8888 file:///tmp/discovery.txt &
sleep 5
echo >/tmp/discovery.txt
sleep 60
docker rm -f test
```
 
This panic is caused by the event monitor. When calling `StopAllMonitorEvents`, dockerclient just set `client.monitorEvents` to 0, but this doesn't prevent the json decoder from decoding new events.
When new events are emitted after engine removal, the event will be passed to `engine.handler` as usual. But at this time the `e.client` is set to nil by `engine.Disconnect`, so the panic happens.

This PR leverages the `MonitorEvents` API of dockerclient, which accepts a `stopChan`. When the stopChan is readable, new events won't be returned to the calling procedure any more. 

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -120,27 +120,28 @@ func (e *Engine) startMonitorEvents() {
 			"engine_id": e.ID,
 			"engine_ip": e.IP,
 		}).Warning("Start event monitor failed")
-		e.healthy = false
 		return
 	}
-	for {
-		select {
-		case ev := <-ch:
-			if ev.Error == nil {
-				go e.handler(&ev.Event)
-			} else {
-				log.WithFields(log.Fields{
-					"engine_id": e.ID,
-					"engine_ip": e.IP,
-				}).Warning("Receive event failed: ", ev.Error)
-				// Perhaps we can set e.healthy = false here?
+	go func() {
+		for {
+			select {
+			case ev := <-ch:
+				if ev.Error == nil {
+					e.handler(&ev.Event)
+				} else {
+					log.WithFields(log.Fields{
+						"engine_id": e.ID,
+						"engine_ip": e.IP,
+					}).Warning("Receive event failed: ", ev.Error)
+					// Perhaps we can set e.healthy = false here?
+					return
+				}
+			case <-e.stopMonitorEventsCh:
+				stopCh <- struct{}{}
 				return
 			}
-		case <-e.stopMonitorEventsCh:
-			stopCh <- struct{}{}
-			return
 		}
-	}
+	}()
 }
 
 func (e *Engine) stopMonitorEvents() {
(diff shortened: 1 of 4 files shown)


mountkin:
I think we can use the MonitorEvents API to implement a realtime engine health monitor. I'll give it a shot this weekend.

I'll fix the broken test tomorrow.

vieux:
@mountkin thank we discovered this bug a few days ago as well. I think @ehazlett is working on a fix in dockerclient directly ?

aluzzardi:
@mountkin Thanks!

A few things:
- I believe this is the right way to do it
- As @vieux mentioned, @ehazlett is working on dockerclient for `StartMonitorEvents`. This is a good short term solution

Basically, we are in the process of releasing `0.3` so we are not comfortable merging this in right before the release (in a few days) since it's an essential part of how we do cluster updates.

However, once we do the final cut, I believe we should merge this.

In the meantime - if you check travis there's a problem with data rces.

cluster/engine.go
```diff
@@ -104,18 +106,53 @@ func (e *Engine) ConnectWithClient(client dockerclient.Client) error {
...
+			} else {
+				log.WithFields(log.Fields{
+					"engine_id": e.ID,
+					"engine_ip": e.IP,
+				}).Warning("Receive event failed: ", ev.Error)
+				// Perhaps we can set e.healthy = false here?
+				return
+			}
+		case <-e.stopMonitorEventsCh:
+			stopCh <- struct{}{}
```
aluzzardi: Problem here:

`Disconnect` sends a message to `stopCh` then to `stopMonitorEventsCh` - here you are sending another message to `stopCh` and this may block forever.

This code is:
- Sending a message to `stopCh`
- Calling `stopMonitorEvents`, which sends a message to `stopCh`

cluster/engine.go
```diff
@@ -104,18 +106,53 @@ func (e *Engine) ConnectWithClient(client dockerclient.Client) error {
...
+	e.stopMonitorEventsCh <- struct{}{}
+}
+
 // Disconnect will stop all monitoring of the engine.
 // The Engine object cannot be further used without reconnecting it first.
 func (e *Engine) Disconnect() {
 	// do not close the chan, so it wait until the refreshLoop goroutine stops
 	e.stopCh <- struct{}{}
-	e.client.StopAllMonitorEvents()
+	e.stopMonitorEvents()
```
aluzzardi: I think we should stop monitoring events before stopping the refresh loop

WDYT?

vieux:
See https://github.com/samalba/dockerclient/pull/117

cluster/engine.go
vieux: I agree.

cluster/engine.go
mountkin: This stopCh is a local variable. It's used for stopping the event monitor.

aluzzardi: Indeed! My bad

mountkin:
fixed in #941

PR #937, opened 2015-06-12
fix panic in disconnect with events

vieux:
related to #836

### DIFF
--- a/Godeps/_workspace/src/github.com/samalba/dockerclient/dockerclient.go
+++ b/Godeps/_workspace/src/github.com/samalba/dockerclient/dockerclient.go
@@ -32,6 +32,7 @@ type DockerClient struct {
 	TLSConfig     *tls.Config
 	monitorEvents int32
 	monitorStats  int32
+	eventStopChan chan (struct{})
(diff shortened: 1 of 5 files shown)


vieux:
ping @aluzzardi

aluzzardi:
Relaunched jenkins 3 times. Fails all 3 times. Not sure if coincidence

Godeps/_workspace/src/github.com/samalba/dockerclient/dockerclient.go
```diff
@@ -354,8 +355,23 @@ func (client *DockerClient) MonitorEvents(options *MonitorEventsOptions, stopCha
...
-	go client.getEvents(cb, ec, args...)
+	client.eventStopChan = make(chan struct{})
+
+	eventErrChan, err := client.MonitorEvents(nil, client.eventStopChan)
+	if err != nil {
+		ec <- err
+		return
+	}
+
+	for {
```
mountkin: This may cause the call to `StartMonitorEvents` never returns.

PR #947, opened 2015-06-14
New documentation tooling for Hugo work

moxiegirl:
Updating title and other things
Removing old mkdocs file
Updating with the new sed scripts
Moving from index.md swarm-overview.md

ping @vieux @aluzzardi @abronan this the new docs tooling for 1.7.  Can I get an LGTM from one or more of you?


vieux:
LGTM

moxiegirl:
Thanks @vieux --- just as an FYI

### The "docs RC plan"

This is the first step to an "docs RC build" if you will.  

After I pull this PR and the other 6, I'm going to do a build to a staging bucket on S3.  That build will include the new theme which is dropping around noon Monday. I'll page through every page on the RC build:

- look for layout issues
- checking all the links
- making sure images appear

Each problem will be noted and then I'll do a set of follow on PRs to fix issues.  That's why I'm wanting to pull this set today, it allows for an RC cycle.  And then, also, if necessary one more before the release this week.

abronan:
LGTM

ping @aluzzardi

aluzzardi:
LGTM

moxiegirl:
Thank you guys!

PR #954, opened 2015-06-15
Fix types for store.Backend list

ahmetalpbalkan:
Unless explicitly specified these consts will be of type `string`; not `Backend`: http://play.golang.org/p/AcP-8qFsii

Signed-off-by: Ahmet Alp Balkan

### DIFF
--- a/pkg/store/store.go
+++ b/pkg/store/store.go
@@ -15,11 +15,11 @@ const (
 	// MOCK backend
 	MOCK Backend = "mock"
 	// CONSUL backend
-	CONSUL = "consul"
+	CONSUL Backend = "consul"
 	// ETCD backend
-	ETCD = "etcd"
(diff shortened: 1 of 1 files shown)


ahmetalpbalkan:
Not sure why the tests are failing, I'm not familiar with bats. Feel free to take over the patch and submit with the right set of fixes.

vieux:
LGTM

abronan:
Hi @ahmetalpbalkan, thanks! Would you mind opening the PR in [docker/libkv](https://github.com/docker/libkv) instead? We are in the process of removing `pkg/store` and vendoring `libkv` inside swarm, so your changes in this PR would be lost.

ahmetalpbalkan:
@abronan opened https://github.com/docker/swarm/pull/954

PR #963, opened 2015-06-15
defer RUnlock in swarm/cluster#Pull()

ahmetalpbalkan:
I'm not sure about the correctness of this one. Actually the Pull mutates the nodes' states as well, so I'm not sure if this should be a reader lock either. Any ideas @vieux ?

Signed-off-by: Ahmet Alp Balkan

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -303,6 +303,7 @@ func (c *Cluster) Pull(name string, authConfig *dockerclient.AuthConfig, callbac
 	var wg sync.WaitGroup
 
 	c.RLock()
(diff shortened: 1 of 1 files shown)


aluzzardi:
The nodes independently maintain their own locks so we don't need to lock access when they are being mutated.

What we want to lock in the cluster is access to the list of nodes itself (`c.engines`).

The problem with this PR is that by deferring the lock release **after** `wg.Wait()`, then the entire cluster stays RLocked for the entire pull duration, which is very long, causing lock contention issues.

ahmetalpbalkan:
Got it now, thanks Andrea.

Issue #743, opened 2015-05-07
Proposal: docker build + docker create without image affinities


aluzzardi:
Right now, in order to build in the cluster and run a container one must:

```
docker build -t foo .
docker create -e affinity:image==foo foo
```

That's because `foo` ends up on one machine but the container using `foo` might end up on a different one since swarm has no idea that the image is local and cannot be pulled from the registry

How about this:

1/ Try a create. It fails with 404
2/ Try a pull. It fails with something like `image not found`
3/ Check if the image exists in the cluster.
4/ Retry with a soft-affinity (but **don't** store it in the `ContainerConfig`)
5/ If it still fails, then there is a problem with the image, abort
6/ If it works, celebrate

aluzzardi:
@vieux Thoughts?

vieux:
Great idea.

electrical:
Sounds a very good solution for the problem i'm having now :-)

marc-:
How about locate the image and replicate it across cluster (or copy to specific node on demand)?

coffenbacher:
This would be awesome for my workflow. Good idea, can't wait!

abronan:
closed by #971

PR #971, opened 2015-06-18
Reschedule with soft image affinity 

jimmyxian:
Implement : https://github.com/docker/swarm/issues/743

>1/ Try a create. It fails with 404
>2/ Try a pull. It fails with something like image not found
>3/ Check if the image exists in the cluster.
>4/ Retry with a soft-affinity (but don't store it in the ContainerConfig)
>5/ If it still fails, then there is a problem with the image, abort
>6/ If it works, celebrate

### DIFF
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -147,11 +147,11 @@ func (c *ContainerConfig) Constraints() []string {
 	return c.extractExprs("constraints")
 }
 
-// AddSoftImageAffinity
+// AddSoftImageAffinity to config
 func (c *ContainerConfig) AddSoftImageAffinity(image string) error {
-	affnities_json := c.Labels[SwarmLabelNamespace+".affinities"]
+	affnitiesJSON := c.Labels[SwarmLabelNamespace+".affinities"]
 	var affinities []string
-	json.Unmarshal([]byte(affnities_json), &affinities)
+	json.Unmarshal([]byte(affnitiesJSON), &affinities)
 	affinities = append(affinities, "image==~"+image)
 
 	labels, err := json.Marshal(affinities)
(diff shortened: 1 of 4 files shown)


jimmyxian:
ping @aluzzardi @vieux

cluster/config.go
```diff
@@ -146,3 +146,18 @@ func (c *ContainerConfig) Affinities() []string {
 func (c *ContainerConfig) Constraints() []string {
 	return c.extractExprs("constraints")
 }
+
+// AddSoftImageAffinity to config
+func (c *ContainerConfig) AddSoftImageAffinity(image string) error {
+	affnitiesJSON := c.Labels[SwarmLabelNamespace+".affinities"]
```
vieux: you can use `c.extractExprs("affinities")`

cluster/config.go
```diff
@@ -146,3 +146,18 @@ func (c *ContainerConfig) Affinities() []string {
 func (c *ContainerConfig) Constraints() []string {
 	return c.extractExprs("constraints")
 }
+
+// AddSoftImageAffinity to config
+func (c *ContainerConfig) AddSoftImageAffinity(image string) error {
```
aluzzardi: I would call this `AddAffinity` and pass `image==~<image>` from the caller.

cluster/config.go
jimmyxian: OK

cluster/config.go
jimmyxian: Yeah..I will fix it

jimmyxian:
ping @aluzzardi @vieux 
Review again, thanks

vieux:
LGTM

aluzzardi:
Amazing :)

LGTM

We probably need to do some manual testing, just in case.

/cc @abronan

abronan:
@vieux @aluzzardi Tested on my cluster, work like a charm :+1: 

LGTM

abronan:
Thanks @jimmyxian! :)

PR #979, opened 2015-06-19
Fixing the headers

moxiegirl:
This is a no-op change to the header.


moxiegirl:
LGTM

PR #987, opened 2015-06-22
Allow task creation timeout to be configurable with mesos.

tnachen:


tnachen:
Just realized this is a dup, closing for now.

Issue #606, opened 2015-04-09
panic while deleting container


phemmer:
I was deleting several hundred containers (`docker ps -aq | xargs docker rm -fv`) when I received the following panic.

```
ERRO[2019] Unable to update state of container "f52cb1ca5e6325c3c81ca4db3e15cb0bf14e73cb22f8486b4d44ee73c674961e"  id=IOPS:R5SS:EWEQ:PWEI:BWYC:HZHH:5ZRA:CJS5:JG3P:3L6K:ERRE:TY6Z name=ip-10-179-134-220.ec2.internal
panic: assignment to entry in nil map

goroutine 50 [running]:
github.com/docker/swarm/cluster.(*Engine).updateContainer(0xc208158820, 0xc20877ba40, 0x40, 0xc20877ba80, 0x1, 0x4, 0xc20894f520, 0x17, 0xc20894f500, 0x1c, ...)
	/home/phemmer/.go/src/github.com/docker/swarm/cluster/engine.go:257 +0x272
github.com/docker/swarm/cluster.(*Engine).refreshContainers(0xc208158820, 0xc2086a4b00, 0x0, 0x0)
	/home/phemmer/.go/src/github.com/docker/swarm/cluster/engine.go:183 +0x20a
github.com/docker/swarm/cluster.(*Engine).refreshLoop(0xc208158820)
	/home/phemmer/.go/src/github.com/docker/swarm/cluster/engine.go:274 +0xab2
created by github.com/docker/swarm/cluster.(*Engine).connectClient
	/home/phemmer/.go/src/github.com/docker/swarm/cluster/engine.go:101 +0x1b3

goroutine 1 [chan receive, 33 minutes]:
github.com/docker/swarm/api.ListenAndServe(0x7f6c5d77c050, 0xc208077b30, 0xc20802afa0, 0x1, 0x1, 0x0, 0xc208001200, 0xc20805da70, 0x0, 0x0)
	/home/phemmer/.go/src/github.com/docker/swarm/api/server.go:83 +0x5dd
main.manage(0xc208086180)
	/home/phemmer/.go/src/github.com/docker/swarm/manage.go:137 +0xcb9
github.com/codegangsta/cli.Command.Run(0x831ab0, 0x6, 0x831a10, 0x1, 0x871870, 0x17, 0x0, 0x0, 0x0, 0x0, ...)
[...]

PR #993, opened 2015-06-26
606- Prevent panic if a container is removed during a refresh

ealang:
Hi, we ran into issue #606 and decided to investigate.  Here is a proposed change and unit test for the issue.

As stated in #606, this can be reproduced when containers are being removed.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -209,9 +209,11 @@ func (e *Engine) RefreshContainers(full bool) error {
 
 	merged := make(map[string]*Container)
 	for _, c := range containers {
-		merged, err = e.updateContainer(c, merged, full)
+		mergedUpdate, err := e.updateContainer(c, merged, full)
 		if err != nil {
-			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Errorf("Unable to update state of container %q", c.Id)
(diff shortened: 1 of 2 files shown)


abronan:
Thanks you for the contribution @ealang!

Jenkins fails for unrelated reasons and those tests must be fixed so LGTM.

ping @vieux @aluzzardi

cluster/engine.go
```diff
@@ -273,7 +273,7 @@ func (e *Engine) updateContainer(c dockerclient.Container, containers map[string
 	if full {
 		info, err := e.client.InspectContainer(c.Id)
 		if err != nil {
-			return nil, err
+			return containers, err
```
aluzzardi: I think we should fix the caller instead.

In Go, it is expected that if `err` is not nil then the rest of the response should be considered garbage.

aluzzardi:
Thanks @ealang for fixing this! This is a serious issue. I believe however that the fix should be different (comments inline)

ealang:
Thanks for the review.  PR updated.

aluzzardi:
Thanks!

One last comment: It would be useful to know why `updateContainer` failed in the first place (root cause).

Would you mind logging the error as well, something like:
```go
Errorf("Unable to update state of container %q: %v", c.Id, err)
```

aluzzardi:
LGTM - thanks!

aluzzardi:
Tests are failing but it's unrelated to this PR

PR #994, opened 2015-06-26
Rnote fix

moxiegirl:
- We can no longer maintain cumulative release notes; this is non-standard practice in any cases
- For older release notes, readers should view the older version of the docs


moxiegirl:
LGTM - this is an emergency fix. Will merge when completed.

PR #1004, opened 2015-06-30
fix api_version test

### LINKED ISSUE #1003: API version test
The docker engine changed the way it display version

```
Client:
 Version:      1.8.0-dev
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   77396bd-dirty
 Built:        Mon Jun 29 21:05:09 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      swarm/0.3.0
 API version:  1.16
 Go version:   go1.3.3
 Git commit:   7d2e4c5
 OS/Arch:      linux/amd64
```

The test greps for `client version` but it doesn't exists anymore.

vieux:

fix #1003

### DIFF
--- a/test/integration/api_version.bats
+++ b/test/integration/api_version.bats
@@ -12,16 +12,20 @@ function teardown() {
 # and your host CLI version differs from DOCKER_VERSION.
 @test "api version" {
 	start_docker 1
-	run docker -H "${HOSTS[0]}" version
+
+	# Get version output
+	out=$(docker -H "${HOSTS[0]}" version)	
+
+	# Check client version
+	run bash -c "echo '$out' | grep -i version | grep -v Go | grep -v API"
 	[ "$status" -eq 0 ]
 
-	# First line should contain the client version.
-	[[ "${lines[0]}" == "Client version: "* ]]
-	local cli_version=`echo "${lines[0]}" | cut -d':' -f2`
-	[[ "${output}" == *"Server version:$cli_version"* ]]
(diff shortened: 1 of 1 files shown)


vieux:
@aluzzardi what's the best way to make the tests fail instead of `false` ?

aluzzardi:
I'm worried this is too fragile and will blow up again since it's based on line count :/

How about instead just parsing `version:`" and making sure they are all the same? No matter the engine, there should be exactly 2 `*version:` and they should both have the same value after `:`

aluzzardi:
**Maybe** (pseudo-code)

docker info | grep -i version
$lines_count -eq 2
$(echo $lines[0] | cut -d':' -f2) == $(echo $lines[1] | cut -d':' -f2)

vieux:
@aluzzardi PTAL, here is the output so I had to use some greps

```
$ docker-1.7 version | grep -I version
Client version: 1.8.0-dev
Client API version: 1.20
Go version (client): go1.4.2
Server version: 1.8.0-dev
Server API version: 1.20
Go version (server): go1.4.2
```

```
$ docker-dev version | grep -i version
 Version:      1.8.0-dev
 API version:  1.20
 Go version:   go1.4.2
 Version:      1.8.0-dev
 API version:  1.20
 Go version:   go1.4.2
```

vieux:
ping @abronan

aluzzardi:
LGTM

abronan:
LGTM

PR #1008, opened 2015-06-30
Update libkv godeps and minor fixes

### LINKED ISSUE #980: Multi-Manager: Followers crashing when using etcd and stopping the store
Followers are crashing when using `etcd` and when we stop the store. This is due to a missing nil check in `etcd` backend. `etcd` may return nil for `result` when the store is not available.

abronan:
Updates `libkv` to the latest version.

fixes #980 

ping @vieux @aluzzardi


vieux:
LGTM

ping @aluzzardi

PR #1009, opened 2015-06-30
fix api/version.bats

vieux:

### DIFF
--- a/test/integration/api/version.bats
+++ b/test/integration/api/version.bats
@@ -15,11 +15,7 @@ function teardown() {
 	# version
 	run docker_swarm version
 	[ "$status" -eq 0 ]
-	[ "${#lines[@]}" -ge 8 ]
 
 	# verify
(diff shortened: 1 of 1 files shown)


vieux:
ping @aluzzardi

test/integration/api/version.bats
```diff
@@ -12,14 +12,14 @@ function teardown() {
 	start_docker 1
 	swarm_manage
 
-	# version
-	run docker_swarm version
+	# Get version output
+	out=$(docker_swarm version)
+
+	run bash -c "echo '$out' | grep swarm/"
```
aluzzardi: I think this test was originally overkill. Just run docker_swarm version and check that `$output =~ swarm/`

vieux: ok

vieux:
@aluzzardi PTAL

PR #1014, opened 2015-07-01
Fix storage integration tests

abronan:
Fix the storage integration tests on Jenkins.

Those were relying on the `--heartbeat` flag for the store failover test. The PR reverts to using `--heartbeat` until we find a proper flag structure for the leader election timeout and the store failover wait time.


vieux:
LGTM

Issue #996, opened 2015-06-29
Issuing commands to Swarm replicas


dreilly1982:
This is more of an issue with the docs at https://docs.docker.com/swarm/multi-manager-setup/ and actual performance.  The second to last paragraph states 'You can use the docker command on any Docker Swarm primary manager or any replica.', while there are many commands that can be issued to the replicas, not all commands are supported.  The control API on the replicas also answers with the underlying Docker API and not the swarm API.  Services such as Jenkins need to point to the Master to work effectively.  I am not positive if there is an issue in the implementation, or just an ambiguity in the Docs.

abronan:
Hi @dreilly1982, +1 the documentation emphasizes on the fact that you can issue commands on any replica but this sounds like you can do this without any impact on performance. Replicas are primarily used in case of the *Primary* **failure**.

The documentation should encourage to always point to the current Primary as the default.

aluzzardi:
Which commands are not supported by the replica?

dreilly1982:
I think now "commands" may be the wrong wording.  The underlying API that is being presented is different on replicas than on the primary.  While the primary when queried through the remote API presents the swarm API, the replicas present the underlying docker engine API.

This can be seen using the /version endpoint.  Example below:
```
# curl http://<primary-hostname>:<swarm-port>/version
{
    "ApiVersion": "1.16",
    "Arch": "amd64",
    "GitCommit": "28cd51b",
    "GoVersion": "go1.3.3",
    "KernelVersion": "",
    "Os": "linux",
    "Version": "swarm/0.3.0"
}
```
```
# curl http://<replica-hostname>:<swarm-port>/version
{
    "ApiVersion": "1.18",
    "Arch": "amd64",
    "GitCommit": "4749651",
    "GoVersion": "go1.4.2",
    "KernelVersion": "3.19.2",
    "Os": "linux",
    "Version": "1.6.0"
}
```

This leaves out necessary information as listed in https://github.com/docker/swarm/tree/master/api when querying against replicas.

If this is indeed as designed, then the wording in the documentation is ambiguous.

aluzzardi:
Hey @dreilly1982, there must a problem.

The replica exposes the **exact** same API - in fact, it just proxies every *command* to the primary, **except** for `/info`.

For whatever reason, I'm guessing `<replica-hostname>:<swarm-port>` are incorrect.

dreilly1982:
I am not connecting to the wrong ports.  I will give you full output of my setup below:

[...]

PR #1016, opened 2015-07-01
Clairify port usage in multi-manager-setup.md

dreilly1982:
Clarified port usages in the assumptions section of multi-manager-setup.md.  This is to address issue #996

### DIFF
--- a/docs/multi-manager-setup.md
+++ b/docs/multi-manager-setup.md
@@ -23,7 +23,7 @@ This section explains how to set up Docker Swarm using multiple **managers**.
 
 ### Assumptions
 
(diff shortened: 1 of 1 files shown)


aanm:
@dreilly1982 I think there should also be a clarification on the agent's IP addresses. It's confusing for new guys to understand that there's:
 * A swarm agent running on a different IP than managers
 * A docker daemon running on the same IP of managers

abronan:
@aanm Agreed.

@dreilly1982 Sorry for the delay on stumbling upon this, are you up to making this change? Otherwise LGTM and we can merge it as it is and make the clarification in another PR.

/ping @moxiegirl

abronan:
This one PR is long standing... let's merge it as it is just a clarification. We can discuss on improving the whole Multi-Manager doc later on.

moxiegirl:
@abronan apologies for letting this linger.

abronan:
@moxiegirl no worries :smiley_cat:

PR #1017, opened 2015-07-02
override hostname in mesos

vieux:
so it should work with our from scratch image

### DIFF
--- a/cluster/mesos/cluster.go
+++ b/cluster/mesos/cluster.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


vieux:
ping @jimenez

cluster/mesos/cluster.go
```diff
@@ -72,10 +73,15 @@ func NewCluster(scheduler *scheduler.Scheduler, store *state.Store, TLSConfig *t
 	// Empty string is accepted by the scheduler.
 	user, _ := options.String("mesos.user", "SWARM_MESOS_USER")
 
+	hostname, _ := os.Hostname()
+	if hostname == "" {
+		hostname = frameworkName
```
tnachen: why framework name?

vieux: @tnachen because I needed something :D do you have another idea ?

tnachen: hmm, I think if we can't get the hostname and there are no overrides, I wonder if we can just don't pass override?
Passing framework name doesn't even seem logical to me..

vieux: the official image we produce has no hostname. we need to give something otherwise mesos-go fails

vieux:
@tnachen PTAL

jimenez:
LGTM

PR #1024, opened 2015-07-04
Troubleshooting revealed need id if title is the same

moxiegirl:
Need an identifier if we want to pages titled the same


moxiegirl:
Author LGTM

ping @aluzzardi @abronan @vieux missed adding the id when I changed the title

moxiegirl:
I'm going to go ahead and merge this.  I want to update the docs this weekend and it is a one line change.

PR #1026, opened 2015-07-05
update libkv godeps, fix zookeeper discovery list

### LINKED ISSUE #974: Swarm manager intermittently unable to see newly joined docker host nodes
Using a test cluster of two docker hosts with _different_ /etc/docker/key.json files, swarm manager (running master branch, rev 174ea0c) occasionally is unable to see the advertisement of the second docker node once its been down past the ttl interval.

When this happens the container running `swarm agent` logs the following:

    time="2015-06-18T15:21:42Z" level=error msg="Discovery error: missing port in address \x01"  

Restarting the container running `swarm manage` "fixes" the issue and `info` accurately reports both nodes.

NB: discovery is an external production zookeeper 3.4.6 three-node cluster Both test hosts are advertised via `swarm join --advertise <external-ip>:2375 zk://$ZOOKEEPER_CLUSTER` and docker is running with `--tlsverify -H 0.0.0.0:2375` (swarm likewise configured and typically working as expected w/ tis). When this occurs the zk cluster has two nodes as expected under `/docker/swarm/nodes`, one for each test-host-ip:2375 tuple with matching node contents.

abronan:
fixes #974 

The only way to get rid of this code in libkv would be in `go-zookeeper` to:

- Investigate and fix the issue with `Get` returning a control character instead of the actual value
- Use `Multi` operations with Get to have a proper `List` implementation in `libkv`

In the meantime the PR fixes the linked issue so we can add and remove nodes properly using zookeeper..

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -52,7 +52,7 @@
 		},
 		{
 			"ImportPath": "github.com/docker/libkv",
-			"Rev": "82ad407bbfaf1ef435a2ebabc498a9afb128ed37"
+			"Rev": "057813e38a46ee5951b1fc33f6f749f7cfce2941"
 		},
 		{
 			"ImportPath": "github.com/gogo/protobuf/proto",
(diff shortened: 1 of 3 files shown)


aluzzardi:
LGTM

vieux:
LGTM

PR #1031, opened 2015-07-06
add name, http_proxy, https_proxy and no_proxy to docker info

### LINKED ISSUE #1029: HttpProxy not showing up with GET /info
My docker host is  behind a proxy. I have set up http_proxy as mentioned in the docs - https://docs.docker.com/articles/systemd/.
But this doesnt show up when I do a GET /info using the remote API. 

"HttpProxy": "",
    "HttpsProxy": "",
    "NoProxy": ""

I am facing a problem similar to the one mentioned in 
https://forums.docker.com/t/proxy-issue-in-swarm/1221

I am not able to start a container, see logs of a container because it keeps throwing a proxy error.

vieux:
Close #1029

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"runtime"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


vieux:
ping @aluzzardi @abronan @vinaychatra

api/handlers.go
```diff
@@ -39,6 +40,19 @@ func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
 		MemTotal:          c.cluster.TotalMemory(),
 	}
 
+	if httpProxy := os.Getenv("http_proxy"); httpProxy != "" {
```
aluzzardi: Is the `if` really necessary? Wouldn't `HttpProxy: os.Getenv("http_proxy")` above have the same resullt?

vieux: probably not, to be sincere I just copy pasted from docker/docker...

I'll update

aluzzardi:
LGTM

PR #1033, opened 2015-07-06
fix static compile with go 1.4

vieux:
See https://github.com/golang/go/issues/9344


vieux:
@tianon is this the best to do or should we keep the tag and add `-installsuffix netgo` ?

tianon:
Yeah that second option is what the go maintainers told us to do (keep the tag and add the installsuffix).

vieux:
@tianon OK, LGTY ?

tianon:
:+1: LGTM

PR #1034, opened 2015-07-07
/info now returns SystemTime

### LINKED ISSUE #949: /info does not return SystemTime
The `/info` endpoint does not return a SystemTime.

This issue might in fact be invalid, because that's added in Remote API version 1.18. I am not sure what version Swarm currently supports in the source tree. I saw it makes the requests to the underlying docker hosts as 1.16, however I'm not sure if that is what actually determines Swarm's Remote API implementation.

vieux:
Fix #949

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -105,7 +105,7 @@
 		},
 		{
 			"ImportPath": "github.com/samalba/dockerclient",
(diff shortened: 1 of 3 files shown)


vieux:
ping @aluzzardi @abronan @ahmetalpbalkan

abronan:
LGTM

ahmetalpbalkan:
LGTM

PR #959, opened 2015-06-15
Container name fix scheduler/filter/README.md

ahmetalpbalkan:
Signed-off-by: Ahmet Alp Balkan

PR #1035, opened 2015-07-07
Fixes 1021 by reformatting the list

### LINKED ISSUE #1021: Lists don't render properly on filters page
originally reported by @alvinr here: https://github.com/docker/docker/issues/14192

# Problem

See 

screen shot 2015-06-24 at 4 06 07 pm

Looks like this should be a list but it is not formatted as a list. May be a generic issue across all lists - would need to be confirmed.

# Reproduce
See https://docs.docker.com/swarm/scheduler/filter/

moxiegirl:
Fix #1021 so lists render. 
Removed future tense.

### DIFF
--- a/docs/scheduler/filter.md
+++ b/docs/scheduler/filter.md
@@ -29,9 +29,9 @@ Constraints are key/value pairs associated to particular nodes. You can see them
 as *node tags*.
 
 When creating a container, the user can select a subset of nodes that should be
-considered for scheduling by specifying one or more sets of matching key/value pairs.
+considered for scheduling by specifying one or more sets of matching key/value
+pairs. This approach has several practical use cases such as:
 
-This approach has several practical use cases such as:
 * Selecting specific host properties (such as `storage=ssd`, in order to schedule
   containers on specific hardware).
 * Tagging nodes based on their physical location (`region=us-east`, to force
@@ -205,7 +205,7 @@ For example, you can run a `nginx` container with the `com.example.type=frontend
     $ docker run -d -p 80:80 --label com.example.type=frontend nginx
      87c4376856a8
 
-    $ docker ps  --filter "label=com.example.type=front"
+    $ docker ps  --filter "label=com.example.type=frontend"
     CONTAINER ID        IMAGE               COMMAND             CREATED                  STATUS              PORTS                           NODE        NAMES
     87c4376856a8        nginx:latest        "nginx"             Less than a second ago   running             192.168.0.42:80->80/tcp         node-1      trusting_yonath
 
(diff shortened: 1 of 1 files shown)


moxiegirl:
ping @vieux

docs/scheduler/filter.md
```diff
@@ -117,13 +117,13 @@ name or  ID.  For example, you can start a container called `frontend` running
...
      87c4376856a8
 
 
     $ docker ps
     CONTAINER ID        IMAGE               COMMAND             CREATED                  STATUS              PORTS                           NODE        NAMES
-    87c4376856a8        nginx:latest        "nginx"             Less than a second ago   running             192.168.0.42:80->80/tcp         node-1      frontend
+    87c4376856a8        nginx:latest        "nginx"             Less than a second ago   running             192.168.0.42:80->80/tcp         node-1      front
 
 
 Then, using `-e affinity:container==frontend` flag schedule a second container to
```
vieux: I think keeping `frontend` is better, it was changed in https://github.com/docker/swarm/pull/959
Also if we decide to change, we should change everywhere there are still lots of `affinity:com.example.type==frontend`

moxiegirl:
ping @vieux Fixed the front to frontend as per your comment

thaJeztah:
@moxiegirl I think these two lists also weren't rendering properly and need a blank line before them as well;

https://github.com/docker/swarm/pull/1035/files#diff-2a607cb41d0d03010fdcc7a59e2b8439R34
https://github.com/docker/swarm/pull/1035/files#diff-2a607cb41d0d03010fdcc7a59e2b8439R233

moxiegirl:
The code in this PR renders this:

image

thaJeztah:
@moxiegirl in the screenshot, the list following "a value must be one of the following:" doesn't render as a list (line 234).

The other list at line 35 ("This approach has several practical use cases such as:") also doesn't render as a list (https://docs.docker.com/swarm/scheduler/filter/)

vieux:
@moxiegirl @thaJeztah what should we do here ?

thaJeztah:
@vieux I think the solution is in https://github.com/docker/swarm/pull/1035#issuecomment-120109420 so if @moxiegirl adds those blank lines I *think* it should be fixed.

thaJeztah:
For the example list in the screenshot (https://github.com/docker/swarm/pull/1035#issuecomment-120167292); perhaps changing that to a table could be nice (but for a separate PR)

moxiegirl:
@vieux I'll fix this and update. @thaJeztah thanks for the clarification on where the issue was

thaJeztah:
LGTM thanks! (do the commits need to be squashed?)

thaJeztah:
umf GitHub cache; already squashed LOL

moxiegirl:
@vieux LGTM and merge?

vieux:
LGTM

PR #1039, opened 2015-07-07
Using the file discovery method, if a slave address failed to registr…

lokikiller:
Using the file discovery method, if a slave address failed to registry itself when the scanning the configuration file at the first time (maybe the slave's docker service donot start), the slave will not be registry itself again, even if the slave's docker service started, unless remove the slave's address from the configuration file and add it again.

Signed-off-by: lokikiller <haotingyi13@otcaix.iscas.ac.cn>


jimmyxian:
@lokikiller Continuous-integration error because of go format error. 
You can use `gofmt -s -w modified_file` to format go code. Just like:
` gofmt -s -w cluster/swarm/cluster.go`
Also, you don't need to create a new PR, just modify --> commit --> push again.

PR #1040, opened 2015-07-07
Fix the slave registry failed

lokikiller:
Using the file discovery method, if a slave address failed to registry itself when the scanning the configuration file at the first time (maybe the slave's docker service donot start), the slave will not be registry itself again, even if the slave's docker service started, unless remove the slave's address from the configuration file and add it again.

Signed-off-by: lokikiller <haotingyi13@otcaix.iscas.ac.cn>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ COPY . /go/src/github.com/docker/swarm
 WORKDIR /go/src/github.com/docker/swarm
 
 ENV GOPATH /go/src/github.com/docker/swarm/Godeps/_workspace:$GOPATH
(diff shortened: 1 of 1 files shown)


jimmyxian:
@lokikiller 
Your signature is failing. Gordon (our bot) expects a format like this:

Signed-off-by: FIRSTNAME LASTNAME \<username@company.com\>

You've only provided a single name, need both your given and your family name.

http://docs.docker.com/project/set-up-git/#set-your-signature-and-an-upstream-remote

lokikiller:
@jimmyxian  Thank you very much, but when I execute
```
$ git commit --amend -s --no-edit
$ git push -f
```
The pull request will be closed?

jimmyxian:
@lokikiller 
When you want rollback code, Maybe you can follow this:
* `# git reset <ID>`      // you can use `git log` to see which ID you want to roll back to
* Re-modify code
* `# git commit -a -s -m "commit message"`
* `git push -f `
```

PR #1044, opened 2015-07-08
Fix the slave registry failed

lokikiller:
When using the file discovery method, if a slave address failed to registry itself when the scanning the configuration file at the first time (maybe the slave's docker service donot start), the slave will not be registry itself again, even if the slave's docker service started, unless remove the slave's address from the configuration file and add it again.

Signed-off-by: Tingyi Hao haotingyi13@otcaix.iscas.ac.cn
Signed-off-by: Tingyi Hao <haotingyi13@otcaix.iscas.ac.cn>

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -18,6 +18,11 @@ import (
 	"github.com/docker/swarm/scheduler/node"
 	"github.com/docker/swarm/state"
 	"github.com/samalba/dockerclient"
+	"time"
+)
+
+const (
+	reAddEnginePeriod = 30 * time.Second
 )
 
 // Cluster is exported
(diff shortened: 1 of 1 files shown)


cluster/swarm/cluster.go
```diff
@@ -158,10 +163,39 @@ func (c *Cluster) hasEngineByAddr(addr string) bool {
...
+		//				_, engine = c.registEngine(addr)
+		//				if engine != nil {
+		//					return
+		//				}
+		//			}
+		//		}()
+		ticker := time.NewTicker(reAddEnginePeriod)
+		go func() {
+			for {
+				select {
```
vieux: This is going to loop like crazy use so much CPU...

cluster/swarm/cluster.go
```diff
@@ -158,10 +163,39 @@ func (c *Cluster) hasEngineByAddr(addr string) bool {
 	return c.getEngineByAddr(addr) != nil
 }
 
-func (c *Cluster) addEngine(addr string) bool {
+func (c *Cluster) addEngine(addr string) {
+	res, engine := c.registEngine(addr)
+	if !res && engine == nil {
+		//		ticker := time.Tick(reAddEnginePeriod)
```
vieux: why keep the comments ?

lokikiller: @vieux I think the comments way maybe better, but I imitate the code in "discovery/file/file.go" from 83 line to 100 line, to maintain the coding style.

aluzzardi:
I understand the issue but I'm not sure this is the right approach.

The reason we don't attempt anymore is because of this;
```go
 			added, removed := currentEntries.Diff(entries)
 			currentEntries = entries
```

Meaning that we won't try to connect to an `added` engine more than once.

Not sure about a better solution yet

aluzzardi:
@lokikiller Could you take a look at #1486?

amitshukla:
@lokikiller Can you please let us know if you are interested in pursuing this PR?

lokikiller:
@amitshukla yeah, I am interested in pursuing this PR

Issue #1037, opened 2015-07-07
key used by master election do not use the given path prefix


cloudfly:
I want to start multiple swarm manager by using etcd.

eg.
```
swarm manager --replication --addr=127.0.0.1:2375 etcd://127.0.0.1:4001/myapp/swarm
```
the key used to elect master, is always `/docker/swarm/leader`, ignoring prefix `/myapp/swarm`

this is designed on purpose, or a negligence ?

jimmyxian:
@cloudfly 
Etcd used here has two purposes:
* Used for discovery serivice:  The path is user defined `etcd://127.0.0.1:4001/myapp/swarm`
* Leader Election: The path is const `const leaderElectionPath = "docker/swarm/leader"`

cloudfly:
@jimmyxian 
yes, I know.

same with `leaderElectionPath`, `discoveryPath` is also a const string in code.
```
 discoveryPath = "docker/swarm/nodes"
```

swarm-0.3 changed the discovery path, it join with prefix, now use `/myapp/swarm/docker/swarm/nodes`. 

but the `leaderElectionPath` not. 

swarm-0.3 now write two place on my etcd. so uncomfortable.

jimmyxian:
@cloudfly 
Yeah, maybe we can use the same prefix.
ping @aluzzardi @abronan

abronan:
I'm not in favor of using the same prefix for a simple reason: it can have a different meaning for both leader election and discovery. Using the leader election key with a `mydiscovery/` prefix does not make much sense in the context of leader election.

You might want to separate the key space for both mechanisms.

Right now we enforce it but a specific flag could be included for leader election if we want to modify the path with a given prefix.

cloudfly:
I think it's necessary for user to custom the prefix for leader election path.  

It will be more convenient to manage the my path. I had save all my data under /my-cluster-name, and do not want anything outside of there.

abronan:
@cloudfly Yes I think it's necessary too, adding a new flag you can then specify the same prefix for `discovery` and `leader election` so you can keep those under the same key space.

vieux:
@cloudfly @abronan I agree not the exact same prefix, but why not append `/leader` to the current path, so in your example it would become `/myapp/swarm/leader` so we don't have to use another flag.

cloudfly:
I do not know how others use swarm and etcd, here is mine.

[...]

PR #1045, opened 2015-07-08
add leader-path-prefix flag

cloudfly:
#1037 

add new flag `--leader-path-prefix` to add prefix to leader-election-path. 

- empty by default, leader election will use `docker/swarm/leader` as before

- if set `--leader-path-prefix="/mycluster"`, will use `/mycluster/docker/swarm/leader`


abronan:
Thanks for the contribution @cloudfly, it just requires a sign-off following Gordon's advice :)

PR #1046, opened 2015-07-08
add leader-path-prefix flag

cloudfly:
Signed-off-by: chenyunfei <chenyunfeihappy@gmail.com>

#1037 

add new flag `--leader-path-prefix` to add prefix to leader-election-path. 

- empty by default, leader election will use `docker/swarm/leader` as before

- if set `--leader-path-prefix="/mycluster"`, will use `/mycluster/docker/swarm/leader`

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -25,7 +25,7 @@ var (
 				flStore,
 				flStrategy, flFilter,
 				flHosts,
-				flLeaderElection, flManageAdvertise,
+				flLeaderElection, flManageAdvertise, flLeaderPathPrefix,
 				flTLS, flTLSCaCert, flTLSCert, flTLSKey, flTLSVerify,
 				flHeartBeat,
 				flEnableCors,
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -122,4 +122,10 @@ var (
 		Name:  "replication",
 		Usage: "Enable Swarm manager replication",
 	}
+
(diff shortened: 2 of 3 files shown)


aluzzardi:
Woops - so I opened #1048 as well.

The difference is that in this one it's configurable, on the other one it's based on the discovery prefix (so they are shared).

Thoughts?

cloudfly:
I agree with your thought. and also prefer to use the same prefix.  

but considering that there maybe other users don't want to share the prefix,  I add a new flag.

abronan:
@aluzzardi I'd like to be able not to share the same key space for both mechanisms, I use my [etcd/consul/zookeeper] cluster specifically for Swarm (not sharing it with any other service). 

With the same key space, someone can mistakenly wipe out the entire cluster role state (leader/follower, nodes inside the cluster) because you can still access and manage the store independently from Swarm. This is a worst case scenario but this can happen and result in an interruption of service.

In my case I'd prefer to have this configurable even though this is one more flag to set ;)

aluzzardi:
If we do that, then we'll have to add a flag for every single file we write out to consul/etcd/zookeeper

abronan:
Fixed by #1048

@aluzzardi I first merged the intended behavior as there seems to be a consensus on using the same prefix for both.

@cloudfly Thank you for the contribution, we can re-open this one later if we improve the UI around the configuration of the store/discovery.

PR #1048, opened 2015-07-09
Leader Election: Use same path prefix as discovery.

### LINKED ISSUE #1037: key used by master election do not use the given path prefix

aluzzardi:
Fixes #1037

### DIFF
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -120,9 +120,10 @@ func setupReplication(c *cli.Context, cluster cluster.Cluster, server *api.Serve
 		log.Fatal("Leader election is only supported with consul, etcd and zookeeper discovery.")
 	}
 	client := kvDiscovery.Store()
+	p := path.Join(kvDiscovery.Prefix(), leaderElectionPath)
 
-	candidate := leadership.NewCandidate(client, leaderElectionPath, addr)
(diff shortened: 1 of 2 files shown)


vieux:
If I use `0.0.0.0:1234/swarm/myapp` the leader election prefix will be `/swarm/myapp/docker/swarm/leader` 

is that expected ?

aluzzardi:
Yep - it will be `/swarm/myapp/docker/swarm/nodes` and `/swarm/myapp/docker/swarm/leader`

aluzzardi:
We could change that by using `/docker/swarm` as default prefix and override it with the command line, although I'm not sure.

The reasoning is that you could have stuff such as `/swarm/myapp/docker/network/peers` or whatnot.

vieux:
LGTM ping @abronan

abronan:
@vieux See my comment on #1046. I'm LGTM and not against merging this PR first but ultimately it would be better if we could configure the prefix independently for both mechanisms.

PR #1049, opened 2015-07-09
update mesos-go

vieux:
We need this commit https://github.com/mesos/mesos-go/commit/10ceb4fb01222c59f313fec154a6af275eadb6c1 to support IP instead of hostname for the mesos master uri


vieux:
ping @jimenez @tnachen

tnachen:
ah cool, lgtm!

PR #1056, opened 2015-07-12
preserve status code from POST /containers/*/exec

### LINKED ISSUE #1055: handler for POST /containers/{name:.*}/exec does not preserve 201 status
The docker remote API specifies that a 201 is returned on successful
creation of an Exec and a strict client will not accept that being
turned into a 200.

See https://docs.docker.com/reference/api/docker_remote_api_v1.19/#exec-create

An example of this causing a problem is the _de facto_ node client, [dockerode](https://github.com/apocas/dockerode).

rmg:
The docker remote API specifies that a 201 is returned on successful
creation of an Exec and a strict client will not accept that being
turned into a 200.

https://docs.docker.com/reference/api/docker_remote_api_v1.19/#exec-create

Resolves #1055


vieux:
LGTM

PR #1069, opened 2015-07-17
ECS discovery backend.

ejholmes:
This adds a Discovery backend for [Amazon EC2 Container Service](http://aws.amazon.com/ecs/) using the [DescribeContainerInstances](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_DescribeContainerInstances.html) API.

I maintain the open source PaaS called [Empire](https://github.com/remind101/empire) and this would be really handy for a couple of features:

1. Running one off tasks and ensuring that they're scheduled onto a machine in the ECS cluster that it manages.
2. Ensuring the docker image is pulled on all machines when deploying.

I've yet to test this on a large ECS cluster, so there may be some limits and pagination that I need to take into account.


discovery/ecs/ecs.go
```diff
@@ -0,0 +1,150 @@
...
+				Port: s.port,
+			})
+		}
+	}
+
+	return entries, nil
+}
+
+// Watch is exported
+func (s *Discovery) Watch(stopCh <-chan struct{}) (<-chan discovery.Entries, <-chan error) {
```
ejholmes: This is a direct clone of [file.Discovery#Watch](https://github.com/docker/swarm/blob/729001c9b458d7467986b189ac9b6647191a0fff/discovery/file/file.go#L64). I could extract a more generic Discovery that takes a `Fetch() (discovery.Entries, error)` interface. Dunno if it's really worth it.

saada:
:heart:

aluzzardi:
Thanks @ejholmes 

A few design questions: Doesn't this bypass entirely ECS? It's basically accessing the underlying EC2 boxes and it schedules containers on them. What would be the difference between doing this and simply running Swarm on EC2?

ejholmes:
ECS is also a cluster manager, much like etcd or consul except it's managed. ECS provides Services and Tasks which use the internal ECS cluster manager to get the state of the cluster, but there's still use cases like I mentioned where you would want direct access to the docker daemon on the cluster hosts where RunTask or StartTask api's don't give enough control.

It would be pretty nice to be able to scale Docker in this way using ECS as the backend. Another example would be using Swarm + ECS as a docker image build system that can be scaled out horizontally with very little operational overhead.

discovery/ecs/ecs.go
```diff
@@ -0,0 +1,154 @@
...
+	s.ecs = ecs.New(config)
+	s.ec2 = ec2.New(config)
+	s.cluster, s.port = parsePath(path)
+	return nil
+}
+
+func (s *Discovery) fetch() (discovery.Entries, error) {
+	var entries discovery.Entries
+
+	list, err := s.ecs.ListContainerInstances(&ecs.ListContainerInstancesInput{
```
samuelkarp: It looks like you're not paginating through the results of `ListContainerInstances`. ECS will return a [maximum of 100 results](http://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_ListContainerInstances.html#ECS-ListContainerInstances-request-maxResults) for this call, but you can use the [`ListContainerInstancePages` function](http://docs.aws.amazon.com/sdk-for-go/api/service/ecs/ECS.html#ListContainerInstancesPages-instance_method ) to iterate through the whole set. Note that ECS will also only allow you to pass up to 100 Container Instances to the `DescribeContainerInstances` API, so you'd want to call that during your iteration through the results.

ejholmes: Thanks for the tip! I'll update this with pagination support.

chanwit:
Correct me if I'm wrong.

If ECS is a cluster manager, does it have its own way for scheduling resources? If so, Swarm + ECS will be likely the same to Swarm + Mesos?

vieux:
I'm going to close this one. It's too specific to be merged in the swarm codebase.
Maybe we will have plugins for discovery backends and this one could be one.

PR #1071, opened 2015-07-17
Fix scheduler typos and broken links

peggyl:
Signed-off-by: Peggy Li <peggyli.224@gmail.com>

### DIFF
--- a/docs/discovery.md
+++ b/docs/discovery.md
@@ -212,7 +212,7 @@ You can contribute a new discovery backend to Swarm. For information on how to d
 
 ## Docker Swarm documentation index
 
-- [User guide](index.md)
-- [Sheduler strategies](/scheduler/strategy.md)
-- [Sheduler filters](/scheduler/filter.md)
+- [User guide](/)
+- [Scheduler strategies](/scheduler/strategy.md)
+- [Scheduler filters](/scheduler/filter.md)
 - [Swarm API](/api/swarm-api.md)
--- a/api/README.md
+++ b/api/README.md
@@ -34,5 +34,5 @@ The Docker Swarm API is mostly compatible with the [Docker Remote API](https://d
 
 - [User guide](https://docs.docker.com/swarm/)
 - [Discovery options](https://docs.docker.com/swarm/discovery/)
-- [Sheduler strategies](https://docs.docker.com/swarm/scheduler/strategy/)
-- [Sheduler filters](https://docs.docker.com/swarm/scheduler/filter/)
+- [Scheduler strategies](https://docs.docker.com/swarm/scheduler/strategy/)
+- [Scheduler filters](https://docs.docker.com/swarm/scheduler/filter/)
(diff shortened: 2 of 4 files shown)


abronan:
LGTM

ping @moxiegirl @thaJeztah

docs/discovery.md
```diff
@@ -212,7 +212,7 @@ You can contribute a new discovery backend to Swarm. For information on how to d
 
 ## Docker Swarm documentation index
 
-- [User guide](index.md)
-- [Sheduler strategies](/scheduler/strategy.md)
-- [Sheduler filters](/scheduler/filter.md)
-- [Swarm API](/api/swarm-api.md)
+- [User guide](https://docs.docker.com/swarm/)
```
moxiegirl: @peggyl Thanks for the contribution. These links are actually built with our static file generator.  If you run `make docs` in the docs directory and test locally, you'll find all but the first was resolving.  So, we don't need the full path name here.  

This is correct:
```
- [User guide](/)
- [Sheduler strategies](/scheduler/strategy.md)
- [Sheduler filters](/scheduler/filter.md)
- [Swarm API](/api/swarm-api.md)
```

peggyl: Thanks for the feedback, @moxiegirl!

I think what happened was I was reading the docs on Github rather than docs.docker.com. I just took another look, and the [User Guide](https://github.com/docker/swarm/blob/master/docs/index.md) is the only of the four links that works on https://github.com/docker/swarm/blob/master/docs/discovery.md,, but the only link that _doesn't_ resolve correctly on https://docs.docker.com/swarm/discovery/.  

It'd be great if the links worked on both Github and docs.docker.com, but if the latter is the supported source, I'm okay stripping this PR to just fix the spelling typo itself. Thanks again!

moxiegirl: Hi @peggyl everything in the `docs` directory is intended to work on `docs.docker.com` -- GitHub repos serve individual pages.  They aren't intended to work like a web. GitHub itself does this: https://goo.gl/Uwh9q3

So, if you just want to do the one typo, please do and thank you for your PR.

peggyl:
@moxiegirl Thanks for the comments; I've removed the hard-coded links. Sorry about the confusion, I was thrown off by the fact the first link did work (on Github). 

Would you prefer I also update the User Guide link to point to `/` instead of `index.md` (as mentioned in https://github.com/docker/swarm/pull/1071#discussion_r34959222), or just leave this PR as-is (as mentioned in https://github.com/docker/swarm/pull/1071#discussion_r34960700)? Thanks!

moxiegirl:
Hey @peggyl thanks for asking. I tested with the (/) and it works perfectly. Let's go with that. :-D

moxiegirl:
LGTM

moxiegirl:
Thanks @peggyl for the contributioni

thaJeztah:
LGTM, thanks!

Issue #1025, opened 2015-07-04
Manager leadership election can fail when paired with Consul 0.5.2


jsipprell:
Using `swarm 0.4.0 (d0b0491)`, `docker 1.7.1-rc1` and `consul 0.5.2` w/ a multi-engine cluster that has two independent manager containers on different docker engines. Heartbeat and ephemeral timing are all unchanged from the defaults (20s and 60s, I believe).

If sufficient time passes (often > 30min or so) without a leadership change and the current primary manager is terminated, leadership election of the failover manager does not complete or does not fully complete and the following errors are logged:

From the swarm instance that should be in replica->primary transition:

> `manager_1 | time="2015-07-04T17:59:29Z" level=error msg="failed to acquire lock: Unexpected response code: 500 (rpc error: rpc error: Invalid session)" `

From the local consul client instance for that docker engine (logs are from two separate occurrences):

> `2015/07/04 15:44:26 [ERR] http: Request /v1/kv/docker/swarm/leader?acquire=3d2f35d7-48fd-031c-c60b-17eadf4f9a0b&flags=3304740253564472344, error: rpc error: rpc error: Invali
d session`

Best Regards,

Jesse

aanm:
@jsipprell how many "Unexpected response code: 500" do you see?

abronan:
Thanks @jsipprell, I will try to reproduce. I think that in that case, the key stays locked forever and the session is not released properly which leaves all other instances unable to lock the key in return.

In the meantime I'm working on changes for the Consul default Lock that we are using for the Leader Election process so it can use custom Sessions and not default ones. This way we are in control of what may happen in the background with sessions (like we do for ephemeral keys and discovery node entries).

jsipprell:
@aanm AFAICT, there is only a single "Unexpected response code: 500" per incident (per consul/swarm/docker engine anyway).

abronan:
@jsipprell Can you reproduce each time your run the cluster for a long time or is this occasional? I tried two times (using `master`) running the cluster for 2 hours and killing the current Primary would rightfully elect a new one. Could you try with the latest `master` and see if you can reproduce again? Thanks!

jsipprell:
@abronan I'll rebuild from latest master and give it a shot.

Unfortunately, I've noticed there can be a very significant delay between occurrences. I just made it happen but it had been up for many hours w/out leadership change beforehand.

Here's a little more consul debug output context from right around the latest event:

[...]

Issue #983, opened 2015-06-20
Leader Election: wrong information reported if store fails


abronan:
Swarm reports wrong informations when the store fails and one of the manager is restarted to recover the leader election process. (This seems to be only a **UI issue** as the requests are proxied to the right machine)

**Manager-1** (role: **Replica**, Primary: **manager-2**)
```
vagrant@swarm-manager-0:~$ export DOCKER_HOST=192.168.42.200:4000
vagrant@swarm-manager-0:~$ docker info
Containers: 278
Images: 28
Storage Driver:
Role: replica
Primary: 192.168.42.201:4000
Strategy: spread
Filters: affinity, health, constraint, port, dependency
Nodes: 3
 swarm-agent-0: 192.168.42.100:2375
  └ Containers: 200
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 2.053 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.13.0-49-generic, operatingsystem=Ubuntu 14.04.2 LTS, storagedriver=aufs
 swarm-agent-1: 192.168.42.101:2375
  └ Containers: 39
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 2.053 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.13.0-49-generic, operatingsystem=Ubuntu 14.04.2 LTS, storagedriver=aufs
 swarm-agent-2: 192.168.42.102:2375
  └ Containers: 39
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 2.053 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.13.0-49-generic, operatingsystem=Ubuntu 14.04.2 LTS, storagedriver=aufs
Execution Driver:
Kernel Version:
Operating System:
CPUs: 3
Total Memory: 6.158 GiB
Name:
ID:
Http Proxy:
Https Proxy:
No Proxy:
```

**Manager-2** (role: **Primary**, Primary: **self**)
```
vagrant@swarm-manager-0:~$ export DOCKER_HOST=192.168.42.201:4000
vagrant@swarm-manager-0:~$ docker info
Containers: 278
Images: 28
Storage Driver:
Role: primary
Strategy: spread
Filters: affinity, health, constraint, port, dependency
Nodes: 3
 swarm-agent-0: 192.168.42.100:2375
  └ Containers: 200
[...]

PR #1078, opened 2015-07-20
Leadership: Fault tolerant Leader Election mechanism

### LINKED ISSUE #982: Leader Election: no elected leader if the store fails and recovers after some time
If the store fails and recovers after some time, there will be no elected leader left and we couldn’t query the Managers anymore. Leader Election should have the same recovery mechanisms as for the discovery watches.

This leaves us with:

```
FATA[0000] Error response from daemon: No elected primary cluster manager
``` 

As a result, we need to restart one of the manager to take the lead again.

(This issue seems to happen only on Consul and Etcd, Leader election on zookeeper seems to work even in the case of a store failure)

abronan:
Makes the leader Election process resilient to failures of the backend storage (on Consul and etcd, zookeeper already allowed that naturally by keeping the leader information) using an error channel.

Fixes #982 #983 #1025 

### DIFF
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -21,10 +21,12 @@ import (
 	"github.com/docker/swarm/scheduler/filter"
 	"github.com/docker/swarm/scheduler/strategy"
 	"github.com/docker/swarm/state"
+	"github.com/gorilla/mux"
 )
 
 const (
 	leaderElectionPath = "docker/swarm/leader"
+	defaultRecoverTime = 10 * time.Second
 )
 
 type logHandler struct {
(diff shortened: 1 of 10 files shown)


leadership/README.md
```diff
@@ -57,3 +55,30 @@ for leader := <-leaderCh {
 
 A typical use case for this is to be able to always send requests to the current
 leader.
+
+## Fault tolerance
+
+Leadership returns an error channel for Candidates and Followers that you can use
```
aluzzardi: Why the need for an error channel?

Basically, you already know that when the channel is closed (as in, you exit the for loop) there **is** an error

ahmetalpbalkan: @aluzzardi I think errCh helps calling `underwood.RunForElection()` only once and outside the loop,

otherwise we'd be asking for an `electedCh` every single time inside the `for{}`. @abronan, is that right ?

abronan: The initial intent of having an error chan is for the client to log errors instead of the Leadership package which is meant to be used as a library (which will potentially be in its own repository like `libkv`). Also `campaign` and `follow` are both running inside a goroutine in the library so you need a way to communicate back potential errors. Also the error chan is optional so the first version of the code still works.

Regarding the annoying usage of two `for` loops and a `label`, the only way I see to get rid of this is to do something like:

```go
func setupReplication(...) {
    [...]
    go run(candidate, server, primary, replica)
    go follow(follower, replica, addr)
    [...]
}

func run(candidate *leadership.Candidate, server *api.Server, primary *mux.Router, replica *api.Replica) {
    electedCh, errCh := candidate.RunForElection()
    for {
        select {
        case isElected := <-electedCh:
		if isElected {
			log.Info("Leader Election: Cluster leadership acquired")
			server.SetHandler(primary)
		} else {
			log.Info("Leader Election: Cluster leadership lost")
			server.SetHandler(replica)
		}

        case err := <-errCh:
            log.Error(err)
            time.Sleep(defaultRecoverTime)
            go run(candidate, server, primary, replica)
            return
        }
    }
}
```

This is *meh* to me, as we exit and create a new goroutine at each attempt. Is there any more elegant solution than this? Anyway we can still make the usage of Leadership more convenient later on.

abronan:
@aluzzardi PTAL, removed the label. Tested on the 3 KV backends. Will add integration tests in a separate PR (I don't need configurable Leader TTL for this and this will take less time than for the Leader test).

vieux:
ping @aluzzardi

cli/manage.go
```diff
@@ -128,34 +130,56 @@ func setupReplication(c *cli.Context, cluster cluster.Cluster, server *api.Serve
...
 			} else {
-				log.Info("Cluster leadership lost")
+				log.Info("Leader Election: Cluster leadership lost")
 				server.SetHandler(replica)
 			}
+
+		case err := <-errCh:
+			log.Error(err)
+			time.Sleep(defaultRecoverTime)
+			go run(candidate, server, primary, replica)
```
aluzzardi: This is a little weird - when you call `run()` it will block but then when it returns it will continue running as a goroutine in the background.

How about just returning here? And in `setupReplication`:
```go
go func() {
  for {
    run(...)
    time.Sleep(...)
   }
}
```

cli/manage.go
```diff
@@ -128,34 +130,56 @@ func setupReplication(c *cli.Context, cluster cluster.Cluster, server *api.Serve
...
-		electedCh := candidate.ElectedCh()
-		for isElected := range electedCh {
+	go run(candidate, server, primary, replica)
+	go follow(follower, replica, addr)
+
+	server.SetHandler(primary)
+}
+
+func run(candidate *leadership.Candidate, server *api.Server, primary *mux.Router, replica *api.Replica) {
+	electedCh, errCh := candidate.RunForElection()
```
aluzzardi: Do you need to defer close those?

cli/manage.go
```diff
@@ -128,34 +130,56 @@ func setupReplication(c *cli.Context, cluster cluster.Cluster, server *api.Serve
...
-	}()
+	}
+}
 
-	go func() {
-		follower.FollowElection()
-		leaderCh := follower.LeaderCh()
-		for leader := range leaderCh {
-			log.Infof("New leader elected: %s", leader)
+func follow(follower *leadership.Follower, replica *api.Replica, addr string) {
```
aluzzardi: same comments as run

cli/manage.go
abronan: They are `defer closed` in the Leadership side. Can't close them here as those are receive-only channels.

cli/manage.go
abronan: +1 will make the change on both `candidate` and `follow`

abronan:
@aluzzardi PTAL

ahmetalpbalkan:
@abronan Hi, will this be able to make it to 0.4 or will be in for 0.5?

abronan:
Hi @ahmetalpbalkan, this will be in 0.4 ;)

ahmetalpbalkan:
@abronan thanks!

Issue #1081, opened 2015-07-21
Nil pointer dereference panic at cluster/swarm.(*Cluster).createContainer


wrotki:
Running Swarm on two machines for a while (a couple hours, with a minor but constant load), I found the following nil pointer dereference panic in the logs. It looks like Engine.Create can return nil for an error, but also nil for a container -  it can happen when Engine.refreshContainer doesn’t pick up the newly created container, due to  a network hiccup perhaps.

Here’s the log (from swarm manage container):

time="2015-07-18T16:05:45Z" level=info msg="New leader elected: " 
time="2015-07-18T16:05:45Z" level=info msg="New leader elected: 54.204.78.44:3376" 
time="2015-07-18T16:05:45Z" level=info msg="Cluster leadership acquired" 
time="2015-07-18T16:05:46Z" level=error msg="HTTP error: No such container app_stateless_PolyWeb_6173736574736962298954567138918" status=404 
time="2015-07-18T16:05:46Z" level=error msg="HTTP error: No such image: app_stateless_PolyWeb_6173736574736962298954567138918" status=404 
time="2015-07-18T16:06:14Z" level=info msg="Cluster leadership lost" 
time="2015-07-18T16:06:14Z" level=info msg="New leader elected: " 
time="2015-07-18T16:06:14Z" level=error msg="HTTP error: No elected primary cluster manager" status=500 
time="2015-07-18T16:06:14Z" level=info msg="New leader elected: 54.221.38.147:3376" 
time="2015-07-18T16:06:32Z" level=error msg="Unable to update state of container \"02425a916ad7a942389b2baf18ad9189efb18af708ea78920524fd347c695c61\": Not found" id="R25B:WT45:XTNF:USYK:PX2V:GFG7:HYJX:VQZA:BYNV:YASA:ENTZ:U7MJ" name="ip-10-95-172-158" 
2015/07/18 16:06:32 http: panic serving 54.221.38.147:59256: runtime error: invalid memory address or nil pointer dereference
goroutine 225 [running]:
net/http.func·011()
/usr/src/go/src/net/http/server.go:1130 +0xbb
github.com/docker/swarm/cluster/swarm.(*Cluster).createContainer(0xc20800a4e0, 0xc208db1180, 0xc208d1c5bd, 0x35, 0x0, 0x0, 0x0, 0x0)
/go/src/github.com/docker/swarm/cluster/swarm/cluster.go:136 +0x6d4
[...]

PR #1082, opened 2015-07-21
Avoid nil pointer dereference in (*Cluster).createContainer

wrotki:
Swarm Engine.Create can return nil,nil when refresh fails due to a network hiccup. It causes a panic described by issue https://github.com/docker/swarm/issues/1081 .

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -425,11 +425,11 @@ func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool) (*
 	e.RLock()
 	defer e.RUnlock()
 
-	justCreated := e.containers[id]
-	if justCreated == nil {
+	container := e.containers[id]
+	if container == nil {
 		err = errors.New("Container created but refresh didn't report it back")
 	}
-	return justCreated, err
+	return container, err
 }
 
 // RemoveContainer a container from the engine.
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -425,7 +425,11 @@ func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool) (*
 	e.RLock()
 	defer e.RUnlock()
 
-	return e.containers[id], nil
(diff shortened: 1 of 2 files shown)


aluzzardi:
We should identify the root cause - this is not supposed to happen.

As a better fix, I'd rather have `Engine.Create()` return an error instead of nil container.

As a convention, one can assume that if `err == nil`, then the value is ready to use (no need to check for nil-ness)

wrotki:
Agreed , returning error from EngineCreate is better.

The root cause - no clue how hard it is to reproduce the problem, I accidentally spotted a single panic instance in my logs. From what understand about what sequence might lead to this, it'd need to be a container creation call returning success, plus subsequent query for the newly created container returning zero elements.  Will try to run more tests in attempt to catch docker red-handed.

aluzzardi:
Awesome - thanks @wrotki 

Feel free to update this PR whenever you can (with the error returned from Engine create) and we'll merge it.

abronan:
:+1: encountered this recently, would be cool to see this merged ASAP.

wrotki:
Alright - I'm in the middle of something right now, but I should be able to update the PR within a day

vieux:
ping @aluzzardi

cluster/engine.go
```diff
@@ -425,7 +425,11 @@ func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool) (*
 	e.RLock()
 	defer e.RUnlock()
 
-	return e.containers[id], nil
+	justCreated := e.containers[id]
```
aluzzardi: Sorry for being pedantic but why not `container` or `c`? Semantically speaking, everything on the left side of `:=` was *just created* :)

wrotki: Fine with me, will update in a sec

aluzzardi:
Thanks @wrotki !

LGTM

PR #1087, opened 2015-07-25
Document consul proto URI correctly

ahmetalpbalkan:


docs/discovery.md
```diff
@@ -116,11 +116,11 @@ You can list the nodes in your cluster.
 On each of your nodes, start the Swarm agent. The node IP address
 doesn't need to be public as long as the Swarm manager can access it.
 
-    swarm join --advertise=<node_ip:2375> consul://<consul_addr>/<optional path prefix>
+    swarm join --advertise=<node_ip:2375> consul://<consul_addr1>,<consul_addr2>/<optional path prefix>
```
thaJeztah: (Not for this PR) Wondering if we should use `[.....]` for optional elements, e.g.

    swarm join --advertise=<node_ip:2375> consul://<consul_addr1>[,<consul_addr2>]/[<optional path prefix>]

thaJeztah:
LGTM, thanks Ahmet

docs/discovery.md
ahmetalpbalkan: @thaJeztah that's actually a nice idea but CLI docs sort of like this, I guess we should either change both or none.

image

thaJeztah: When done, it should be done consistently, I agree. Perhaps you could create a separate issue (so that we don't delay merging this PR). It was just something I noticed :smile:

ahmetalpbalkan: @thaJeztah separate PR would be nicer so we can discuss in detail. right now this is just making this doc consistent within itself. :)

thaJeztah: sgtm!

vieux:
@abronan needs to confirm, but I think consul only takes one addr.
@ahmetalpbalkan are you using an outdated version of swarm ? here is the current usage:

screen shot 2015-07-27 at 10 57 12

ahmetalpbalkan:
@vieux yes I think I was using `0.2.0`, however why in your screenshot consul doesn't list multiple IPs? Is it not supported?

abronan:
@ahmetalpbalkan Consul only takes one address :) See #583 for more informations.

abronan:
@ahmetalpbalkan It was a documentation error that we fixed in 0.3.

ahmetalpbalkan:
@abronan hmm, I'm using `swarm version 0.3.0 (28cd51b)` and `join --help` also shows 1 IP for consul. However my manager runs with multiple consul IPs without giving any errors?

```
azureuser@swarm-master-0:~$ docker inspect compose_swarm_1 | head -n 12
[
{
    "Id": "8122b9bb9930e2ca419702d91555bc92c87538bf92c4bf29f438a9af175cbb99",
    "Created": "2015-07-26T02:11:20.3822512Z",
    "Path": "/swarm",
    "Args": [
        "manage",
        "--replication",
        "--advertise",
        "10.0.0.4:2376",
        "consul://10.0.0.4:8500,10.0.0.5:8500,10.0.0.6:8500/nodes"
    ],
```

abronan:
@ahmetalpbalkan Good point! The discovery part changed a lot since `0.2` and this is actually due to the migration to `libkv`. So `swarm` will accept multiple IPs but `libkv` will discard other specified IPs for Consul. I guess that we should add a check upstream in swarm to avoid confusion or just ignore and document that properly.

ahmetalpbalkan:
@abronan ok opening another bug.

PR #1099, opened 2015-07-29
Small cleanup of Cluster.createContainer

aluzzardi:


jimmyxian:
@aluzzardi This cleanup will save the soft-image-affinity in ContainerConfig
>4/ Retry with a soft-affinity (but don't store it in the ContainerConfig)

aluzzardi:
@jimmyxian You are right :)

PR #1109, opened 2015-07-30
fix CI

vieux:


abronan:
LGTM

ping @aluzzardi

aluzzardi:
Why overlay?

vieux:
@aluzzardi the CI was switched to overlay, dind has to use the same

vieux:
It will require other fixed (mostly mesos) but the CI is now green. It's good to merge

aluzzardi:
LGTM

PR #1115, opened 2015-08-04
return an error when the scheduled engine no longer exists

mountkin:


abronan:
Hi @mountkin, thanks for the contribution. I'll test it first to see if I can still reproduce with the fix (with single and multi-master) and come back here once done ;)

abronan:
@mountkin Closing, sorry for the lack of attention on this one. We've changed the scheduler in the meantime to allow for non blocking operations and we've changed the way we manage engines to avoid this issue. This PR should not be relevant anymore. Thanks for opening it in the first place though. This has brought our attention on this issue.

PR #1124, opened 2015-08-06
improve usage for discovery

### LINKED ISSUE #951: Confusing error when discovery proto not specified
When I forget then `token://` part in the discovery URI, it tried to parse the token as `<ip1>,<ip2>`:

```
docker run -d swarm join --addr=0.0.0.0:2376 8a07dda59613edxxxxxxxxxxxxxxxx
time="2015-06-15T05:19:55Z" level=fatal msg="missing port in address 8a07dda59613ed702cf94dea56a4d07c"
```

This probably would be hard to diagnose for a newcomer falling into the same mistake. If the input was parsed more systematically, I think Swarm can show user a better error message.

Since the `<discovery>` argument in `swarm join` is defined as:

```
   discovery	discovery service to use [$SWARM_DISCOVERY]
            	 * token://<token>
            	 * consul://<ip1>,<ip2>/<path>
            	 * etcd://<ip1>,<ip2>/<path>
            	 * file://path/to/file
            	 * zk://<ip1>,<ip2>/<path>
            	 * <ip1>,<ip2>
```

I think the list of static IPs option should also be in `proto://input` format, like:

* `tcp://<ip1>,<ip2>`
* `static://<ip1>,<ip2>`

to be consistent with the rest of the input formats. Does anybody else feel the same?

vieux:
fixes #951

I'm not in favor on enforcing `nodes://` I like the fact that it's optional, so I tried to update the error message.

### DIFF
--- a/cli/help.go
+++ b/cli/help.go
@@ -39,7 +39,7 @@ Arguments:
                    * etcd://<ip1>,<ip2>/<path>
                    * file://path/to/file
                    * zk://<ip1>,<ip2>/<path>
-                   * <ip1>,<ip2>{{end}}{{if .Flags}}
+                   * [nodes://]<ip1>,<ip2>{{end}}{{if .Flags}}
 
 Options:
    {{range .Flags}}{{.}}
--- a/discovery/nodes/nodes.go
+++ b/discovery/nodes/nodes.go
@@ -1,6 +1,7 @@
 package nodes
 
 import (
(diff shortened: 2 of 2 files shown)


vieux:
ping @ahmetalpbalkan

aluzzardi:
LGTM

chanwit:
LGTM

PR #1135, opened 2015-08-07
add support for role

vieux:

### DIFF
--- a/cli/help.go
+++ b/cli/help.go
@@ -47,6 +47,7 @@ Options:
                                     {{printf "\t * mesos.address=\taddress to bind on [$SWARM_MESOS_ADDRESS]"}}
                                     {{printf "\t * mesos.port=\tport to bind on [$SWARM_MESOS_PORT]"}}
                                     {{printf "\t * mesos.offertimeout=10m\ttimeout for offers [$SWARM_MESOS_OFFER_TIMEOUT]"}}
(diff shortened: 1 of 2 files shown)


cli/help.go
```diff
@@ -47,6 +47,7 @@ Options:
                                     {{printf "\t * mesos.address=\taddress to bind on [$SWARM_MESOS_ADDRESS]"}}
                                     {{printf "\t * mesos.port=\tport to bind on [$SWARM_MESOS_PORT]"}}
                                     {{printf "\t * mesos.offertimeout=10m\ttimeout for offers [$SWARM_MESOS_OFFER_TIMEOUT]"}}
+                                    {{printf "\t * mesos.role=\trole for offers [$SWARM_MESOS_ROLE]"}}
```
aluzzardi: What's the default? And what are the possible options? We should reject invalid options (e.g. `mesos.role=drunk`)

aluzzardi:
Can you add a more explicit commit message?

aluzzardi:
Is this `changelog` worthy? I added the `impact` label

cli/help.go
vieux: any string is valid, it's defined by the mesos master, so yes `drunk` can be valid.

vieux: :)

vieux:
closing for now, I need to add test before merging

PR #1144, opened 2015-08-11
Fixing links after check

moxiegirl:
- Ran a crawler over swarm docs to find broken links --- these are fixes

### DIFF
--- a/docs/discovery.md
+++ b/docs/discovery.md
@@ -208,7 +208,7 @@ And for the `nodes` discovery method.
 
 ## Contributing a new discovery backend
 
-You can contribute a new discovery backend to Swarm. For information on how to do this, see [our discovery README in the Docker Swarm repository](https://github.com/docker/swarm/blob/master/discovery/README.md).
+You can contribute a new discovery backend to Swarm. For information on how to do this, see <a href="https://github.com/docker/swarm/blob/master/discovery/README.md>our discovery README in the Docker Swarm repository</a>.
 
 ## Docker Swarm documentation index
 
--- a/docs/scheduler/strategy.md
+++ b/docs/scheduler/strategy.md
@@ -116,7 +116,10 @@ strategy prefers the node with most containers running.
 
 ## Docker Swarm documentation index
 
-- [User guide](https://docs.docker.com/swarm/)
-- [Discovery options](https://docs.docker.com/swarm/discovery/)
-- [Scheduler filters](https://docs.docker.com/swarm/scheduler/filter/)
-- [Swarm API](https://docs.docker.com/swarm/API/)
+
+- [User guide](/)
+- [Discovery options](/discovery/)
+- [Scheduler strategies](/scheduler/strategy)
+- [Scheduler filters](/scheduler/filter)
+- [Swarm API](/api/swarm-api)
+
--- a/docs/scheduler/filter.md
+++ b/docs/scheduler/filter.md
@@ -406,7 +406,7 @@ This filter will prevent scheduling containers on unhealthy nodes.
 
 ## Docker Swarm documentation index
 
-- [User guide](https://docs.docker.com/swarm/)
-- [Discovery options](https://docs.docker.com/swarm/discovery/)
-- [Scheduler strategies](https://docs.docker.com/swarm/scheduler/strategy/)
-- [Swarm API](https://docs.docker.com/swarm/API/)
+- [User guide](/)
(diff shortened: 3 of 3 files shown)


moxiegirl:
ping @vieux @aluzzardi @abronan Can y'all check this and merge if it ok with you?


author LGTM

chanwit:
LGTM

docs/discovery.md
```diff
@@ -208,7 +208,7 @@ And for the `nodes` discovery method.
 
 ## Contributing a new discovery backend
 
-You can contribute a new discovery backend to Swarm. For information on how to do this, see [our discovery README in the Docker Swarm repository](https://github.com/docker/swarm/blob/master/discovery/README.md).
+You can contribute a new discovery backend to Swarm. For information on how to do this, see <a href="https://github.com/docker/swarm/blob/master/discovery/README.md>our discovery README in the Docker Swarm repository</a>.
```
chanwit: I tried and this link is not working.

docs/scheduler/strategy.md
```diff
@@ -116,7 +116,10 @@ strategy prefers the node with most containers running.
 
 ## Docker Swarm documentation index
 
-- [User guide](https://docs.docker.com/swarm/)
-- [Discovery options](https://docs.docker.com/swarm/discovery/)
-- [Scheduler filters](https://docs.docker.com/swarm/scheduler/filter/)
-- [Swarm API](https://docs.docker.com/swarm/API/)
+
+- [User guide](/)
```
abronan: Tried and not working

docs/scheduler/strategy.md
```diff
@@ -116,7 +116,10 @@ strategy prefers the node with most containers running.
...
+- [Discovery options](/discovery/)
+- [Scheduler strategies](/scheduler/strategy)
+- [Scheduler filters](/scheduler/filter)
+- [Swarm API](/api/swarm-api)
```
abronan: Same for this one

moxiegirl:
@abronan are you building the docs?

abronan:
Oh! nope was just checking quickly and didn't realize :smile: 

LGTM then

chanwit:
Sorry for not understanding this. It means the links will be working fine after build, but not in the MD form on Github?

abronan:
@chanwit when you build the doc it creates the right links and it works. Although I agree that this can be confusing and that I would assume that the links on the `.md` files in the repository should be working the same way :confused:

chanwit:
@abronan @moxiegirl so should we fix the doc generator instead of hacking around `.md` files itself?

thaJeztah:
> so should we fix the doc generator instead of hacking around .md files itself?

I think that should be the ultimate goal; not sure if we can realize that in the short term (releases), but I definitely think we should try to get that fixed.

@moxiegirl do we have a list of issues / examples with links that break (i.e., what Hugo created, and what it should be?)

moxiegirl:
@chanwit Thank you for being concerned we really appreciate it. Our generator isn't broken.  @thaJeztah I know there have been comments about getting our Markdown to work as both GitHub repository pages and as a website. The pages in the `docs` directory aren't intended to as a website any more than the code you see there.  They have to be processed to become a website -- just like code is meant to built.

chanwit:
I see @moxiegirl thank you for explanation in details.
Now I understand that `.md` files under `docs` are not necessary to links correctly in its MD form. The website version takes priority.

aluzzardi:
Would it make sense **not** to name those `.md`? There's an implied understanding that `.md` files on a repo are supposed to be viewable on GitHub.

moxiegirl:
@aluzzardi The files are viewable on GitHub.  What people are assuming is that because they can open a file in a browser and view its content *as if it were HTML content* that the page should act as a web page.  I've done a survey of projects similar to ours in size/or popularity.  Ruby on Rails puts a note on each page.

https://github.com/rails/rails/blob/master/guides/source/3_0_release_notes.md

Right now, I'm just opting for educating users.

chanwit:
@moxiegirl This will solve the problem! It's a great idea to put a similar note on top of each page saying that users are expected to go to the rendered website.

moxiegirl:
@chanwit Yes it would solve this issue.  Any solution tho can't be limited to this repository. @aluzzardi and the other Docker project owners need to be on board with it.  Andrea let us know what you think...

chanwit:
LGTM

PR #1085, opened 2015-07-23
Improve build

jimmyxian:
Improve build
Implement : https://github.com/docker/swarm/issues/632  https://github.com/docker/swarm/issues/694

PR #1146, opened 2015-08-12
Add support images -a

jimmyxian:

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -185,7 +185,7 @@ func (e *Engine) RemoveImage(image *Image, name string) ([]*dockerclient.ImageDe
 
 // RefreshImages refreshes the list of images on the engine.
 func (e *Engine) RefreshImages() error {
-	images, err := e.client.ListImages(false)
+	images, err := e.client.ListImages(true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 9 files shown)


chanwit:
@jimmyxian It seems we need some integration tests for this PR :)

jimmyxian:
@chanwit Yeah..Forget to `git commit`. :)

cluster/swarm/cluster.go
```diff
@@ -298,7 +298,7 @@ func (c *Cluster) RemoveImages(name string) ([]*dockerclient.ImageDelete, error)
 	errs := []string{}
 	var err error
 	for _, e := range c.engines {
-		for _, image := range e.Images() {
+		for _, image := range e.Images(true) {
```
chanwit: Correct me if I am wrong. The default value is `false` so why `true` is used here?

scheduler/node/node.go
```diff
@@ -34,7 +34,7 @@ func NewNode(e *cluster.Engine) *Node {
 		Name:        e.Name,
 		Labels:      e.Labels,
 		Containers:  e.Containers(),
-		Images:      e.Images(),
+		Images:      e.Images(true),
```
chanwit: Here is also.

cluster/engine.go
```diff
@@ -185,7 +185,7 @@ func (e *Engine) RemoveImage(image *Image, name string) ([]*dockerclient.ImageDe
 
 // RefreshImages refreshes the list of images on the engine.
 func (e *Engine) RefreshImages() error {
-	images, err := e.client.ListImages(false)
+	images, err := e.client.ListImages(true)
```
chanwit: I am not sure why we need to `ListImages(true)` instead of the old `false`?

jimmyxian:
@chanwit https://github.com/docker/swarm/pull/1085#issuecomment-126522219
* `all=true` : get all images(including intermediate layers)
* `all=false` : get the images which have repotag

chanwit:
@jimmyxian Thank for clarifying.
LGTM

abronan:
LGTM

PR #1154, opened 2015-08-16
bump version to 0.5.0-dev

### LINKED ISSUE #1153: Proposal: versioning scheme
Currently `swarm` version is changed only when releasing. This is a bit confusing for someone to report issues. (For example, `swarm` version is still marked as 0.4.0 but we started committing 0.5.0 works into master.)

We should follow Docker's versioning scheme to help solve this problem. For example, 

 *  We are now in the new sprint (Sprint 22). At the beginning of a new sprint just after releasing, we should bump the version to `0.${x+1}.0-dev'.
 * After the latest RC, we should strip `-dev` out of the version number, just before tagging a new release.

I think this is a minimal change to help maintaining versions in a bit better way.

WDYT?

chanwit:
This PR fixes #1153 to support versioning scheme that switches between `release` (version containing only digits) and `-dev` (development versions).

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -2,7 +2,7 @@ package version
 
 var (
 	// VERSION should be updated by hand at each release
-	VERSION = "0.4.0"
+	VERSION = "0.5.0-dev"
 
 	// GITCOMMIT will be overwritten automatically by the build system
 	GITCOMMIT = "HEAD"


jimmyxian:
LGTM

chanwit:
ping @abronan 
As it's a bit changing the release process. I'm wait for comments @aluzzardi and @vieux too.

abronan:
LGTM (let's wait for @aluzzardi and @vieux's comments on this)

aluzzardi:
LGTM

PR #1155, opened 2015-08-17
prevent double logs with mesos

vieux:
`/create` with with Mesos backend will also start the container.

When you do a `docker run` the cli is doing `/create` and `/start`
For short lived containers, like `busybox ls` it's an issue because it will start the container another time.
You will see 2 `ls` in the logs

With this PR the start is handled by the cluster, it's not proxied anymore. So the mesos backend can choose to no start the container.


abronan:
@vieux Seems like the build is timing out for Mesos on Jenkins :confused:

chanwit:
@vieux both `janky` and my local regression hang after displaying this line:
`ok 89 mesos - docker commit`.

abronan:
LGTM

chanwit:
LGTM

vieux:
It's my fault but we shouldn't have merge this one, there is a bug in it, that's why `mesos - docker run short lived` never ends.

chanwit:
should we revert it then?

vieux:
@chanwit https://github.com/docker/swarm/pull/1192 :)

vieux:
again, it's my fault, I should have added the `mesos - docker run short lived` test in this PR instead of a new one.

PR #1166, opened 2015-08-25
Implement grouping list of images by Id

### LINKED ISSUE #1164: "docker images" de-duplication
Just realized `docker images` swarm returns duplicates of the same images on multiple nodes (b/c image was fetched using `docker pull`). Looks like bad UX and we can probably remove duplicates from the list. Thoughts?

image

chanwit:
This PR attempts to fix #1164. Currently, `docker images` list every duplicated image on every node. This is very verbose when the cluster grows.

Although, it's understandable that keeping the Engine info for each image is necessary when calling `docker images --filter`.

This PR is trying to address both problems.

  * To make `docker images` less verbose, this PR introduces an inner `map[string]*cluster.Image` to group images by Id.
  * To keep `docker images --filter node=` work, this PR does grouping after the filter.
  * ~~In addition, Engine information for the grouped images are still kept in the struct and therefore accessible by `/images/json`. Will be useful when a client is able to process them in the future.~~

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -123,8 +123,9 @@ func getImagesJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	accepteds, _ := filters["node"]
-	images := []*cluster.Image{}
-
+	// this struct helps grouping images
+	// but still keeps their Engine infos as an array.
+	groupImages := make(map[string]dockerclient.Image)
 	for _, image := range c.cluster.Images(all) {
 		if len(accepteds) != 0 {
 			found := false
@@ -138,9 +139,31 @@ func getImagesJSON(c *context, w http.ResponseWriter, r *http.Request) {
 				continue
 			}
 		}
-		images = append(images, image)
+
+		// grouping images by Id, and concat their RepoTags
+		if entry, existed := groupImages[image.Id]; existed {
+			entry.RepoTags = append(entry.RepoTags, image.RepoTags...)
+			groupImages[image.Id] = entry
+		} else {
+			groupImages[image.Id] = image.Image
+		}
 	}
 
+	images := []dockerclient.Image{}
+
+	for _, image := range groupImages {
+		// de-duplicate RepoTags
+		result := []string{}
+		seen := map[string]bool{}
+		for _, val := range image.RepoTags {
+			if _, ok := seen[val]; !ok {
+				result = append(result, val)
+				seen[val] = true
+			}
+		}
+		image.RepoTags = result
+		images = append(images, image)
+	}
 	w.Header().Set("Content-Type", "application/json")
 	json.NewEncoder(w).Encode(images)
 }
--- a/test/integration/api/tag.bats
+++ b/test/integration/api/tag.bats
@@ -46,6 +46,10 @@ function teardown() {
 	docker_swarm tag busybox tag_busybox:test
 
 	# verify
-	run docker_swarm images
-	[[ $(echo ${output} | grep -o "tag_busybox" | wc -l) == 2 ]]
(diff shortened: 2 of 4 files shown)


chanwit:
ping @ahmetalpbalkan @abronan @aluzzardi @jimmyxian @vieux

api/handlers.go
```diff
@@ -123,8 +123,12 @@ func getImagesJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	accepteds, _ := filters["node"]
-	images := []*cluster.Image{}
-
+	// this struct helps grouping images
+	// but still keeps their Engine infos as an array.
+	groupImages := make(map[string]*struct {
+		dockerclient.Image
+		Engines []*cluster.Engine
```
ahmetalpbalkan: Do we read .Engines anywhere?

chanwit: @ahmetalpbalkan  It will be serialized to JSON.

In `swarm` there's a way to keep Image/Engine info (the struct inside `cluster/image.go`). This `Engines` here is do the same. It's intended to not discard Image/Engine information.

ahmetalpbalkan: :+1:

jimmyxian:
LGTM

vieux:
@chanwit as engine isn't used by the client, I would remove it for now from the API.

chanwit:
@vieux if so, it will be easier to implement deduplication.

I'll improve the PR.

vieux:
@chanwit thanks

chanwit:
@vieux it's getting complicated than I thought.
I propose the assumption (actually a fact) that there are several image instances across a cluster.

If you agree, the `cluster.Image` should be changing to:
```
struct {
  dockerclient.Image
  []*Engine
}
```
With this, I think we could solve some fundamental structure towards the client management API.

WDYT?

api/handlers.go
```diff
@@ -123,8 +123,9 @@ func getImagesJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	accepteds, _ := filters["node"]
-	images := []*cluster.Image{}
-
+	// this struct helps grouping images
+	// but still keeps their Engine infos as an array.
+	groupImages := make(map[string]*cluster.Image)
```
vieux: I think `groupImages` could be a `map[string]*dockerclient.Image`

chanwit: @vieux fixing it. thank you.

test/integration/api/tag.bats
```diff
@@ -46,6 +46,10 @@ function teardown() {
 	docker_swarm tag busybox tag_busybox:test
 
 	# verify
-	run docker_swarm images
-	[[ $(echo ${output} | grep -o "tag_busybox" | wc -l) == 2 ]]
+	# change the way to verify tagged image on each node after image deduplication
+	run docker_swarm images --filter node=node-0
+	[[ $(echo ${output} | grep -o "tag_busybox" | wc -l) == 1 ]]
```
chanwit: @jimmyxian could you please help me review this part a bit more?

chanwit:
ping @vieux 
(the build failed because a mesos test case is timed out.)

ping @jimmyxian for reviewing tag-related test cases again.

vieux:
@chanwit yes see my comment in https://github.com/docker/swarm/pull/1155

chanwit:
ping @vieux

test/integration/api/tag.bats
jimmyxian: LGTM

jimmyxian:
@chanwit The test cases are LGTM

vieux:
LGTM

PR #1171, opened 2015-08-26
update docker daemon initialization to use subcommand vs -d flag

MHBauer:
this is kind of a nit change, but there was a switch from `docker -d` to `docker daemon` in base docker.

### DIFF
--- a/docs/install-manual.md
+++ b/docs/install-manual.md
@@ -55,8 +55,9 @@ This example uses the Docker Hub based `token` discovery service. Log into **eac
 
 1. Start the Docker daemon with the `-H` flag. This ensures that the Docker remote API on *Swarm Agents* is available over TCP for the *Swarm Manager*.
 
-		$ docker -H tcp://0.0.0.0:2375 -d
- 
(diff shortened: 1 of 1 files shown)


abronan:
Thanks @MHBauer! Yes we should stay in line with the docker engine doc ;) LGTM

thaJeztah:
Yes, LGTM (but be aware that this will only work on docker 1.8; not sure what the current requirements for Swarm are)

abronan:
@thaJeztah Good point! We should probably add a note instead saying that `-d` is deprecated in favor of `daemon` on engines > 1.8 then?

thaJeztah:
Might be worth adding. In Docker we only document the current version, but for Swarm it could well be that users are still on a slightly older version.

MHBauer:
Change it back and add a note?
Or keep the changes and add a note like:
"versions of docker prior to 1.8 used the deprecated `-d` flag instead of the `docker daemon` subcommand."

chanwit:
LGTM

thaJeztah:
thanks @MHBauer!

PR #1178, opened 2015-08-29
Add reliance on external internet connectivity

MHBauer:
I'm not sure this is the best way or place to say this. It needs to be said somewhere. Suggestions welcome!

There are two nearly identical copies of the discovery docs. Not sure if that is intentional.

### DIFF
--- a/discovery/README.md
+++ b/discovery/README.md
@@ -10,7 +10,18 @@ Docker Swarm comes with multiple Discovery backends.
 
 ## Backends
 
+You use a hosted discovery service with Docker Swarm. The service
+maintains a list of IPs in your swam. There are several available
+services, such as `etcd`, `consul` and `zookeeper` depending on what
+is best suited for your environment. You can even use a static
+file. Docker Hub also provides a hosted discovery service which you
+can use.
+
 ### Hosted Discovery with Docker Hub
+
+This example uses the hosted discovery service on Docker Hub. Using
+Docker Hub's hosted discovery service requires that each node in the
+swarm is connected to the internet. To create your swarm:
 
 First we create a cluster.
 
--- a/docs/discovery.md
+++ b/docs/discovery.md
@@ -15,7 +15,18 @@ Docker Swarm comes with multiple Discovery backends.
 
 ## Backends
 
+You use a hosted discovery service with Docker Swarm. The service
+maintains a list of IPs in your swam. There are several available
+services, such as `etcd`, `consul` and `zookeeper` depending on what
+is best suited for your environment. You can even use a static
+file. Docker Hub also provides a hosted discovery service which you
+can use.
+
 ### Hosted Discovery with Docker Hub
+
+This example uses the hosted discovery service on Docker Hub. Using
+Docker Hub's hosted discovery service requires that each node in the
+swarm is connected to the internet. To create your swarm:
 
 First we create a cluster.
 


thaJeztah:
SGTM, maybe it should be in a `> **Note:** ....`, but it's better to wait with changing until others have commented

vieux:
ping @moxiegirl

discovery/README.md
```diff
@@ -12,6 +12,9 @@ Docker Swarm comes with multiple Discovery backends.
 
 ### Hosted Discovery with Docker Hub
 
+This service is hosted by Docker Hub and requires all nodes to have
```
moxiegirl: @vieux "This service" is ambiguous -- do you mean Docker Swarm?  We haven't really said it was a service anywhere else.  So, maybe  "The Docker Swarm image is hosted on Docker Hub. Using the Swarm image requires all the nodes to have internet access"

MHBauer: The token discovery service. Should it say that instead?

thaJeztah: The "service" referred to here is the "Hosted Discovery (service)", which is hosted on Docker Hub (but perhaps "Docker Hub" should be left out here, because that's usually associated with the Docker Hub *Registry*)

@vieux is there an "official" / vetted name for the Discovery Service?

moxiegirl:
@MHBauer The README is just intended for information that doesn't go in the docs. For example, tips for contributors on how to mod the feature. We need to remove the material from the README that duplicates what is in the user documentation Let me know if you want to do that; otherwise we will carrry.

@thaJeztah I've confirmed with @vieux that the  "hosted discovery service" is simply a list of IPs.  It is lower case because this is a reference to a generic service.  From our website:
image

discovery/README.md
moxiegirl: @MHBauer try this (ping @vieux for tech check) It is wordier but I think it is clearer

---> https://gist.github.com/moxiegirl/fda81731b94925ff1fef

You use a hosted discovery service with Docker Swarm. The service maintains a list of IPs in your swam.  There  are several commercial services `etcd`, `consul` and `zookeeper` depending on what is best suited for your environment. You can even use a static file.  Docker Hub also provides a hosted discovery service which you can use.  

This example uses the hosted discovery service on Docker Hub. Using Docker Hub's hosted discovery service requires that each node in the swarm is connected to the internet.  To create your swarm:

MHBauer:
fixed up, I moved the first paragraph up a section because it's a more generic take on discovery services in general vs the specific discovery service of that section.

discovery/README.md
```diff
@@ -10,8 +10,18 @@ Docker Swarm comes with multiple Discovery backends.
 
 ## Backends
 
+You use a hosted discovery service with Docker Swarm. The service
+maintains a list of IPs in your swam. There are several commercial
```
thaJeztah: Hm, re-reading this now, but don't think these are *commercial* services (I.e., the software is open-source, there probably are commercial hosted solutions *using* that software)

moxiegirl: @thaJeztah crap! that is a good point. Sorry @MHBauer LGTM when this change is made. We will merge!

MHBauer: tried to rewrite it, but I feel like that sentence reads kind of funky now.

discovery/README.md
```diff
@@ -10,8 +10,19 @@ Docker Swarm comes with multiple Discovery backends.
 
 ## Backends
 
+You use a hosted discovery service with Docker Swarm. The service
+maintains a list of IPs in your swam. There are several available
+services, such as `etcd`, `consul` and `zookeeper` depending on what
```
thaJeztah: Perhaps use "solutions" in stead of "services"? @moxiegirl wdyt?

moxiegirl: @thaJeztah I see your point but really it is splitting hairs...I'd rather merge it then send Morgan back for such a fine grained change.  LGTM

thaJeztah: sgtm; just adding possible alternatives because of https://github.com/docker/swarm/pull/1178#issuecomment-136859981 :smile:

thaJeztah:
LGTM thanks @MHBauer !

thaJeztah:
ping @vieux for the merge

vieux:
LGTM

discovery/README.md
MHBauer: I have no problem making small changes. I would rather have it right, than quick. Sorry if I move too slow some times.

thaJeztah: @MHBauer you're not moving slow, you're doing excellent work! It's sometimes hard to find the right words, but (as a wise girl once told me): "perfect is the enemy of good" :smile:

So sometimes its better to get things merged, we can always improve in another PR.

PR #1186, opened 2015-09-01
fix gitignore to specifically ignore the binary called swarm in the root directory

MHBauer:
The current gitignore will not see any new files added to cluster/swarm directory. It will see changes to existing files.

This caused a problem for me, when using a search tool called "ag" that takes into account the .gitignore present in a git repository.

I *think* the original intention was to exclude go generated `swarm` binary in the projects root.

Unfortunately we can't ignore based on file type, only on a matching name.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,1 @@
-swarm
+/swarm


MHBauer:
Looks like that was the original intention of b2fa0609571a45fa01b116fe4f6dc19041f091c7, but all that did was add a newline.

vieux:
LGTM, ping @abronan @aluzzardi

abronan:
LGTM

aluzzardi:
LGTM

PR #1188, opened 2015-09-01
switch to golang 1.5

vieux:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - 1.4
+  - 1.5
 
 # let us have speedy Docker-based Travis workers
 sudo: false
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.4
+FROM golang:1.5
 
 COPY . /go/src/github.com/docker/swarm
 WORKDIR /go/src/github.com/docker/swarm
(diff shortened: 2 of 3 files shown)


chanwit:
You did it!!
LGTM

vieux:

aluzzardi:
I think there are other places.

travis?

aluzzardi:
And `test/integration/Dockerfile`

abronan:
LGTM we should be careful about the first failure on leadership tests though.. might happen again. Might be worth re-running the tests.

vieux:
ping @aluzzardi good catch, I added them.

jimmyxian:
@vieux Do we need update README.md? :)
> Beginning with Swarm 0.4 golang 1.4.x or later is required for building Swarm. Refer to the Go installation page to download and install the golang 1.4.x or later package.

vieux:
@jimmyxian good point

MHBauer:
Is go1.5 required? Should travis have a 1.5 build in addition to a 1.4.2 build? 

I am still unfamiliar with the go ecosystem, so I am not sure of the pros and cons. I come from a Java background where point releases are huge and happen with years between them.

This more of a question of curiosity rather than a suggestion.

vieux:
@MHBauer go 1.x will always be backward compatible, no breaking changes.
We decided not to test 1.4.2 because, at some point we might drop the 1.4.2 support if we use features that are only in 1.5

Performances and cross compilation are much better in 1.5

Generally, we update to the latest version of go as soon as it's stable.

vieux:
@jimmyxian I don't think we need to update the Readme yet, as 1.5 is not required, so 1.4.x or later is still valid.

abronan:
Ok merging then ;) we'll change the doc in another PR if required

jimmyxian:
@vieux OK

Issue #1185, opened 2015-09-01
Manager does not retry for failed nodes (token:// discovery)


ahmetalpbalkan:
version: 0.4.0

When Swarm manager starts (using discovery `token://`), if the node in the list is down, the manager flags it as dead **indefinitely** and does not pick it up again when the node comes afterwards.

This also happens after the Swarm manager was started. Once a node dies, the engine logs `Removed engine %s` and even though engine comes up afterwards, in the `--debug` logs the only `Updated engine state` belongs to the node that never failed.

Some logs... Around 05:36:20 I took node `alp` down and brought up 2 minutes later. But it never showed up in the logs nor in the `/info`... But when I started another manager container that targets the same token://, it was able to pick up the node died earlier. 

Hence, the manager is most definitely not retrying for failed nodes. It should be more tolerant to faults.

```
time="2015-09-01T05:36:15Z" level=debug msg="Updated engine state" id="VJEO:U3TS:HJ7T:IKNS:JHJP:YTOW:UVLP:NUMC:RSA7:U42X:OMNC:QCYW" name=hellodocker
time="2015-09-01T05:36:16Z" level=info msg="Registered Engine hellodocker at hellodocker.REDACTED:xxxx"
time="2015-09-01T05:36:16Z" level=debug msg="Updated engine state" id="2MJA:DPJM:6ACB:GUN2:G4VK:GZRT:NNMS:JUYJ:IEYB:A5AP:HAMX:XCDA" name=alp
time="2015-09-01T05:36:16Z" level=info msg="Registered Engine alp at alp.co:xxxx"
time="2015-09-01T05:36:34Z" level=info msg="Removed Engine alp"
time="2015-09-01T05:36:46Z" level=debug msg="Updated engine state" id="VJEO:U3TS:HJ7T:IKNS:JHJP:YTOW:UVLP:NUMC:RSA7:U42X:OMNC:QCYW" name=hellodocker
time="2015-09-01T05:37:16Z" level=debug msg="Updated engine state" id="VJEO:U3TS:HJ7T:IKNS:JHJP:YTOW:UVLP:NUMC:RSA7:U42X:OMNC:QCYW" name=hellodocker
time="2015-09-01T05:37:46Z" level=debug msg="Updated engine state" id="VJEO:U3TS:HJ7T:IKNS:JHJP:YTOW:UVLP:NUMC:RSA7:U42X:OMNC:QCYW" name=hellodocker
[...]

Issue #1136, opened 2015-08-08
Expose un-healthy nodes in info


aluzzardi:
`docker info` should expose un-healthy nodes along with an error message.

Pseudo-Example:
```
Nodes: 1
 dev: 10.0.0.123:4242
  └ Status: Down
  └ Error: 10.0.0.123: No route to host
```

Thoughts?

abronan:
:+1: Good idea as it helps troubleshoot connectivity issues to Swarm nodes. Most of the time we would think it's a Swarm bug (in discovery for example) while in reality it might just be a misconfigured firewall or an unexposed port.

chanwit:
It's a great step towards cluster management APIs that you have mentioned earlier.

UserTaken:
When using a token with dozens of registered machines, what if you just want to show the active nodes instead of every entry? When using a backend such as Etcd, inactive nodes are automatically removed after a set TTL so how would this work with that?

jimmyxian:
:+1: More easy to find what's wrong.

aluzzardi:
Merged into the node management proposal (#1486)

PR #1195, opened 2015-09-03
first crack at solving gh 1185

MHBauer:
This works, but needs to be cleaned up. I left my debug logging in so anyone trying to repro doesn't have to do it themself. Please tag what you think should stay vs be removed. 

How does it look otherwise?

A similar change as in token.go needs to be made for the file and node versions of discovery. I am not sure about the other ways, like etcd, zookeeper, consul, etc.

unit test for file expects it to update immediately, but it could have the old value until it updates. I added a loop, but it needs to timeout with a failure. I need to learn how to use timers in go.

nodes.go only ever sends it's list once, and then exits. Not sure if adding an infinite loop there is wanted or if it should be left alone.

The logic for repeating needs to either be pushed down, or nodes that are fail to be added need to be pushed back upwards to the individual discovery implementations so that they can suggest retrying (or not).

for #1185

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -155,10 +155,11 @@ func (c *Cluster) hasEngineByAddr(addr string) bool {
 	return c.getEngineByAddr(addr) != nil
 }
 
-func (c *Cluster) addEngine(addr string) bool {
+func (c *Cluster) addEngine(addr string, done chan bool, fail chan string) {
 	// Check the engine is already registered by address.
 	if c.hasEngineByAddr(addr) {
-		return false
+		fail <- addr
+		return
 	}
 
 	// Attempt a connection to the engine. Since this is slow, don't get a hold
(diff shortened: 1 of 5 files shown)


vieux:
@MHBauer are you able to write an integration test ?

MHBauer:
Have since figured it out. I'll give it a shot now. Need to write a test that fails in master and passes here.

jimmyxian:
@MHBauer Thanks for contributing this.
I think we have to take this proposal(https://github.com/docker/swarm/issues/1136) into consideration. 
Maybe we need to save the error node, and mark it `unhealth`, and then loop connect util it's OK.

/cc @vieux @abronan

MHBauer:
It's a lot more difficult than I thought to right some integration tests. I don't think I can use the helper functions to set things up.

The test needs to run join before starting the engine, and the helper functions require engines before calling join. Then the engines need to be started. I'm not confident yet, but I'll keep thinking about it.

vieux:
@aluzzardi @abronan can you please take a look ? Thanks

abronan:
SGTM, code needs a little cleanup though so we can review properly.

aluzzardi:
@MHBauer I'd like to get your input on #1486 and if you think it could be a better approach for this problem

MHBauer:
Agree with your assessment that this *fix* splits up the responsibilities too much. It's been sitting here too long anyway. If we're going to rewrite it, let's do it.

Issue #1189, opened 2015-09-02
Support volume API


bfirsh:
Engine 1.9 will have a new volumes API, which Swarm should probably support in some way. It is perhaps not practical to support local volumes, but it could be pretty interesting to support network storage with volume drivers.

https://github.com/docker/docker/pull/14242

/cc @cpuguy83 @lukemarsden

lukemarsden:
Hey @bfirsh - thanks for the ping! @vieux could we do a hangout to discuss what you have in mind for this, and how it can fit into the volume plugins model?

vieux:
@lukemarsden let's keep the discussion here, so it's open to everyone who wants to participate.

vieux:
What do you think of this naive approche to start with, we do the same thing as we do with images.

`GET /volumes`: we do kind of like images, we display volumes for all the machines, without duplication (like https://github.com/docker/swarm/pull/1166)
`GET /volumes/{name:.*}`: we forward to *a* node that has the volume and return the volume (like inspect for images)
`POST /volumes`: we do it on every node, like for pull
`DELETE /volumes/{name:.*}`: we do it on every node, like rmi

thaJeztah:
> POST /volumes: we do it on every node, like for pull
> DELETE /volumes/{name:.*}: we do it on every node, like rmi

One thing to take into account here is that reference counting is performed *per daemon*; What if a plugin offers centralized storage (I.e,, "volume A" on each node actually points to the *same* volume), and that volume is no longer referenced on a node? In that case, the daemon on *that* node will proceed, and remove/delete the volume, whereas it can still be referenced on *another* node.

I also raised this issue that may need some attention in this respect; https://github.com/docker/docker/issues/15997

myechuri:
```
 GET /volumes: we do kind of like images, we display volumes for all the machines, without duplication (like #1166)
GET /volumes/{name:.*}: we forward to a node that has the volume and return the volume (like inspect for images)
POST /volumes: we do it on every node, like for pull
DELETE /volumes/{name:.*}: we do it on every node, like rmi
```

This sounds like duplication of a lot of work done in Flocker. Instead of Swarm reimplementing volume driver's control service functionality, would it be more efficient to make these Swarm calls passthrough to volume driver's control service (like Flocker control service), since the volume driver has greater visibility into backend types (shared/DAS/etc)), therefore dealing with scenarios like @thaJeztah outlined correctly?

Thanks!

vieux:
[...]

PR #1196, opened 2015-09-04
add support for docker volume ls & inspect

vieux:
Related to #1189

Add support for `docker volume ls` & `docker volume inspect`

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -104,7 +104,7 @@
 		},
 		{
 			"ImportPath": "github.com/samalba/dockerclient",
-			"Rev": "a87700686584dc383ddb7e2bfb042e0f7a8c371b"
+			"Rev": "77b723e2c0d07a21cc9483dc6108f95bd403b576"
 		},
 		{
 			"ImportPath": "github.com/samuel/go-zookeeper/zk",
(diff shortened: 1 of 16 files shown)


chanwit:
@vieux are you going to have `/volumes/json` similar to `/images/json` in this PR?

vieux:
@chanwit no because unfortunately, the engine didn't respect the same convention /cc @icecrime @cpuguy83

chanwit:
@vieux also does `volume ls` support `--filter`? 
it's going to be similar to `images` when we'd like to filter by a node name (and maybe in another PR).

/cc @icecrime @cpuguy83

cpuguy83:
@vieux We can have it follow the same convention, I think that makes sense.  @icecrime ?

vieux:
@cpuguy83 @icecrime as you want, I know the `/json` is ugly and not rest like (especially because we never did the `/xml`) but it would follow the same convention

vieux:
@chanwit I think you should open an issue on https://github.com/docker/docker to propose --filter on volume

vieux:
I think this one is good to merge as is as a 1st step /cc @abronan @chanwit @jimmyxian

abronan:
:+1: LGTM

chanwit:
LGTM

PR #1201, opened 2015-09-07
Fix token discovery Watch on failure

### LINKED ISSUE #1200: swarm manage with token discovery hangs after errors
hang at 88 line:

https://github.com/docker/swarm/blob/6678a167fe873a320d00aa99be09945ac5f9ca96/discovery/token/token.go#L88

Trace:
```
goroutine 7 [chan send, 7 minutes]:
github.com/docker/swarm/discovery/token.func·001()
        /root/gocode/src/github.com/docker/swarm/discovery/token/token.go:88 +0x12c
created by github.com/docker/swarm/discovery/token.(*Discovery).Watch
        /root/gocode/src/github.com/docker/swarm/discovery/token/token.go:113 +0x226

```
This problem is appears when we run 'swarm manage' with temporary inaccessible internet, after access is returned the swarm will not retrieve node list, but swarm join work correctly.

abronan:
fixes #1200 

The error chan was not passed back to the caller, thus the hang in token discovery because the `select` clause in `swarm/cluster.go` -> `monitorDiscovery()` would wait on a nil chan (also explaining why the successive joins were succeeding).

Tested by savagely unplugging my ethernet cable and plugging it back after some time. #fun

### DIFF
--- a/discovery/token/token.go
+++ b/discovery/token/token.go
@@ -112,7 +112,7 @@ func (s *Discovery) Watch(stopCh <-chan struct{}) (<-chan discovery.Entries, <-c
 		}
 	}()
 
-	return ch, nil
+	return ch, errCh
 }
 
 // Register adds a new entry identified by the into the discovery service


jimmyxian:
LGTM

chanwit:
LGTM

Issue #1090, opened 2015-07-27
manage --rootdir argument is not documented


ahmetalpbalkan:
The `--rootdir` argument below has no documentation. Looks confusing.

```
$ docker run swarm:0.4.0-rc1 manage -h
Usage: swarm manage [OPTIONS] <discovery>

Manage a docker cluster

Arguments:
   <discovery>    discovery service to use [$SWARM_DISCOVERY]
                   * token://<token>
                   * consul://<ip>/<path>
                   * etcd://<ip1>,<ip2>/<path>
                   * file://path/to/file
                   * zk://<ip1>,<ip2>/<path>
                   * <ip1>,<ip2>

Options:
   --rootdir "/.swarm"
   --strategy "spread"						placement strategy to use [spread, binpack, random]
   --filter, -f [--filter option --filter option]		filter to use [affinity, health, constraint, port, dependency]
   --host, -H [--host option --host option]			ip/socket to listen on [$SWARM_HOST]
   --replication						Enable Swarm manager replication
   --advertise, --addr 						Address of the swarm manager joining the cluster. Other swarm manager(s) MUST be able to reach the swarm manager at this address. [$SWARM_ADVERTISE]
   --tls							use TLS; implied by --tlsverify=true
   --tlscacert 							trust only remotes providing a certificate signed by the CA given here
[...]
```

thaJeztah:
Interested in making a PR? :smile:

ahmetalpbalkan:
I'm not sure what it is abou @thaJeztah  :)

thaJeztah:
Ah, ok, apologies in that case, I thought you did :+1:

aluzzardi:
Indeed :) I think we are going to deprecate it soon enough.

MHBauer:
@aluzzardi this is fixed by #1205 and can be close.

jimmyxian:
@MHBauer Thanks :+1:

PR #1205, opened 2015-09-09
remove deprecated unused flag

MHBauer:
for #1090

The variables aren't referenced anywhere, and aren't retrieved from the cli. Seemed easy and safe to remove.

It does turn providing that flag into a fatal error, so it would break anyone who was using it, vs silently succeeding and doing nothing.

Not sure if this PR is worth it or not right now, but it is available.

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -22,7 +22,6 @@ var (
 			ShortName: "m",
 			Usage:     "Manage a docker cluster",
 			Flags: []cli.Flag{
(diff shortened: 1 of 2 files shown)


abronan:
Thanks @MHBauer, forgot about those flags during the store cleanup :smile: 

LGTM

vieux:
LGTM

MHBauer:
fixes #1090

Issue #1151, opened 2015-08-15
"docker pull" returning 0 on error


tchk1228:
Kernel: 3.10.0-229.el7.x86_64
OS: CentOS Linux release 7.1.1503 (Core)

Swarm Manager:

docker -H $CLUSTER_ENDPOINT version
 Client version: 1.5.0
 Client API version: 1.17
 Go version (client): go1.4.1
 Git commit (client): a8a31ef
 OS/Arch (client): linux/amd64
 Server version: swarm/0.4.0
 Server API version: 1.16
 Go version (server): go1.4.2
 Git commit (server): HEAD

Node 1:
 Server version: 1.6.2
 Server API version: 1.18
 Go version (server): go1.4.2
 Git commit (server): ba1f6c3/1.6.2
 OS/Arch (server): linux/amd64

Node 2:
 Server version: 1.6.0
 Server API version: 1.18
 Go version (server): go1.4.2
 Git commit (server): 8aae715/1.6.0
 OS/Arch (server): linux/amd64

docker -H $ENDPOINT pull $REGISTRY/qwerty; echo $?
NODE1: Pulling REGISTRY_IP/qwerty:latest... : Error: image qwerty:latest not found
NODE2: Pulling REGISTRY_IP/qwerty:latest... : Error: image qwerty:latest not found
0

jimmyxian:
@tchk1228 Thanks for reporting this. I will take a look..

jimmyxian:
Done to confirm. `pull / load / import` all these three cmd always return 0 even errors.

chanwit:
@jimmyxian It seems we need a PR then.

jimmyxian:
@chanwit Yeah. I'm trying to slove this locally. :)

vieux:
@jimmyxian @chanwit any news on this one ?

jimmyxian:
@vieux Yeah, here is my idea:
Once pull/load/import error on some nodes, do not return util all nodes done.
Then collect all the errors in swarm and finally return all the error to client. 

WDYT?

jimmyxian:
Close via #1206 #1213

PR #1206, opened 2015-09-10
Fix pull return code

jimmyxian:
Fix: https://github.com/docker/swarm/issues/1151
If code logic is OK,  i will fix `import` and `load` in the same way in a separate PR.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -296,7 +296,7 @@ func (c *Cluster) RemoveImages(name string, force bool) ([]*dockerclient.ImageDe
 }
 
 // Pull is exported
-func (c *Cluster) Pull(name string, authConfig *dockerclient.AuthConfig, callback func(where, status string)) {
+func (c *Cluster) Pull(name string, authConfig *dockerclient.AuthConfig, callback func(where, status string, err error)) {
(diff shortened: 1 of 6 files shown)


cluster/swarm/cluster.go
```diff
@@ -307,14 +307,14 @@ func (c *Cluster) Pull(name string, authConfig *dockerclient.AuthConfig, callbac
...
 
 			if callback != nil {
-				callback(engine.Name, "")
+				callback(engine.Name, "", false)
 			}
 			err := engine.Pull(name, authConfig)
 			if callback != nil {
 				if err != nil {
-					callback(engine.Name, err.Error())
+					callback(engine.Name, err.Error(), true)
```
vieux: maybe the callback could take `status string, err error` so we call

`callback(engine.Name, "", err.Error)` and `callback(engine.Name, "downloaded", nil)`

What do you think ?

jimmyxian: Yeah, Good idea!

jimmyxian:
ping @vieux 
Check again, thanks.

abronan:
LGTM

vieux:
LGTM

PR #1209, opened 2015-09-10
try to use gnu version of readlink if available

MHBauer:
On osx the integration script will fail if it is not run from /swarm/test/integration directory.

1. osx has bsd readlink, which doesn't have the -f flag. By pulling it out into separate subshell, it fails the script early.
2. osx users (and others) tend to also install the gnu version of tools (using `brew install coreutils` or some such) but with a `g` prefix, so try to find and use that version before using a different version.

tested on osx and a linux vm.

^ I don't know if you want this information in the commit message, or if it's good enough to have it on a github PR.

### DIFF
--- a/test/integration/run.sh
+++ b/test/integration/run.sh
@@ -1,7 +1,11 @@
 #!/usr/bin/env bash
 set -e
 
-cd "$(dirname "$(readlink -f "$BASH_SOURCE")")"
+# try to use gnu version of readlink on non-gnu systems (e.g. bsd, osx)
+# on osx, install with 'brew install coreutils'
+READLINK_LOCATION=$(which greadlink readlink | head -n 1)
+THIS_SCRIPT=$(${READLINK_LOCATION} -f "$BASH_SOURCE")
+cd "$(dirname "${THIS_SCRIPT}")"
 
 # Root directory of Swarm.
 SWARM_ROOT=$(cd ../..; pwd -P)


tianon:
Oh, interesting solution; seems reasonably sane to me! :+1:

MHBauer:
I'm not bash expert, so if there's a better way to go about achieving this, I'd like to learn.

tianon:
There are more verbose ways of accomplishing this (replacing error messages with something more helpful for how to "fix" the problem, for example), but IMO this is a really, _really_ simple way since I'd imagine the swarm maintainers aren't really interested in having their tooling scripts get more complicated than they really need to be.  Perhaps a comment noting that `brew install coreutils` is the way to fix this for OSX would help, but I'd also note that I'm not a maintainer here, was just asked by them to take a look and give my thoughts. :smile: :+1:

abronan:
thanks @tianon! :smile: 

LGTM :+1: 

Even though it's self-explanatory, I would probably just add a single comment there to explain why we do this.

vieux:
LGTM after @MHBauer adds a comment as @abronan mentioned.

MHBauer:
good? or too verbose?
I'm trying not to make any assumptions about the dev userbase, but I 
figure after native linux, osx is probably the next most popular, so it 
merits instructions.

MHBauer:
Like @tianon said, I did not want to overcomplicate things. I've seen this *trick* done elsewhere for projects that need to be basically runable on bash, and probably have a version of the gnutools available. I don't think any of us want to bring autotools into this...

abronan:
Thanks for adding the comment @MHBauer!

PR #1210, opened 2015-09-13
Test commit to fix CI failure

mrjana:
This PR fixes a race in the port filter test case where if two containers are started quickly in succession before the first container has been completely cleaned up, the second container may fail to start because we share the host port space and the port might be still in use. Fixed it by introducing a 1 second delay in between container starts.

Signed-off-by: Jana Radhakrishnan <mrjana@docker.com>

### DIFF
--- a/test/integration/port-filters.bats
+++ b/test/integration/port-filters.bats
@@ -34,7 +34,10 @@ function teardown() {
 
 	run docker_swarm run -d --expose=80 -p 80:80 busybox sh
 	[ "$status" -eq 0 ]
+	# Sleep for 1 second to make sure the port 80 is available again
+	sleep 1
 	run docker_swarm run -d --expose=80 -p 80:80 busybox sh
+	echo $output
 	[ "$status" -eq 0 ]
 
 	# When trying to start the 3rd one, it should be error finding port 80.


chanwit:
@mrjana you won't see any output if the test passed. Try to assert `$status` with a strange value :)

abronan:
Jenkins frozen for now :/

mrjana:
Yeah been waiting for this to get scheduled just to get a small data point :-(

abronan:
@mrjana CI running ;)

mrjana:
ping @vieux

abronan:
Thanks for investigating @mrjana, although I'm curious to know the real reason on why this was working without the sleep until recently, was there any change on `libnetwork`'s side that might have impacted this? :confused:

mrjana:
@abronan Yes there was networking related change in docker(not libnetwork). We made `--userland-proxy` to be disabled by default in docker so while earlier the port binding was happening from separate proxy process when `--userland-proxy` was enabled by default, now this port binding and subsequent cleanup happens in the docker daemon itself. This has resulted in a change in timing when exactly the port gets cleaned up. While I don't know exactly what timing changed in the docker control flow, there is race in the swarm test case which is becoming more prominent with current docker master.

abronan:
Well I guess we should go forward with this even though I would add a placeholder `FIXME` just to make sure we keep track of this and also to understand why this happens with `--userland-proxy` disabled.

mrjana:
@abronan The most deterministic way to fix this would be to check the first containers status through inspect to make sure it's stopped and then start the second container. If the race is not clear, I can explain it to you in person. Trying to track down a timing change because a process is not getting spawned any more can get very hard with the need to reason it out in the context of how kernel schedules the process and how go runtime schedules the go routines.

abronan:
I mean yes, the port release is delayed by the added overhead involved in disabling `--userland-proxy`. The `FIXME` is just to make sure we check this cleanly instead of doing a sleep. Or if we do this now that'd be better I guess.

mavenugo:
@abronan fyi for other reasons : https://github.com/docker/docker/pull/16347 temporarily. Shall we check if swarm tests are back to normal ?

mrjana:
Regardless we need to fix the race in the test case. But now that there is no rush to push this fix in let's do the deterministic way. I will update the PR with a better fix.

Issue #786, opened 2015-05-14
Overloaded swarm node breaks the rest of the cluster


zstarer:
Hi guys -

We've been experimenting (A LOT) with swarm lately. One issue we've just discovered is that overloading any particular node in the cluster can break certain functionality expected from the master.

On one node, we had ~150 containers running , and can no longer hit the docker daemon (on that node) to see what's running (I think there's just no resources left) however, we can still hit the swarm master to return some data:

 ip-10-0-.ec2.internal: 10.0.:2375
  └ Containers: 169
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 2.734 GiB / 15.69 GiB
 ip-10-0-.ec2.internal: 10.0.:2375
  └ Containers: 35
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 800 MiB / 15.69 GiB
 ip-10-0-.ec2.internal: 10.0.:2375
  └ Containers: 35
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 1.562 GiB / 15.69 GiB
 ip-10-0-.ec2.internal: 10.0.:2375
  └ Containers: 35
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 3.125 GiB / 15.69 GiB
 ip-10-0-.ec2.internal: 10.0.:2375
  └ Containers: 65
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 2.344 GiB / 15.69 GiB

running a 'docker ps -a' from the master also returns data..however, we cannot run commands that then impact -any- slave in our pool.

imagine container1 was running on the node with 169 containers, and container2 is elsewhere.
I could not run 'docker rm -f container1' or 'docker rm -f container2'

please let me know if there are already open items for anything similar, I'd like to read more about this.

aluzzardi:
Thanks for reporting this, @zstarer.

There must be a deadlock somewhere making the `Cluster` object freeze.

We are going to reproduce the issue in our integration tests so we can fix it and make sure it stays fixed.

aluzzardi:
So this is related to parallel scheduling which we are going to address in the next release

abronan:
This issue should be fixed now with #1261. Closing this one. Thanks and let us know if you encounter any more issue with master.

PR #1212, opened 2015-09-15
WIP: Remove the lock at the scheduler level

### LINKED ISSUE #1194: Swarm has deadlock with the scheduler
I think swarm has a deadlock with the scheduler if a request is lost between swarm master and the docker daemon.
```
$ DOCKER_HOST=localhost:2375 docker info
Containers: 31
Images: 60
Role: primary
Strategy: spread
Nodes: 3
 localhost.localdomain: 192.168.50.1:2371
  └ Containers: 11
  └ Reserved CPUs: 0 / 8
  └ Reserved Memory: 0 B / 16.23 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=4.1.6-200.fc22.x86_64, operatingsystem=Fedora 22 (Twenty Two), storagedriver=overlay
 node1: 192.168.50.5:2371
  └ Containers: 10
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 0 B / 1.019 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.13.0-24-generic, operatingsystem=Ubuntu 14.04 LTS, storagedriver=aufs
 node2: 192.168.50.6:2371
  └ Containers: 10
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 0 B / 1.019 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.13.0-24-generic, operatingsystem=Ubuntu 14.04 LTS, storagedriver=aufs
CPUs: 12
Total Memory: 18.27 GiB
Name: 38dc57b17c85

$ DOCKER_HOST=localhost:2375 docker version
Client:
 Version:      1.8.1
 API version:  1.20
 Go version:   go1.4.2
 Git commit:   d12ea79
 Built:        Thu Aug 13 02:35:49 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      swarm/0.4.0
 API version:  1.16
 Go version:   go1.4.2
 Git commit:   d647d82
 Built:        
 OS/Arch:      linux/amd64
```

https://gist.github.com/aanm/cd073fe77f74d63732e4

ping @abronan 

EDIT: This occurs with replication mode

abronan:
fixes #1194 #786 

This PR removes the `Lock` at the `scheduler` level and moves it to `node.Node` for better granularity. `cluster` now maintains a list of Nodes with resource accounting protected by a `Lock` which avoids races for resource reservation. Processes not lucky enough to get a resource slice on a machine that gets full in the meantime are retrying the process of scheduling unless they get the resources they need (it fails if there is no more resource available cluster-wide).

This is I think the most simple approach to remove the `lock` at the scheduler level for now and avoid the races. Note that this does not solve the naming issue (we should settle if we allow duplicate names and solve the ambiguity through the prefix which is the node name or if each container must absolutely have a unique name).

*Downside*: The blocked `pull` problem still persists so some resources are going to be locked up for a while if this happens because we have no way to know if this is due to the issue or just a *very slow* `pull`.

TODO items:
- [ ] Integration tests
- [ ] Take a look at the implications on strategies and constraints, and parallel create/remove/build

### DIFF
--- a/scheduler/node/node.go
+++ b/scheduler/node/node.go
@@ -3,12 +3,20 @@ package node
 import (
 	"errors"
 	"strings"
+	"sync"
 
 	"github.com/docker/swarm/cluster"
 )
 
+var (
+	// ErrNotEnoughResources is thrown when a Node does not have any CPU or Memory left
+	ErrNotEnoughResources = errors.New("No more resources available on this node")
+)
+
 // Node is an abstract type used by the scheduler.
 type Node struct {
+	sync.RWMutex
+
 	ID         string
 	IP         string
 	Addr       string
@@ -73,11 +81,40 @@ func (n *Node) AddContainer(container *cluster.Container) error {
 		memory := container.Config.Memory
 		cpus := container.Config.CpuShares
 		if n.TotalMemory-memory < 0 || n.TotalCpus-cpus < 0 {
-			return errors.New("not enough resources")
+			return ErrNotEnoughResources
 		}
 		n.UsedMemory = n.UsedMemory + memory
 		n.UsedCpus = n.UsedCpus + cpus
 	}
 	n.Containers = append(n.Containers, container)
+	return nil
+}
+
+// ReserveResource reserve some resources for a container
+func (n *Node) ReserveResource(config *cluster.ContainerConfig) error {
+	n.Lock()
+	defer n.Unlock()
+	if config != nil {
+		memory := config.Memory
+		cpus := config.CpuShares
+		if (n.TotalMemory-n.UsedMemory)-memory < 0 || (n.TotalCpus-n.UsedCpus)-cpus < 0 {
+			return ErrNotEnoughResources
+		}
+		n.UsedMemory = n.UsedMemory + memory
+		n.UsedCpus = n.UsedCpus + cpus
+	}
+	return nil
+}
+
+// ReleaseResource releases resource on the node
+func (n *Node) ReleaseResource(config *cluster.ContainerConfig) error {
+	n.Lock()
+	defer n.Unlock()
+	if config != nil {
+		memory := config.Memory
+		cpus := config.CpuShares
+		n.UsedMemory = n.UsedMemory - memory
+		n.UsedCpus = n.UsedCpus - cpus
+	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


abronan:
CI failing on all the `constraints`/`affinity` tests, kind of makes sense.. This solves the simple parallel `create` issue on resource constraints but impacts the rest. Should look into it..

abronan:
Failure is obvious, I give a list of `Nodes` emptied from their containers and images thus it fails on constraint and affinities. Works fine otherwise. Will fix the issue with `listNodes` to statically account for used resources while including the actual state of existing Containers and Images.

scheduler/node/node.go
```diff
@@ -81,3 +89,32 @@ func (n *Node) AddContainer(container *cluster.Container) error {
...
+}
+
+// ReleaseResource releases resource on the node
+func (n *Node) ReleaseResource(config *cluster.ContainerConfig) error {
+	n.Lock()
+	defer n.Unlock()
+	if config != nil {
+		memory := config.Memory
+		cpus := config.CpuShares
+		n.UsedMemory = n.UsedMemory - memory
```
tnachen: I think you should check that this is >= 0, otherwise something really wrong happened

scheduler/node/node.go
```diff
@@ -81,3 +89,32 @@ func (n *Node) AddContainer(container *cluster.Container) error {
...
 }
+
+// ReserveResource reserve some resources for a container
+func (n *Node) ReserveResource(config *cluster.ContainerConfig) error {
+	n.Lock()
+	defer n.Unlock()
+	if config != nil {
+		memory := config.Memory
+		cpus := config.CpuShares
+		if (n.TotalMemory-n.UsedMemory)-memory < 0 || (n.TotalCpus-n.UsedCpus)-cpus < 0 {
```
tnachen: I'm wondering if you are going to run into precision problems here, as the check and the addition lower down there might not end up with the same answer for cpus.

scheduler/node/node.go
abronan: Agreed :+1:

scheduler/node/node.go
abronan: Cpu and Memory are plain integers (we don't define them as floating-point numbers nor convert them to floating-point notation), so this is ok.

abronan:
Closing this one in favor of #1261, will maybe revisit for 0.6 but until then the solution in #1261 plays nicer with scheduler plugins while this one could've ended with some side effects.

Issue #930, opened 2015-06-10
Leader Election with configurable TTL


aluzzardi:

feiyang21687:
@aluzzardi are you still working on it?

abronan:
Hi @feiyang21687, it's done on the `libkv` side for at least `etcd` and `zookeeper`.

The only bits left are the `Consul` TTL scheme (with session delete **behavior**) in `libkv` and finally add the flag here to configure the `TTL` accordingly. I guess that this issue should be tracked on both sides now that we have a separate repository to handle the distributed locking logic.

abronan:
Closed by #1228

PR #1228, opened 2015-09-22
Add the support for specifying Leader Election TTL

### LINKED ISSUE #1227: Leader failover too slow, taking 40 seconds
Hi there,
   I've started two swarm managers with advertise set to their own hostnames. With consul, I could implement the leader failover successfully, but found the whole process always took 40 seconds before the new leader information was written to the KV store in consul. I guess this is becuase the default TTL for consul session is 15 seconds. Consul takes a conservative method to multiple it with a factor of two plus 10 seconds as the check interval. 
   Do you guys know how to descrease the failover interval? Sounds like no existing config parameters help. Thanks.

abronan:
Fixes #1227 #930 


abronan:
@vieux hmm still the same failure on `images`:

```
not ok 53 docker images
# (in test file api/images.bats, line 44)
#   `[ "${#lines[@]}" -ge 4 ]' failed
```

abronan:
@jimmyxian Sounds like there is a race with the engine `connect/disconnect` on the `etcd` integration tests, at a first glance it seems that previous tests are not cleaning their entry properly with the `TTL` expiration and we don't get the event as we are supposed.

Might be worth adding some more checks to make sure the store is empty before trying this test? WDYT?

abronan:
@vieux @chanwit @jimmyxian Ready for review ;)

jimmyxian:
@abronan I will check it. :)
LGTM

chanwit:
Awesome @abronan!!

LGTM

vieux:
LGTM

PR #1229, opened 2015-09-23
fix test images

vieux:
busybox is only one layer now, since https://github.com/docker-library/official-images/commit/ab7d119a7728e5ddf845eed35ec6b7981293e3b5


vieux:
ping @abronan

abronan:
LGTM :v:

jimmyxian:
The test pass... 
LGTM  :)

PR #1234, opened 2015-09-24
Update developer instructions to use docker to build the binary

dnephin:
This change splits the `Dockerfile` into a `Dockerfile.build` that is used to build or develop (so no operations need to be done on the host, and a runtime `Dockerfile` for running the binary.  The runtime `Dockerfile` is pretty similar to https://github.com/docker/swarm-library-image/blob/master/Dockerfile but without the certs (so I guess it's really only for development).

I realize this is a rather intrusive change, feedback appreciated. 

I tried to use the `Makefile` conventions from the docker engine repo.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,12 @@
-FROM golang:1.5
+FROM    debian:jessie
 
-COPY . /go/src/github.com/docker/swarm
-WORKDIR /go/src/github.com/docker/swarm
+RUN     apt-get update && \
+        apt-get install -y --no-install-recommends ca-certificates
 
-ENV GOPATH /go/src/github.com/docker/swarm/Godeps/_workspace:$GOPATH
-RUN CGO_ENABLED=0 go install -v -a -tags netgo -installsuffix netgo -ldflags "-w -X github.com/docker/swarm/version.GITCOMMIT `git rev-parse --short HEAD`"
+COPY    dist/bin/swarm /swarm
+ENV     SWARM_HOST :2375
+EXPOSE  2375
+VOLUME  /.swarm
 
-ENV SWARM_HOST :2375
-EXPOSE 2375
-
-VOLUME $HOME/.swarm
-
-ENTRYPOINT ["swarm"]
+ENTRYPOINT ["/swarm"]
 CMD ["--help"]
--- a/README.md
+++ b/README.md
@@ -32,9 +32,20 @@ documentation on [docs.docker.com](http://docs.docker.com/swarm/).
 Developers should always download and install from source rather than
 using the Docker image.
 
-### Prerequisites
+### Developing with Docker
 
-1. Beginning with Swarm 0.4 golang 1.4.x or later is required for building Swarm. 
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -34,39 +34,16 @@ using the Docker image.
...
-You can also do this manually using the following commands:
-
 ```bash
-$ mkdir -p $GOPATH/src/github.com/docker/
-$ cd $GOPATH/src/github.com/docker/
-$ git clone https://github.com/docker/swarm
-$ cd swarm
-$ $GOPATH/bin/godep go install .
+$ git clone git@github.com:docker/swarm && cd swarm
+$ make binary
```
vieux: I'm on mac & I don't have docker, I don't think it should be a requirement.

dnephin: That seems really strange to me since this is one of the primary use cases for Docker (building things independent from the host), and this is a docker project!

That said, it's not so much a requirement just the "suggested path for getting started" to make it easy for contributors.

vieux: Installing boot2docker/machine is WAY more complicated than doing a go get

vieux:
If we decide to merge this, we should also update https://github.com/docker/swarm-library-image

dnephin:
Cool, I was thinking I would need to make a change there as well, I can prepare a PR for it

Dockerfile
```diff
@@ -1,15 +1,9 @@
-FROM golang:1.5
+FROM    scratch
 
-COPY . /go/src/github.com/docker/swarm
-WORKDIR /go/src/github.com/docker/swarm
+COPY    dist/bin/swarm /swarm
```
vieux: I don't think it's going to work with TLS, the certs will be missing.

dnephin: The old Dockerfile had the same issue, right? 

I'm not sure what to do with this Dockerfile. We could leave it as an example that lets developers run it locally (but document that it doesn't have certs).

vieux: I just saw your comment, sorry, I would still add TLS even for development.

vieux: no the old dockerfile is from golang:1.5 so it has certs

vieux:
ping @abronan @aluzzardi

Dockerfile
dnephin: Ah right, ok

dnephin:
Alright, I changed the `Dockerfile` base to be `debian:jessie` and installed the certs, and fixed the README to include instructions for both

abronan:
thanks @dnephin 

SGTM, having alternative ways to develop and build swarm using docker was needed imho. As long as it's complementary to the default way to install and develop with swarm, it's all good for everyone.

dnephin:
cool, anything else I should change here?

dnephin:
Should I start working on a PR for https://github.com/docker/swarm-library-image to use this new setup?  Do you think this will be merged?

ojongerius:
FWIW; I'd love to see this merged. It this blocked on @dnephin creating a PR for swarm-library-image?

PR #1236, opened 2015-09-24
Add integration test for leader election in the event of a backend store failure

abronan:

### DIFF
--- a/test/integration/replication.bats
+++ b/test/integration/replication.bats
@@ -2,8 +2,8 @@
 
 load helpers
 
-# Address on which the store will listen (random port between 8000 and 9000).
-STORE_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 8000 ))
+# Address on which the store will listen
+STORE_HOST=127.0.0.1:8500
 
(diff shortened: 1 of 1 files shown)


jimmyxian:
LGTM

test/integration/replication.bats
```diff
@@ -2,8 +2,8 @@
 
 load helpers
 
-# Address on which the store will listen (random port between 8000 and 9000).
-STORE_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 8000 ))
+# Address on which the store will listen
+STORE_HOST=127.0.0.1:8500
```
chanwit: Should we make the port of the store host configurable? (But maybe later is OK).

abronan: I was thinking about abstracting the test for the 3 supported backends because it's just Consul for now so I'll make sure I add this enhancement later on :+1: 

I fixed the port on this one because I shut down the store and bring it back, the swarm Manager expects the same endpoint to be reachable.

chanwit:
LGTM

Issue #1237, opened 2015-09-24
/info api returns with unicode back spaces


u2mejc:
When calling the api request for /info, unicode \b chars in ascii are sent.

Background, built today:
```shell
$ docker version
Client:
 Version:      1.9.0-dev
 API version:  1.21
 Go version:   go1.4.2
 Git commit:   743dc1c
 Built:        Mon Sep 21 13:20:09 UTC 2015
 OS/Arch:      linux/amd64

Server:
 Version:      swarm/0.5.0-dev
 API version:  1.16
 Go version:   go1.5.1
 Git commit:   HEAD
 Built:        
 OS/Arch:      linux/amd64

$ docker info
Containers: 38
Images: 4
Role: primary
Strategy: spread
Filters: affinity, health, constraint, port, dependency
Nodes: 1
 ip-192-168-100-206: 192.168.100.206:2375
  └ Containers: 38
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 2.051 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.13.0-62-generic, operatingsystem=Ubuntu 14.04.2 LTS, storagedriver=aufs
CPUs: 1
Total Memory: 2.051 GiB
Name: go-devil-flathead
```

When I use curl, I'm getting a malformed string back: (including unicode backspaces "\u0008")
```json
$ curl -v localhost:2375/v1.20/info
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 2375 (£0)
> GET /v1.20/info HTTP/1.1
> User-Agent: curl/7.35.0
> Host: localhost:2375
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Type: application/json
< Date: Thu, 24 Sep 2015 21:26:32 GMT
< Content-Length: 945
< 
[...]

Issue #1214, opened 2015-09-15
/info JSON output of docker swarm is not machine friendly


gauravphoenix:
Here is sample output given by "/info" api

```
{
    "ID": "",
    "Containers": 83,
    "Driver": "",
    "DriverStatus": [
        [
            "\bRole",
            "primary"
        ],
        [
            "\bStrategy",
            "spread"
        ],
        [
            "\bFilters",
            "affinity, health, constraint, port, dependency"
        ],
        [
            "\bNodes",
            "2"
        ],
        [
            "dockerhost2",
            "10.240.148.187:2376"
        ],
        [
            " └ Containers",
            "51"
        ],
        [
            " └ Reserved CPUs",
            "0 / 2"
        ],
        [
            " └ Reserved Memory",
            "0 B / 1.851 GiB"
        ],
        [
            " └ Labels",
            "executiondriver=native-0.2, kernelversion=3.16.0-4-amd64, operatingsystem=Debian GNU/Linux 8 (jessie), storagedriver=aufs"
        ],
        [
            "dockerhost3",
            "10.240.15.206:2376"
        ],
        [
            " └ Containers",
            "32"
        ],
        [
            " └ Reserved CPUs",
            "0 / 2"
        ],
        [
            " └ Reserved Memory",
            "0 B / 1.851 GiB"
        ],
        [
            " └ Labels",
            "com.example.environment=production, executiondriver=native-0.2, kernelversion=3.16.0-4-amd64, operatingsystem=Debian GNU/Linux 8 (jessie), storagedriver=aufs"
        ]
    ],
    "ExecutionDriver": "",
    "Images": 7,
    "KernelVersion": "",
    "OperatingSystem": "",
    "NCPU": 4,
    "MemTotal": 3975392870,
    "Name": "0ac63565be62",
    "Labels": null,
    "Debug": false,
    "NFd": 0,
    "NGoroutines": 0,
    "SystemTime": "2015-09-15T19:21:59.258278674Z",
    "NEventsListener": 0,
    "InitPath": "",
    "InitSha1": "",
    "IndexServerAddress": "",
    "MemoryLimit": true,
    "SwapLimit": true,
    "IPv4Forwarding": true,
    "BridgeNfIptables": true,
    "BridgeNfIp6tables": true,
    "DockerRootDir": "",
    "HttpProxy": "",
    "HttpsProxy": "",
    "NoProxy": ""
}
```

[...]

PR #1147, opened 2015-08-13
remove backspace characters in json

MHBauer:
I saw some escape chars coming back from /info api.

PR #1238, opened 2015-09-24
remove \b (\u0008) prefix from /info api calls

u2mejc:
This is related to issue https://github.com/docker/swarm/issues/1237

### DIFF
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -52,12 +52,12 @@ func (h *statusHandler) Status() [][]string {
 
 	if h.candidate != nil && !h.candidate.IsLeader() {
 		status = [][]string{
-			{"\bRole", "replica"},
-			{"\bPrimary", h.follower.Leader()},
+			{"Role", "replica"},
+			{"Primary", h.follower.Leader()},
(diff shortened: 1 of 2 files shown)


vieux:
duplicate of https://github.com/docker/swarm/pull/1147

u2mejc:
@vieux It seems like we're mixing data with presentation here?

This does appear to breaking the ability to programmatically parse the output (as noted in #1214). It seems like this should be modified in the cli where output presentation is formatted, would you be able point me in the right direction, I'll be happy to take a crack at fixing it.

vieux:
@u2mejc I agree it's not good as it is right now, changing the docker cli would be better so we could remove the \b in swarm.

nishanttotla:
Closing this PR since it needs more changes. Follow the discussion on #1625 instead.

vieux:
This is fixed in https://github.com/docker/swarm/pull/1721 :tada:

u2mejc:
@nishanttotla and @vieux, thank you! I'm looking forward to testing master as time allows!

PR #1240, opened 2015-09-25
Fix zookeeper regression and CI issues on Node removal

abronan:
Fixes a regression on discovery `List` and `Watch` and fix CI issue on node removal test for zookeeper.

ping @vieux

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -47,7 +47,7 @@
 		},
 		{
 			"ImportPath": "github.com/docker/libkv",
-			"Rev": "9a4eca472f7e780dfa13c37e63f406dbf8a628b9"
+			"Rev": "1d2361d74ecd9aa75b17992cf5c63d2c5ab717ab"
 		},
 		{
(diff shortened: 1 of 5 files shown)


chanwit:
LGTM

abronan:
After some more testing this is not a regression at least for the latest zookeeper version.

The problem only hits when using zookeeper `3.4.5` and earlier. This is a conservative measure as we are still using zookeeper `3.4.5` in the CI.

Now I think this is time to maintain our own zookeeper image with the latest version. This has been a pain point for me to narrow down the issue as I assumed that the zookeeper image was using the latest `3.4.6`... and the fix in itself is disgusting so either we merge this or we use the latest and leave this PR out, migrating to our own zookeeper image.

Now I'm not quite sure because the current zookeeper image is still widespread and a lot of users are going to hit that bug with `3.4.5` when assessing the new discovery in docker.

/cc @vieux @chanwit @jimmyxian

vieux:
LGTM

PR #1247, opened 2015-09-29
test against mesos 0.24.1

vieux:


abronan:
The last commit is :+1: but can you squash it? :smile:

abronan:
LGTM

PR #1248, opened 2015-09-30
Support filtering images by labels

### LINKED ISSUE #1244: /images/json ignores label filters
Looks like it only works with the `node` filter.  The `docker-compose` test suite needs this.

dnephin:
Fixes #1244

Blocked on https://github.com/samalba/dockerclient/pull/166, but I made the change locally for review/testing.

### DIFF
--- a/test/integration/api/images.bats
+++ b/test/integration/api/images.bats
@@ -43,3 +43,16 @@ function teardown() {
 	run docker_swarm images -a
 	[ "${#lines[@]}" -ge 3 ]
 }
+
+@test "docker images -f label" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	docker_swarm build -t image-with-labels $TESTDATA/imagelabel
+
+	run docker_swarm images \
+		--filter label=com.docker.swarm.test.integration.images=labeltest
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -eq 2 ]
+	[[ "${lines[1]}" == *"image-with-labels"* ]]
+}
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -104,7 +104,7 @@
 		},
 		{
 			"ImportPath": "github.com/samalba/dockerclient",
-			"Rev": "cfb489c624b635251a93e74e1e90eb0959c5367f"
+			"Rev": "73edd1c3a9d280bcec6a22a12fc90c1a46558b4e"
 		},
 		{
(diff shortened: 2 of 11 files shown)


vieux:
@dnephin thanks, I merged https://github.com/samalba/dockerclient/pull/166 can you rebase your PR. Thanks

aluzzardi:
Thanks @dnephin - mind adding an integration test? We usually do that for every API change since they are not directly tested by the UTs

dnephin:
Cool, I'll look at adding one

dnephin:
Rebased with dockerclient update and added an integration test

chanwit:
@dnephin it seems the build failed because `gofmt` validation. could you please re-format your codes again?

dnephin:
Oops, my git pre-commit hooks were not running with `gofmt -s` so I missed that. 

I also realized I messed up the rebase of updaing godeps somehow.

Both should be fixed now.

vieux:
Thanks @dnephin LGTM

vieux:
ping @abronan

abronan:
LGTM

test/integration/api/images.bats
```diff
@@ -43,3 +43,16 @@ function teardown() {
 	run docker_swarm images -a
 	[ "${#lines[@]}" -ge 3 ]
 }
+
+@test "docker images -f label" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	docker_swarm pull dnephin/swarm-integration-test-image-filter-labels:latest
```
aluzzardi: Can we get away with no pull? The problem is that it slows down the integration tests - even pulling an empty image is slow because of the back and forth with the registry, the extraction, and so on.

May I suggest to do a build instead and place the `Dockerfile` somewhere in `test/integration/testdata`, just like `test/integration/api/build.bats` is doing?

Thanks!

dnephin: Sounds good, I wasn't sure which way was better for the suite, or where to put the Dockerfile. I'll make this change now.

dnephin:
Changed the pull to a build in the integration test

aluzzardi:
Thanks @dnephin ! Sorry to be over pedantic, this looks really good I have a tiny last change: Can we do `FROM scratch` instead of `busybox` and get rid of the `CMD`? We are only interested in the label and a `FROM scratch` + `LABEL` will get us there.

dnephin:
Good call, I made that change

vieux:
ping @aluzzardi can you take another look ?

aluzzardi:
LGTM

PR #1250, opened 2015-09-30
Use smaller zk image for integration tests

dnephin:
This is an image I built for https://github.com/dnephin/compose-swarm-sandbox, it's basically the same `Dockerfile` except that it uses `alpine:edge` base, so it's half the size of the `jplock/zookeeper` image, so the tests should run a bit faster.

https://hub.docker.com/r/dnephin/docker-zookeeper/~/dockerfile/


vieux:
LGTM

abronan:
Nice @dnephin!

LGTM when CI good

chanwit:
This is awesome! Thank you @dnephin. It will save a lot of regression time.

PR #1251, opened 2015-09-30
Fix etcd lock TTL refresh value in seconds

abronan:
Latest changes in `libkv` to include the `LeaderTTL` flag here in swarm included a wrong calculation for the TTL lock refresh value in etcd. This would cause the refresh event to be triggered way more than expected for etcd.

This PR updates libkv to the latest including the change.


vieux:
LGTM

aluzzardi:
LGTM

PR #1252, opened 2015-10-02
fix exec -d

### LINKED ISSUE #1239: Swarm should not hijack Execs with Detach=true
The docker daemon doesn't hijack the request when Detach is true:
https://github.com/docker/docker/blob/634a848b8e3bdd8aed834559f3b2e0dfc7f5ae3a/api/client/exec.go#L52

Because swarm hijacks regardless, it may be contributing to https://github.com/samalba/dockerclient/pull/163 where the hijacked connection is reused and subsequent requests go directly to the node, bypassing swarm.

vieux:
fix #1239

ping @kgoudeaux

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -552,6 +552,14 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
 	c.eventsHandler.Wait(r.RemoteAddr, until)
 }
 
+// POST /exec/{execid:.*}/start
+func postExecStart(c *context, w http.ResponseWriter, r *http.Request) {
+	if r.Header.Get("Connection") == "" {
+		proxyContainer(c, w, r)
+	}
+	proxyHijack(c, w, r)
+}
+
 // POST /containers/{name:.*}/exec
 func postContainersExec(c *context, w http.ResponseWriter, r *http.Request) {
 	name := mux.Vars(r)["name"]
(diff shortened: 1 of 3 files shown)


test/integration/api/exec.bats
```diff
@@ -26,3 +26,22 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[ "$output" == "foobar" ]
 }
+
+@test "docker exec -d" {
+	start_docker_with_busybox 2
+	swarm_manage
+	docker_swarm create --name test_container busybox sleep 100
+
+	# if container is not running, exec will failed
```
MHBauer: I realize it's just copied from above, but `s/will failed/will fail/`

api/handlers.go
```diff
@@ -552,6 +552,14 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
 	c.eventsHandler.Wait(r.RemoteAddr, until)
 }
 
+// POST /exec/{execid:.*}/start
+func postExecStart(c *context, w http.ResponseWriter, r *http.Request) {
+	if r.Header.Get("Connection") == "" {
```
MHBauer: What does the connection header imply? Is this something that would be usable anywhere else?

vieux: it's currently the only way I found to check is it's hijacked by the cli or not. hijack uses "Connection: Upgrade" otherwise it's nothing.

test/integration/api/exec.bats
vieux: thanks

vieux:
ping @abronan @jimmyxian

jimmyxian:
LGTM

abronan:
LGTM

PR #1255, opened 2015-10-02
Support iterative integration testing in a container

dnephin:
When running integration tests using `test/integration/run.sh` I noticed there was a lot of startup cost. I had to wait for the `dind-master` image to pull and a few other things.

With this change I can start a shell in the container, and rerun integration tests. The startup cost is only a few seconds (it still runs a pull, but it should hit cached layers most of the time), and the tests run much faster.

I'm using:

```
INTEGRATION_CMD=bash STORAGE_DRIVER=aufs ./test/integration/run.sh
```

to get the shell, and from inside the container I can run:

```
./test_runner.sh <test name>
```

### DIFF
--- a/test/integration/test_runner.sh
+++ b/test/integration/test_runner.sh
@@ -18,8 +18,11 @@ TESTS=${@:-. compose discovery api mesos/api mesos/compose}
 export SWARM_BINARY=`mktemp`
 
 # Download docker-compose
-execute time curl -L --silent https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > /usr/local/bin/docker-compose
-execute chmod +x /usr/local/bin/docker-compose
+compose_path=/usr/local/bin/docker-compose
+if [ ! -e $compose_path ]; then
+	execute time curl -L --silent https://github.com/docker/compose/releases/download/${DOCKER_COMPOSE_VERSION}/docker-compose-`uname -s`-`uname -m` > $compose_path
+	execute chmod +x $compose_path
+fi
 
 # Build Swarm.
 execute time go build -o "$SWARM_BINARY" ../..
@@ -27,7 +30,8 @@ execute time go build -o "$SWARM_BINARY" ../..
 # Start the docker engine.
 execute docker daemon --log-level=panic \
 	--storage-driver="$STORAGE_DRIVER" &
-DOCKER_PID=$!
+
+trap "pkill -f docker" EXIT
 
 # Wait for it to become reachable.
 tries=10
--- a/test/integration/run.sh
+++ b/test/integration/run.sh
@@ -16,17 +16,18 @@ INTEGRATION_IMAGE=${INTEGRATION_IMAGE:-dockerswarm/swarm-test-env}
 # Make sure we upgrade the integration environment.
 docker pull $INTEGRATION_IMAGE
 
+# Command to run
+INTEGRATION_CMD=${INTEGRATION_CMD:-./test_runner.sh}
+
 # Start the integration tests in a Docker container.
-ID=$(docker run -d -t --privileged \
+BUILD_ID=${BUILD_ID:-swarm-integration-$USER}
+
+docker run -it --rm --privileged \
+	--name "$BUILD_ID" \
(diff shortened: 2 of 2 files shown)


vieux:
ping @aluzzardi

dnephin:
oops, looks like I used some spaces in the bash files, will fix

aluzzardi:
Wouldn't it make sense to just run `bats` if you don't want the startup costs of the container?

dnephin:
It's not really the startup cost of the containter that is the problem. Most of the cost is pulling the `dind-master` image (~70-90s)  inside the container (it's not cached after the container exits, so evern re-run of `test/integration/run.sh` has to pull it again).

If I were to run bats directly on the host, I'd have to ensure that my environment was setup in the way that `test_runner.sh` sets it up, right? 

So the advantage to this container+shell approach is that it lets me run inside a container (I don't have to worry about setting up my host, installing bats etc), but I still get almost the same runtime as I would from the host (only takes a couple seconds to hit the layer  cache).

test/integration/test_runner.sh
```diff
@@ -27,7 +30,8 @@ execute time go build -o "$SWARM_BINARY" ../..
 # Start the docker engine.
 execute docker --daemon --log-level=panic \
 	--storage-driver="$STORAGE_DRIVER" --exec-driver="$EXEC_DRIVER" &
-DOCKER_PID=$!
+
+trap "pkill -f docker" EXIT
```
dnephin: This `DOCKER_PID` was incorrect I think. It was getting a pid, but it wasn't the pid of the docker daemon. I guess it was the pid of eval ?

vieux: isn't this going to kill the docker on my host as well ?

dnephin: Only if you run it on the host. From the container it is fine. Does this ever get run on the host?

I've been trying to find another way to do this. I tried `$(cat /var/run/docker.pid)` but that isn't available immediately, so would require another retry block like below (which is maybe warranted in this case).

dnephin:
I'd like to use this setup for testing my other changes.  Could we continue this discussion?  

Do you understand my reasons for wanting to be able to run in a container, without having to wait on setup each run?

dnephin:
bump

abronan:
I would really like to see this as well, reason: My internet connection is unusually slow, and I have to pull those Images every time I have to run the tests. I don't mind the cost to pull Images usually but I can definitely see the value of this PR now :cry: 

So LGTM on this one.

(needs a rebase by the way, sorry @dnephin)

aluzzardi:
The reason I've been hesitating on this one is I believe we could achieve
similar results with a caching system.

We could cache images in a volume on the machine and then simply docker
import them when found - similar to what we've been doing with busybox.

This would of course be disabled on Jenkins.

Regarding this PR, I'm wondering whether:
1. Could you just start a swarm-integration-env container instead, and do
whatever you need to do inside?
2. Could you just set up bats locally so it works?
3. Would caching remove the need of doing this?

dnephin:
1. What is the swarm-integration-env container? Isn't it the image that's already being used from `/test/integration/Dockerfile` ? There are a bunch of steps done by `test_runner.sh` that need to be moved into the dockerfile, many of which won't work as part of build:
  * move the docker-compose download to the Dockerfile
  * move the `go build` to the Dockerfile
  * we can't move the `pull` or `import` to the Dockerfile, so no matter what we're left having to repeat a `import` at runtime.

2. I don't think installing bats on the host is a good option. I wan't to be able to use the same environment as the automated test runs, not waste time debugging unexpected failures. This is what Docker is all about.

3. Caching is another solution. I don't really understand why it's a better solution, and I don't think they are necessarily mutually exclusive. I prefer to have parameters for as many things as possible.  The change to check if `docker-compose` already exists is basically caching, so I have implemented one of the caching steps already.

I don't think you can reliably cache the `go build`, so I think you're forced to waste a few seconds there.
You can't cache the daemon startup, so you're forced to waste a few more seconds again.
A `docker import` from cache (and then create container) is still slower than skipping the import entirely, although probably pretty fast overall.


Is there something about this change that you think is not correct, or makes the build worse?  I'm happy to make changes to it.  I can cleanup the `pkill -f` by removing the `eval` I think.

nishanttotla:
Closing due to lack of activity. Please rebase and reopen if you wish to continue working on it.

PR #1258, opened 2015-10-02
Replace --leaderTTL flag by --replication-ttl

### LINKED ISSUE #1256: leaderTTL flag is weird
- All our other flags are `like-this` and not `likeTHIS`
- We use `leader` here but the other flag is `replication`

ping @abronan

abronan:
Fixes #1256 

ping @aluzzardi 


vieux:
LGTM

PR #1262, opened 2015-10-06
add 'docker network' support

vieux:
### List networks

This example assumes there are two nodes `node-0` and `node-1` in the cluster.

    $ docker networks ls
    NETWORK ID          NAME                   DRIVER
    3dd50db9706d        node-0/host            host
    09138343e80e        node-0/bridge          bridge
    8834dbd552e5        node-0/none            null
    45782acfe427        node-1/host            host
    8926accb25fd        node-1/bridge          bridge
    6382abccd23d        node-1/none            null

As you can see, each network name is prefixed by the node name.

## Create a network

By default, swarm is using the `overlay` network driver, a global
scope driver.

    $ docker network create swarm_network
    42131321acab3233ba342443Ba4312
    $ docker networks ls
    NETWORK ID          NAME                   DRIVER
    3dd50db9706d        node-0/host            host
    09138343e80e        node-0/bridge          bridge
    8834dbd552e5        node-0/none            null
    42131321acab        node-0/swarm_network   overlay
    45782acfe427        node-1/host            host
    8926accb25fd        node-1/bridge          bridge
    6382abccd23d        node-1/none            null
    42131321acab        node-1/swarm_network   overlay

As you can see here, the ID is the same on the two nodes, because it's the same
network.

If you want to want to create a local scope network (for example with the bridge
driver) you should use `<node>/<name>` otherwise your network will be created on a
random node.

    $ docker network create node-0/bridge2 -b bridge
    921817fefea521673217123abab223
    $ docker network create node-1/bridge2 -b bridge
    5262bbfe5616fef6627771289aacc2
    $ docker networks ls
    NETWORK ID          NAME                   DRIVER
    3dd50db9706d        node-0/host            host
    09138343e80e        node-0/bridge          bridge
    8834dbd552e5        node-0/none            null
    42131321acab        node-0/swarm_network   overlay
    921817fefea5        node-0/bridge2         brige
    45782acfe427        node-1/host            host
    8926accb25fd        node-1/bridge          bridge
    6382abccd23d        node-1/none            null
    42131321acab        node-1/swarm_network   overlay
    5262bbfe5616        node-1/bridge2         bridge

## Remove a network

To remove a network you can use its ID or its name.
If two different network have the same name, use may use `<node>/<name>`.

    $ docker network rm swarm_network
    42131321acab3233ba342443Ba4312
    $ docker network rm node-0/bridge2
    921817fefea521673217123abab223
    $ docker networks ls
    NETWORK ID          NAME                   DRIVER
    3dd50db9706d        node-0/host            host
    09138343e80e        node-0/bridge          bridge
    8834dbd552e5        node-0/none            null
    45782acfe427        node-1/host            host
    8926accb25fd        node-1/bridge          bridge
    6382abccd23d        node-1/none            null
    5262bbfe5616        node-1/bridge2         bridge

`swarm_network` was removed from every node, `bridge2` was removed only
from `node-0`.


vieux:
Here is the initial version @abronan @aluzzardi @dhiltgen @mavenugo @mrjana @aboch what do you think ?

vieux:
also, as soon as `docker network inspect` tells us if it's a `local` or `global` scope network, we can improve the UX (for example display `global` networks only once)

test/integration/store.bats
```diff
@@ -0,0 +1,47 @@
...
+load helpers
+
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+	stop_store
+}
+
+# Address on which the store will listen (random port between 8000 and 9000).
+STORE_HOST=127.0.0.1:$(( ( RANDOM % 1000 )  + 8000 ))
```
abronan: Note for later: not necessary to take a random port, it can be fixed to 8500 as we should test the scenario where the store is restarted to see if swarm behaves correctly in that case. (same as for Leader Election)

test/integration/store.bats
```diff
@@ -0,0 +1,47 @@
...
+
+	run docker_host -H ${HOSTS[1]} info
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"$DISCOVERY"* ]]
+
+	swarm_manage
+	run docker_swarm info
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *"Nodes: 2"* ]]
+}
```
abronan: Might be worth adding a skipped test here with a `TODO` item: `docker network restart store`.

There were known issues with the store in libnetwork when the store is restarted so just to make sure we don't forget it and we implement the test soon.

abronan:
LGTM, nothing to add, cool to have the support for `libnetwork` at last :+1: 

Will test it with global scoped network and external store but this is more dependent to `libnetwork` than swarm at this point so this shouldn't block this PR getting merged.

mavenugo:
@vieux woot !!!
just a question.  In a multi-host scenario, even though a network extends across multiple hosts and containers can belong to different hosts in the same network, docker engine's `docker network inspect` can display only those containers that are local to that engine. It clearly explains the engine behaviour. But, in a Swarm environment, `docker network inspect` on a given multi-host network should be able to consolidate all the containers across the hosts. That would be a killer feature to give a consolidated view of all the containers. Is it something that can be done ?

abronan:
@mavenugo We should be able to fetch the nodes on which there are containers attached to a given network and aggregate the information for `docker network inspect`. Can be done in a follow-up PR I guess.

vieux:
@mavenugo I believe so! I'm waiting for the next update in the engine and will give it a try

test/integration/store.bats
vieux: :+1:

chanwit:
@vieux Got 404 not found when doing ` network ls`. I am eagerly waiting for this PR :):):)

abronan:
@chanwit Hopefully this should be fixed in the new `libnetwork` vendoring if we get past this ;) there was a known issue with `docker network ls`.

abronan:
Merging, the next `libnetwork` vendoring will fix the tests with Mesos.

vieux:
\o/

chanwit:
Yeah :):):):)

PR #1263, opened 2015-10-07
engine: More robust refresh loop.

aluzzardi:
- Random heartbeat (between 30 and 60 seconds).
- Requires 3 failures before marking a node as dead.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"math/rand"
 	"net"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


cluster/engine.go
```diff
@@ -20,7 +21,9 @@ import (
 
 const (
 	// Force-refresh the state of the engine this often.
-	stateRefreshPeriod = 30 * time.Second
+	stateRefreshMinRange = 30
```
aluzzardi: Not sure about those values

vieux:
LGTM
ping @abronan

cluster/engine.go
abronan: We can tweak those values as needed. (And ideally be able to change those values when launching the Manager.)

abronan:
LGTM

PR #1264, opened 2015-10-07
Initialize the key/directory before watching on kv discovery

abronan:
Before that change, etcd and zookeeper would fail to instantiate
the discovery without the key being already there in the store or
created beforehand and implicitely by a 'swarm join'.

### DIFF
--- a/discovery/kv/kv.go
+++ b/discovery/kv/kv.go
@@ -112,6 +112,17 @@ func (s *Discovery) Watch(stopCh <-chan struct{}) (<-chan discovery.Entries, <-c
 		// Forever: Create a store watch, watch until we get an error and then try again.
 		// Will only stop if we receive a stopCh request.
 		for {
+			// Create the path to watch if it does not exist yet
+			exists, err := s.store.Exists(s.path)
+			if err != nil {
+				errCh <- err
+			}
+			if !exists {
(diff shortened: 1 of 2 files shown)


MHBauer:
Makes sense to me. I think you should be able to start join and manage in any order.

Unit test needs help with a mock of the new function.

abronan:
@MHBauer Thanks for reviewing, yes forgot about the Mock.. the usual :smile:

abronan:
I had to cheat with `testify` or it keeps the latest signature and throws a method mismatch error if done the right way...

On the other hand CI failures seems to be unrelated now.

vieux:
@abronan can you try with

```diff
diff discovery/kv/kv_test.go gofmt/discovery/kv/kv_test.go
--- /tmp/gofmt363354986	2015-10-08 18:07:21.910441239 -0700
+++ /tmp/gofmt472320449	2015-10-08 18:07:21.910441239 -0700
@@ -92,7 +92,7 @@
 	assert.EqualError(t, <-errCh, "test error")
 	// We have to drain the error channel otherwise Watch will get stuck.
 	go func() {
-		for _ = range errCh {
+		for range errCh {
 		}
 	}()

```

abronan:
Well, apparently that did the trick for `go fmt`.. :+1:

abronan:
PTAL, errors in Jenkins are unrelated.

vieux:
LGTM

PR #1270, opened 2015-10-08
Fix soft filter

jimmyxian:
This Case(2 nodes:node1 and node2):
```
[root@localhost ~]docker -H 127.0.0.1:2222 run -d -e contraint:storagedriver==~abc -e contraint:node==node1 busybox sleep 1000
a585b05dbc899fd724bbe7c5b2f35b9cf9ad85b02c1f8fd110bae4d005caee82
[root@localhost ~]# docker -H 127.0.0.1:2222 ps -a
a585b05dbc89        busybox:latest      "sleep 1000"        13 seconds ago                                              node2/berserk_banach
```
We can see that even use ` -e contraint:node==node1`,  the container was created on node2.

This PR fix this.
When we use soft Affinities/Constraints, we should not use `return`, should use `continue` so that meet the other Affinities/Constraints.

### DIFF
--- a/scheduler/filter/affinity.go
+++ b/scheduler/filter/affinity.go
@@ -64,7 +64,7 @@ func (f *AffinityFilter) Filter(config *cluster.ContainerConfig, nodes []*node.N
 		}
 		if len(candidates) == 0 {
 			if affinity.isSoft {
-				return nodes, nil
+				continue
 			}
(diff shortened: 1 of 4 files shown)


abronan:
LGTM (failures are unrelated), thanks @jimmyxian

aluzzardi:
Thanks !

LGTM

Issue #1232, opened 2015-09-24
Images on a single node ignore any strategies and filters.


lwhit3:
I have three different images (all websphere-liberty) with three different names: websphere-liberty, wlp1, wlp2. websphere-liberty is on both of my nodes in my two node swarm, wlp1 is on node1 and wlp2 is on node2. The last two are an assumption based on my issue. 

```
$ docker images
REPOSITORY          TAG                 IMAGE ID            CREATED     VIRTUAL SIZE
swarm               latest              207e8b983242        6 weeks ago      10.2 MB
websphere-liberty   latest              5126fa9711d2        7 days ago     635.1 MB
wlp1                latest              5126fa9711d2        7 days ago             635.1 MB
swarm               latest              207e8b983242        6 weeks ago        10.2 MB
wlp2                latest              5126fa9711d2        7 days ago             635.1 MB
websphere-liberty   latest              5126fa9711d2        7 days ago     635.1 MB
swarm               latest              207e8b983242        6 weeks ago        10.2 MB
```

I then run the images explicitly placing them on a node using the node name constraint.
WLP1 -> Node1 - works fine
WLP2 -> Node2 - works fine
WLP1 -> Node2 - Error response from daemon: unable to find a node that satisfies node==swarm-agent2
WLP2 -> Node1 - Error response from daemon: unable to find a node that satisfies node==swarm-agent1
websphere-liberty-> Node1 - works fine
websphere-liberty-> Node2 - works fine

```
CONTAINER ID        IMAGE           COMMAND       CREATED     STATUS    PORTS            NAMES
fb1911e51af6        websphere-liberty   "liberty-run /opt/ibm"   3 minutes ago  Up 3 minutes      9080/tcp, 9443/tcp   swarm-agent2/sad_heisenberg
a5069d00d01f        websphere-liberty   "liberty-run /opt/ibm"   4 minutes ago Up 3 minutes        9080/tcp, 9443/tcp   swarm-agent1/reverent_brattain
01f6396a4ad5        wlp1                "liberty-run /opt/ibm"   5 minutes ago    Up 5 minutes        9080/tcp, 9443/tcp   swarm-agent1/elegant_ritchie
5665b0df1113        wlp2                "liberty-run /opt/ibm"   5 minutes ago    Up 5 minutes        9080/tcp, 9443/tcp   swarm-agent2/elated_mccarthy
[...]

PR #1271, opened 2015-10-08
Do not try retry with soft-image-affinity when have node constraint

jimmyxian:
Fix: https://github.com/docker/swarm/issues/1232

Two nodes: node1 and node2, node 2 has `mycentos`
Before:
```
[root@localhost ~]# docker -H 127.0.0.1:2222 run -d -e constraint:node==node1 mycentos sleep 1000
Error response from daemon: unable to find a node that satisfies node==node1
```
After:
```
[root@localhost ~]# docker -H 127.0.0.1:2222 run -d -e constraint:node==node1 mycentos sleep 1000
Error response from daemon: Error: image library/mycentos:latest not found
```

### DIFF
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -158,3 +158,15 @@ func (c *ContainerConfig) AddAffinity(affinity string) error {
 	c.Labels[SwarmLabelNamespace+".affinities"] = string(labels)
 	return nil
 }
+
+// HaveNodeConstraint in config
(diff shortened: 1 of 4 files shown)


vieux:
ping @aluzzardi @abronan

abronan:
design LGTM, the error was really confusing before...

vieux:
LGTM

PR #1277, opened 2015-10-10
Engine should refresh the container on container rename event

mijia:


jimmyxian:
@mijia Thanks for contributing.
Can you try to add some tests for this? :)

abronan:
This might be tricky to write a test for this one. We can figure this out in another PR.

LGTM

ping @aluzzardi

aluzzardi:
LGTM

PR #1283, opened 2015-10-12
docs additions

### LINKED ISSUE #804: Getting started documentation is confusing
Hi, I am trying to follow this guide (http://docs.docker.com/swarm/) to setup swarm with two nodes. As I am relatively new to Docker and Swarm, I wrote down everything I tried here. Might be helpful to beginners like me and the documentation maintainer to see where people got stuck at. 

Here is what I have done. 

1.created 3 new Ubuntu instances (14.04 LTS) from AWS, install docker by following the instructions here (https://docs.docker.com/installation/ubuntulinux/)
Note, I added docker into the ubuntu usergroup as recommended for convenience.

Lets call them instance A, B and C. A meant to be used as the manager, B and C would be nodes.

2.On all 3 instances , do `docker pull swarm` (dont know if thats needed, maybe its only needed for the 'manager' node). 

3.On instance A, do `docker run --rm swarm create`. Succeeded.

4.On instance B, do `docker -H tcp://0.0.0.0:2375 -d`. This resulted in error: 

````
INFO[0000] +job serveapi(tcp://0.0.0.0:2375)
INFO[0000] Listening for HTTP on tcp (0.0.0.0:2375)
INFO[0000] /!\ DON'T BIND ON ANY IP ADDRESS WITHOUT setting -tlsverify IF YOU DON'T KNOW WHAT YOU'RE DOING /!\
FATA[0000] Shutting down daemon due to errors: pid file found, ensure docker is not running or delete /var/run/docker.pid

````

This is caused by docker -d is already running. So I did a 'sudo service stop docker' then run this commend again `docker -H tcp://0.0.0.0:2375 -d` , now I got this error:

````
INFO[0000] +job serveapi(tcp://0.0.0.0:2375)
INFO[0000] Listening for HTTP on tcp (0.0.0.0:2375)
INFO[0000] /!\ DON'T BIND ON ANY IP ADDRESS WITHOUT setting -tlsverify IF YOU DON'T KNOW WHAT YOU'RE DOING /!\
FATA[0000] Shutting down daemon due to errors: open /var/run/docker.pid: permission denied
````

It seems to be because of permission of that /var/run folder, so added sudo to the run command now it succeeded and running in daemon mode.

5.I did a `ctrl+z` and `bg` to put that daemon into background.
 

6.On the same instance (B), do `docker run -d swarm join --addr=<node_ip:2375> token://<cluster_id>
`. Now I got this error:

````
FATA[0000] Post http:///var/run/docker.sock/v1.18/containers/create: dial unix /var/run/docker.sock: no such file or directory. Are you trying to connect to a TLS-enabled daemon without TLS?
````

After doing this command: `export DOCKER_HOST="tcp://0.0.0.0:2375"`. The swarm join command worked.  **Worth adding this to the document for beginners like me?**

7.On instance A, do `docker run -d -p 8888:2375 swarm manage token://a5e663675dd0093d4f443cf633970cda &`.  succeeded. 

Do a `docker ps` to verify:

````
CONTAINER ID        IMAGE               COMMAND                CREATED             STATUS              PORTS                    NAMES
12e4ca329b8c        swarm:latest        "/swarm manage token   7 minutes ago       Up 7 minutes        0.0.0.0:8888->2375/tcp   distracted_jang
````

8.Following the guide to do a `docker -H tcp://0.0.0.0:2375` got an error:

````
ubuntu@ip-xxxx:~$ docker -H tcp://0.0.0.0:2375 info
FATA[0000] Cannot connect to the Docker daemon. Is 'docker -d' running on this host?

````

I simply tried to change the port to 8888 (the port for the docker manager) and run again:

''''
ubuntu@ip-172-30-0-177:~$ docker -H tcp://0.0.0.0:8888 info
Containers: 0
Strategy: spread
Nodes: 0
''''

Kind of worked, but got 0 nodes. Shouldn't it be 1 (instance B)? So the question here would be  **is 8888 the right port?**

9.I read from other issues, this might be caused by the docker daemon not started on host mode. 
So , on instance A, I did 'sudo service docker stop' and 'docker -H tcp://0.0.0.0:2375 -d&'. 

Then I repeated step 7 command and step 8. Got the same result: 0 containers. 

**What did I do wrong?**

### PR DESCRIPTION
**Pull request by MHBauer**
- add the unix socket, the default of the docker client
 - partially resolves #804 

the docs didn't start up the daemon on the unix socket, which results in a confusing error like 
```
FATA[0000] Post http:///var/run/docker.sock/v1.18/containers/create: dial unix /var/run/docker.sock: no such file or directory. Are you trying to connect to a TLS-enabled daemon without TLS?
```

### DIFF
--- a/docs/install-manual.md
+++ b/docs/install-manual.md
@@ -53,9 +53,12 @@ Docker daemon, monitors it, and updates the discovery backend with the node's st
 
 This example uses the Docker Hub based `token` discovery service. Log into **each node** and do the following.
 
-1. Start the Docker daemon with the `-H` flag. This ensures that the Docker remote API on *Swarm Agents* is available over TCP for the *Swarm Manager*.
+1. Start the Docker daemon with the `-H` flag. This ensures that the
+Docker remote API on *Swarm Agents* is available over TCP for the
+*Swarm Manager*, as well as the standard unix socket which is
+available in default docker installs.
 
-		$ docker daemon -H tcp://0.0.0.0:2375
+		$ docker daemon -H tcp://0.0.0.0:2375 -H unix:///var/run/docker.sock
 
 	> **Note**: versions of docker prior to 1.8 used the `-d` flag instead of the `docker daemon` subcommand.
 


MHBauer:
@moxiegirl @aluzzardi make sure this makes sense.

abronan:
SGTM :+1:

aluzzardi:
Awesome. Thanks!

LGTM

/cc @moxiegirl

moxiegirl:
Nice catch @MHBauer LGTM

PR #1287, opened 2015-10-13
Fix filter images by repo

### LINKED ISSUE #1257: /images/json ignores repo name filter
Similar to #1244 but for repo name instead of labels

dnephin:
Fixes #1257

Support `/images/json`  filter by repo name/tag

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -28,7 +28,7 @@ const APIVERSION = "1.21"
 func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
 	info := dockerclient.Info{
 		Containers:        int64(len(c.cluster.Containers())),
-		Images:            int64(len(c.cluster.Images(false, nil))),
+		Images:            int64(len(c.cluster.Images().Filter(cluster.ImageFilterOptions{}))),
 		DriverStatus:      c.statusHandler.Status(),
 		NEventsListener:   int64(c.eventsHandler.Size()),
 		Debug:             c.debug,
@@ -78,7 +78,7 @@ func getImages(c *context, w http.ResponseWriter, r *http.Request) {
 
 	// Create a map of engine address to the list of images it holds.
 	engineImages := make(map[string][]*cluster.Image)
-	for _, image := range c.cluster.Images(true, nil) {
+	for _, image := range c.cluster.Images() {
 		engineImages[image.Engine.Addr] = append(engineImages[image.Engine.Addr], image)
 	}
 
@@ -114,8 +114,6 @@ func getImagesJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	all := boolValue(r, "all")
-
 	filters, err := dockerfilters.FromParam(r.Form.Get("filters"))
 	if err != nil {
 		httpError(w, err.Error(), http.StatusInternalServerError)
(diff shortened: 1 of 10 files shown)


test/integration/api/images.bats
```diff
@@ -56,3 +56,15 @@ function teardown() {
 	[ "${#lines[@]}" -eq 2 ]
 	[[ "${lines[1]}" == *"image-with-labels"* ]]
 }
+
+@test "docker images imagetag" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	docker_swarm build -t testimage:latest $TESTDATA/imagelabel
```
dnephin: I tried using `docker tag` here but for some reason I'm seeing (what looks like) responses from the engine where `RepoTags` actually has tags from different repos (ex: `busybox:latest`, `testimage:latest`).

I haven't been able to reproduce this behaviour outside of the swarm test suite, so I'm not sure how that is happening.

This seems to work fine though.

vieux:
@dnephin what about filtering at the API level so we won't have to touch the cluster interface ?

dnephin:
That's certainly possible. What's wrong with touching the cluster interface?

The API layer kind of feels like the wrong place for this. Generally an api layer should only be for input validation, output formatting, handling http headers, and that kind of thing. I believe engine makes this distinction and tries to move the other logic into some other package. Moving it to the API also means that `parseImageName()` needs to become publish and the api needs to know more about image names.

It would be nice to have all the filtering happen in one place. Right now we have two of the filters in engine, and one in api. I think if the filtering by node that is done in api was actually done in engine, it would be a lot less code. It would be as simple as checking if it's own name was in the list of filtered nodes.  If not return.  Aggregation still needs to happen outside of engine, but maybe that could be part of the clustering code instead of api.

If the issue is having to deal with optional params (and golang's lack of keyword arguments). I saw a nice pattern in https://github.com/fsouza/go-dockerclient/blob/master/image.go#L85 of using a struct for the arg. That way if new ones get added the number of lines of code to change is still pretty minimal.


What do you think?

vieux:
You're right, the API might not be the right place. 
When I look at the diff in `cluster/mesos/cluster.go` it feels wrong to put it in the interface.
I'd like to keep `Containers()` and `Images()` the same.

What about changing `Images(all bool, filters dockerfilters.Args) []*Image` to `Images() Images` introduce an Images type and put all filtering inside ?

dnephin:
Cool, that sounds good

vieux:
@dnephin  thanks.

dnephin:
Ok, I made that change to move filtering to a new `Images` type.

I haven't moved the node filtering yet, but I can look at doing that next.

dnephin:
I think this test failure is unrelated to my change. 

Let me know if this looks good.  I'd like to save the refactor of node filtering for another branch (since it's not touched by this change).

dnephin:
@vieux how does this look?

api/handlers.go
```diff
@@ -26,9 +26,10 @@ const APIVERSION = "1.16"
 
 // GET /info
 func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
+	images := c.cluster.Images().Filter(cluster.ImageFilterOptions{})
```
vieux: you do need to declare this variable ?

dnephin: It's not necessary, but otherwise the line goes way over 80 char.

vieux:
besides a small nit, LGTM thanks @dnephin

dnephin:
Ok, I addressed the nit

@abronan it would be good to get this into 1.0.0 rc2 as well, for compose compatibility

dnephin:
rebased on master

abronan:
LGTM provided that the CI is green :)

aluzzardi:
LGTM

abronan:
Hmm there are some unrelated failures on `save`/`import`/`export`.

@vieux @jimmyxian not sure but looking at the failures, this seems like this is related to #1271? At least the tests are outdated with the image placement I guess.

jimmyxian:
@abronan 
OK, I will check it..

vieux:
LGTM

PR #1300, opened 2015-10-15
small refactor retry with soft-image-affinity

vieux:


vieux:
ping @jimmyxian I think it's cleaner with way but I'm wondering if we should store the soft affinity.

@abronan what do you think ?

chanwit:
LGTM

jimmyxian:
@vieux We can't do this. :)
Same as https://github.com/docker/swarm/pull/1099

vieux:
right :)

PR #1303, opened 2015-10-15
Integration: Fix timing issue on zookeeper failure test

abronan:
/cc @vieux 


vieux:
LGTM

PR #1308, opened 2015-10-16
Add error messages when pulling/loading images

pdevine:
This change adds error strings when there is an error while pushing or
pulling and image.

Signed-off-by: Patrick Devine <patrick.devine@docker.com>

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -491,10 +491,12 @@ func postImagesCreate(c *context, w http.ResponseWriter, r *http.Request) {
 			image += ":" + tag
 		}
 
+		var errorMessage string
 		errorFound := false
 		callback := func(what, status string, err error) {
 			if err != nil {
 				errorFound = true
+				errorMessage = err.Error()
 				sendJSONMessage(wf, what, fmt.Sprintf("Pulling %s... : %s", image, err.Error()))
 				return
 			}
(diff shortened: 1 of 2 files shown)


vieux:
LGTM

vieux:
ping @jimmyxian @chanwit

chanwit:
LGTM

jimmyxian:
Nice :) 
Thanks @pdevine

jimmyxian:
@pdevine @vieux @chanwit 
But I think the error message here is not proper .
Eg. Different node have different error message.
WDYT?

dnephin:
I think it's necessary. The old error format was not compatible with the format returned by docker engine.  It happened to work with the docker cli, but other clients (`docker-compose`, `samalba/dockerclient`, etc) were incompatible with the swarm error format.

This change makes it more compatible with the engine error format.

Yam408:
H
Sent from Yahoo Mail on Android 
 
  On السبت, أكتوبر ١٧, ٢٠١٥ at ٣:٤٢ ص, Chanwit Kaewkasi<notifications@github.com> wrote:   
LGTM

pdevine:
@jimmyxian Here's a link to the Docker v1.20 API reference:  https://docs.docker.com/reference/api/docker_remote_api_v1.20/  If you look at the "POST /images/create" endpoint, it's expecting there to be an Error detail because the http response code is set to 200.  If you don't set it, you end up breaking a lot of docker clients, as @dnephin mentioned.

That said, you do make a good point that different nodes could have had different error messages, but right now it looks like the callback is popping at the first one which fails, in which case the error message would be correct.  A better error message might include the host name that was causing the error.

jimmyxian:
@dnephin @pdevine Thanks.
Got it. :+1: :)

PR #1309, opened 2015-10-17
small improvment on network UX

vieux:


vieux:
This fix the jenkins issue . I'd like to get it merged asap. ping @docker/swarm-maintainers

jimmyxian:
LGTM :+1:

chanwit:
LGTM

cluster/swarm/cluster.go
```diff
@@ -176,7 +176,9 @@ func (c *Cluster) RemoveContainer(container *cluster.Container, force, volumes b
 
 // RemoveNetwork removes a network from the cluster
 func (c *Cluster) RemoveNetwork(network *cluster.Network) error {
-	return network.Engine.RemoveNetwork(network)
+	err := network.Engine.RemoveNetwork(network)
+	c.refreshNetworks()
```
aluzzardi: Uhm, why are we refreshing the entire cluster?

aluzzardi: Also, don't we usually do that in the proxy code rather than here? (for instance with images)

aluzzardi: As in, `proxyContainerAndForceRefresh` for instance

vieux:
unfortunately the replication is managed by libnetwork doing a network rm on 1 node can potentially remove a network on other nodes in the cluster. 

With images doing an action on one node cannot change a completely different node. 

Since there is no event emited or anything I decided to do a full refresh. 

Feel free to open a pr if you have another idea

aluzzardi:
Is there a way to know the scope of a network ?

vieux:
@aluzzardi we might want to range over the engines, remove the network from memory and let the "random" refresh pick up the change. Wdyt?

vieux:
Since rc2 yes

PR #1313, opened 2015-10-19
Fix the problem for the orders of filters

denverdino:
The default order of the filters is listed in /scheduler/filter/filter.go

```
filters = []Filter{
		&AffinityFilter{},
		&HealthFilter{},
		&ConstraintFilter{},
		&PortFilter{},
		&DependencyFilter{},
	}
}
```

But the AffinityFilters and ConstraintFilter support the "soft" constrains, which may filter out the candidates before applying other filters.

E.g. two nodes in the cluster, the container has the soft affinity to node1, so the its result will be node1, but the node 1 is unhealthy. So after the HealthFilter, there will be no candidate any more.  Details could be found in the test case.

The proper behavior for filtering should be handle all "hard" constrains firstly. And then find the candidates with the "soft" constrains.

So, I adjust the filters as following

```
	filters = []Filter{
		&HealthFilter{},
		&PortFilter{},
		&DependencyFilter{},
		&AffinityFilter{},
		&ConstraintFilter{},
	}
```

Signed-off-by: Li Yi <denverdino@gmail.com>

### DIFF
--- a/scheduler/filter/filter.go
+++ b/scheduler/filter/filter.go
@@ -24,11 +24,11 @@ var (
 
 func init() {
 	filters = []Filter{
-		&AffinityFilter{},
 		&HealthFilter{},
-		&ConstraintFilter{},
 		&PortFilter{},
 		&DependencyFilter{},
+		&AffinityFilter{},
+		&ConstraintFilter{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


vieux:
ping @aluzzardi 

LGTM

aluzzardi:
LGTM

PR #1314, opened 2015-10-19
Fix nullptr panic in pending containers.

aluzzardi:
Fixes #1289

Signed-off-by: Andrea Luzzardi <aluzzardi@gmail.com>

PR #1317, opened 2015-10-21
Integration test for regression with parallel docker run when specifying a port

abronan:
Integration test for #1314

ping @chanwit @aluzzardi @vieux 

Tested before (failure with the nil pointer panic) and after the patch (test succeeded). This integration test should ensure that we don't see a regression in the future for this specific issue.

Not sure if the best way so let me know if I can improve it (`parallel` command is not available in the swarm test image). Also don't know where to put the test so it ended up in `port.bats` :smile: 

### DIFF
--- a/test/integration/api/port.bats
+++ b/test/integration/api/port.bats
@@ -20,3 +20,21 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ "${lines[*]}" == *"8000"* ]]
 }
+
+@test "docker port parallel" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	declare -a pids
+	for i in 1 2 3; do
+		# Use a non existing image to ensure that we pull at the same time
+		docker_swarm run -d -p 8888 alpine:edge sleep 2 &
+		pids[$i]=$!
+	done
+
+	# Wait for jobs in the background and check their exit status
+	for pid in "${pids[@]}"; do
+		wait $pid
+		[ "$?" -eq 0 ]
+	done
+}


abronan:
ouch `golint` :facepunch: (will fix)

chanwit:
LGTM

This is awesome @abronan. I think it's OK to merge right away.

test/integration/api/port.bats
```diff
@@ -20,3 +20,21 @@ function teardown() {
...
 }
+
+@test "docker port parallel" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	declare -a pids
+	for i in 1 2 3; do
+		# Use a non existing image to ensure that we pull at the same time
+		docker_swarm run -d -p 8888 alpine:edge sleep ::: 2 2 &
```
aluzzardi: the `:::` syntax is part of the `parallel` program and doesn't make sense here.

this is the output of that command:
`sleep: invalid time interval ‘:::’`

test/integration/api/port.bats
```diff
@@ -20,3 +20,21 @@ function teardown() {
...
+		pids[$i]=$!
```
aluzzardi: I'm not sure this will work alright.

Using `$!` after using a shell function yields unexpected behavior, I don't recall exactly. @vieux ?

aluzzardi:
You might want to use the `parallel` command instead. That would require updating the testing image `test/integration/Dockerfile`

test/integration/api/port.bats
abronan: huh right :+1:

abronan:
@aluzzardi not sure if really necessary to have `parallel` on this one, but your call :)

@vieux WDYT?

vieux:
LGTM

vieux:
Let's update to parallel in another PR

PR #1322, opened 2015-10-21
Update api version to match the version that swarm supports

dnephin:
Swarm reports that it uses api version 1.16, but this is a lie.  With the 1.0.0 release it supports api version 1.21.


vieux:
I think this change will prevent swarm to manage docker engines with API < 1.21

vieux:
not at all my bad
LGTM

PR #1327, opened 2015-10-22
return 200 OK in /images/create

### LINKED ISSUE #1326: /images/create returns 201 with swarm, but 200 with docker, no documentation
It seems that when you call /images/create you get a 201 not a 200, but there is no documentation (at least that I could find) to support that the endpoint was modified from the default docker api, which returns a 200.

vieux:
fix #1326


vieux:
ping @jimmyxian

chanwit:
LGTM

jimmyxian:
LGTM

Issue #1183, opened 2015-08-31
mesos: no resources available to schedule container


ahmetalpbalkan:
Hi, I have a Mesos cluster consisting of 2 agent nodes:

![image](https://cloud.githubusercontent.com/assets/159209/9588871/94873d2c-4fdf-11e5-94e2-075a0c16cc04.png)

![image](https://cloud.githubusercontent.com/assets/159209/9588873/971008c6-4fdf-11e5-8201-41ed2dec1852.png)

![image](https://cloud.githubusercontent.com/assets/159209/9588918/e42f2d4e-4fdf-11e5-9dad-165651e2c45a.png)

I started 2 containers yesterday like the following:

```
docker run -d -c 1 -m 100 busybox sleep 999
docker run -d -c 1 -m 100 busybox sleep 999
```

and I can see they are completed now:

![image](https://cloud.githubusercontent.com/assets/159209/9588886/ab4436b4-4fdf-11e5-8400-574744183ecf.png)

However when I try to schedule another container I'm getting this “no resources available” error:

```
➜  ~  docker run -d -c 1 -m 100 busybox sleep 999999
FATA[0005] Error response from daemon: no resources available to schedule container
```

I tried cleaning up some stopped containers but

    docker rm -f `docker ps -aq`

is taking forever (been a few minutes, not returned yet, `docker rm` is getting stuck). I'm setting `-c 1 -m 100` it seemed okay and the cluster has plenty of resources offered. Any ideas what's going wrong here?

tnachen:
Hi there, I think there might be some problems around resource accounting in the swarm mesos scheduler that we need to figure out. Another thing is that we're refactoring the scheduler so we don't use a full offer per task, but try to run multiple tasks on a offer when we can. So short answer is that once we have the refactoring merged hopefully this will be fixed too.

h0tbird:
+1

rgardler:
This makes Swarm+Mesos completely unusable. What is even worse is that SWARM is sucking up all the resources from other frameworks too. @tnachen can you point me at the refactoring work you mention, this is a blocker for me right now and I'd like to follow along and at least provide some testing, if not patches.

vieux:
@ahmetalpbalkan hey, sorry for the delay, we are aware of the issue. Can you try with https://github.com/docker/swarm/pull/1212 to see if it helps ? thanks.

abronan:
[...]

PR #1330, opened 2015-10-23
fix issue with timeouts in mesos

vieux:
Fix https://github.com/docker/swarm/issues/1183

This PR fixes a race, unfortunately I cannot reproduce it with our current CI (docker in docker means it's localhost, so super fast)

But I was able to confirm this fixes some issues, I believe it's going to fix https://github.com/docker/swarm/issues/1183


vieux:
ping @ahmetalpbalkan @jimenez @tnachen

jimenez:
LGTM for now, we should merge this to have a quick fix but start thinking on a more maintainable refactor

PR #1342, opened 2015-10-26
switch to go1.5.1

vieux:


MHBauer:
Same changes as made in #1188. Makes sense to me.

dongluochen:
LGTM.

Issue #1269, opened 2015-10-08
Send authConfig when engine perfoms pull during create container


ezrasilvera:
Is there any special reason why authConfig is passed during /image/create but is not passed during the image pull initiated by swarm engine during the /container/create (see bellow passing nil) 

		// Otherwise, try to pull the image...
		if err = e.Pull(config.Image, nil); err != nil {
			return nil, err
		}

vieux:
Because the docker CLI unfortunately doesn't send us  the config on `/container/create`

ezrasilvera:
I understand but if I use remote API and not the client I could pass X-Registry-Auth in the header and that could have work, right?   Can't we check if X-Registry-Auth exists and pass it if so ?

ezrasilvera:
Probably would need small changes in postContainersCreate to pull the auth as well ..

ezrasilvera:
Just as a test case - we are currently using this change (#1359)  in our local Swarm based cloud service and we already have a customer that is successfully using private registry by passing the X-Registry-Auth header

thaJeztah:
@vieux is this related to https://github.com/docker/swarm/pull/1384?

ezrasilvera:
@thaJeztah I believe that  #1384 has the same code changes as in #1350 (for the Swarm part) so probably #1350  and this issue can be  closed now that it was merged.

thaJeztah:
Thanks @ezrasilvera! I'll close this issue, and will leave a comment on #1350

Issue #1197, opened 2015-09-05
docker run fails when using private registry that needs authentication


kawasakitoshiya:
`docker run` fails when using private registry that needs authentication, while `docker pull` works.
```
$ docker run -d quay.io/XXX/YYY
Error response from daemon: Error: Status 403 trying to pull repository XXX/YYY: "{\"error\": \"Permission Denied\"}"
```

I'm using `swarm:0.4.0`

kcraft:
I am having this issue as well. If I login to each node of the swarm cluster individually, then I can run private images on those nodes. If I login to the "swarm", then it does not work. Example output:

```
$ eval $(docker-machine env swarm-node0)

$ docker login my-private-repo.com
Username: kcraft
Password: 
Email: 
WARNING: login credentials saved in /Users/kcraft/.docker/config.json
Login Succeeded

$ docker run --rm -ti my-private-repo.com/my-image
Unable to find image 'my-private-repo.com/my-image:latest' locally
latest: Pulling from my-image
...
Status: Downloaded newer image for my-private-repo.com/my-image:latest
...
```

I can do the same with the other nodes in the swarm. However, if I try to login to the "swarm", it does not work:

```
$ eval $(docker-machine env --swarm swarm-node0)

$ docker login my-private-repo.com
Username: kcraft
Password: 
Email: 
WARNING: login credentials saved in /Users/kcraft/.docker/config.json
Login Succeeded

$ docker run --rm -ti -e "constraint:node==swarm-node0" my-private-repo.com/my-image
Error response from daemon: Authentication is required.
```

Docker version 1.9.1, build a34a1d5

kawasakitoshiya:
thanks to @ezrasilvera, PR https://github.com/docker/swarm/pull/1350 fixed this :smile:

PR #1350, opened 2015-10-28
Pass the X-Registry-Auth info in createContainer for the internal pull

ezrasilvera:
Swarm performs internally a pull request if a createContainer failed. For that pull operation Swarm doesn't have the registry auth parameters so it can't support private registries.
While it's true that "docker run" doesn't pass such information it is still valuable capability as the user can either use the REST API directly and pass X-Registry-Auth during create-container or even use a config file with X-Registry-Auth info forcing "docker run" to pass this header
This change can be much simpler without changing any of the function signatures if the ContainerConfig in samalba.dockerclient would include also the registry auth info
This was briefly discussed in #1269 and I believe  that also #1197 

Signed-off-by: Ezra Silvera <ezra@il.ibm.com>

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -125,7 +125,7 @@
 		},
 		{
 			"ImportPath": "github.com/samalba/dockerclient",
-			"Rev": "9433689e5c6bec5ac2ffa8debf721e88abfb6285"
+			"Rev": "9445e25ed06943cf2828bce5a92391d768513d21"
 		},
 		{
 			"ImportPath": "github.com/samuel/go-zookeeper/zk",
--- a/Godeps/_workspace/src/github.com/samalba/dockerclient/dockerclient.go
+++ b/Godeps/_workspace/src/github.com/samalba/dockerclient/dockerclient.go
@@ -335,6 +335,29 @@ func (client *DockerClient) ExecResize(id string, width, height int) error {
 	return nil
 }
 
+func (client *DockerClient) AttachContainer(id string, options *AttachOptions) (io.ReadCloser, error) {
+	v := url.Values{}
+	if options != nil {
+		if options.Logs {
+			v.Set("logs", "1")
+		}
+		if options.Stream {
+			v.Set("stream", "1")
+		}
+		if options.Stdin {
+			v.Set("stdin", "1")
+		}
+		if options.Stdout {
+			v.Set("stdout", "1")
+		}
+		if options.Stderr {
+			v.Set("stderr", "1")
+		}
+	}
+	uri := fmt.Sprintf("/%s/containers/%s/attach?%s", APIVersion, id, v.Encode())
+	return client.doStreamRequest("POST", uri, nil, nil)
+}
+
 func (client *DockerClient) StartContainer(id string, config *HostConfig) error {
 	data, err := json.Marshal(config)
 	if err != nil {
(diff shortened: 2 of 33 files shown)


aluzzardi:
Hey @ezrasilvera - thanks for the contribution.

The slight problem is that it requires the client to do custom things.

We have something in mind: change the Docker API so that `create` can actually pull the image if not there.

This means:
- `docker run` against swarm will actually give you the pull progress. Right now it just hangs since swarm is doing the pull for you in the background.
- `docker run` against swarm will actually pass the authentication details. Private registries will work with that

Thoughts? /cc @docker/swarm-maintainers 

We may still merge this PR while we work to do the previous

ezrasilvera:
@aluzzardi thanks for the feedback. I fully agree that having the pull progress is an excellent idea. 
However I don't think the client need to do anything to support the change I proposed - if you just point the docker run (with the --config) to a config file with the X-Registry-Auth this will work out of the box (and is working already for us :-)  The "docker pull " is basically doing the same but pulling the information from the file generated by the login , so basically I don't see any additional requirement that we introduce to docker run . 
As this is a simple and quick solution I suggest we merge this while we wait for the more complete solution that you suggested. This will also allow more time to develop that other solution by giving the users a way to work with private registries (and this is mandatory for some of our customers)

aluzzardi:
@ezrasilvera Could you provide an example of how to use it?

I wasn't aware there was a way to inject a X-Registry-Auth in a run :)

ezrasilvera:
yep just put in your config.json something like:
{
  "HttpHeaders": {
    "X-Registry-Auth": <base64 string of your X-Registry-Auth info>
  }
}
and run will inject this header ...

jimmyxian:
@ezrasilvera 
`docker run` do not have `--config` option. :)

ezrasilvera:
@jimmyxian correct. My bad. Indeed for docker run we need to change the client to load the file. (I forgot I did it for our tests sorry for that ..   )
But again on the API (containers/create) there is no problem in doing that. So at least you can do something if this is a mandatory requirement for you (as it's the case for us)
I guess the main Q is  How simple is this fix (merge complexity + potential risk)   Vs., how soon are we going to have an alternative solution

ezrasilvera:
@jimmyxian Actually, you can use the un-modfified client using "docker create" and the "docker start" . create does load the config AND send the info on the X-Registry-Auth.

ezrasilvera:
@jimmyxian But you can still do "docker create" which does send the headers (and read the config) and then docker start, right ?    So one can use the unmodified client and be able to send the registry auth info...

vieux:
@ezrasilvera I don't think `docker create` sends the header.

ezrasilvera:
@vieux This is strange - As far as I can see (I just tested again just to be sure) both create and run do pass the headers from the config and I can also see the header in Swarm (the exact one I defined in my config).
I also looked in the docker code itself and it seems that the headers are sent for these commands ..
Which docker version are you using? This is supported in 1.8.1 and above.

ezrasilvera:
@aluzzardi @vieux I tested the code with the latest Docker 1.9.0 and latest Swarm (I rebased the PR) and everything is working. In Docker 1.9 (as in 1.8) all relevant commands support the --config parameter allowing you to pass the X-Registry-Auth.

garagatyi:
@aluzzardi Yes, this would be nice! 
It also solves problem when container should be created on one node only and it is waste of resources to pull it on all nodes.

thaJeztah:
@ezrasilvera looks like you pulled in some none-related commits into your PR, also it's missing a sign-off

thaJeztah:
@jimmyxian @vieux this PR can probably be closed, per https://github.com/docker/swarm/issues/1269#issuecomment-167292641 - and this should already be covered by https://github.com/docker/swarm/pull/1384 (didn't want to close this PR, because I'm not a maintainer here, but feel free to close it yourself, @ezrasilvera, if you think it should be closed)

PR #1358, opened 2015-10-30
Update for new nsqio image

kynrai:
nsqio have changed they way the package the binaries into docker. There is now one image with all the binaries in it and each one is called on run.
This should fix the docker compose as per http://nsq.io/deployment/docker.html


abronan:
Thanks @kynrai

LGTM

ping @vieux

vieux:
LGTM thanks

Issue #1328, opened 2015-10-22
Scheduler does not evict engine from scheduling decision before it is flagged as dead


dongluochen:
In Swarm 1.0.0 it may take 90 seconds or more to flag a node down. In a swarm set up with 2 nodes 'vm2' and 'vm3'. When docker daemon on vm2 was shutdown, 'docker run’ commands to swarm failed for around 90 seconds until swarm ‘flag engine as dead’. During that time requests to swarm got refused. In these 2 node setup it has about 50% chance to report the issue. 

```
dchen@vm3:~/go/src/github.com/docker/swarm$ docker -H 0.0.0.0:2375 info
Containers: 37
Images: 7
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 2
 vm2: 192.168.56.202:4444
  └ Containers: 18
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 4.053 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.19.0-25-generic, operatingsystem=Ubuntu 14.04.3 LTS, storagedriver=aufs
 vm3: 192.168.56.203:4444
  └ Containers: 19
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 4.053 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.19.0-25-generic, operatingsystem=Ubuntu 14.04.3 LTS, storagedriver=aufs
CPUs: 2
Total Memory: 8.106 GiB
Name: vm3
```

```
dchen@vm3:~/go/src/github.com/docker/swarm$ ./swarm manage 192.168.56.202:4444,192.168.56.203:4444
INFO[0000] Listening for HTTP                            addr=127.0.0.1:2375 proto=tcp
INFO[0000] Registered Engine vm3 at 192.168.56.203:4444
INFO[0000] Registered Engine vm2 at 192.168.56.202:4444
ERRO[3135] HTTP error: Post http://192.168.56.202:4444/v1.15/containers/create: dial tcp 192.168.56.202:4444: getsockopt: connection refused  status=500
ERRO[3146] HTTP error: Post http://192.168.56.202:4444/v1.15/containers/create: dial tcp 192.168.56.202:4444: getsockopt: connection refused  status=500
[...]

PR #1363, opened 2015-10-31
add engine options for refresh interval

dongluochen:
This is part of the solution to #1328. This change supports engine refresh with the following 3 options. They can be used to control how fast a node's state will change. 

   --engine-refresh-min-interval "30s"					set engine refresh minimum interval
   --engine-refresh-max-interval "60s"					set engine refresh maximum interval
   --engine-refresh-retry "3"						set engine refresh retry count on failure

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -61,6 +61,21 @@ var (
 		Value: "10s",
 		Usage: "timeout period",
 	}
+	flRefreshIntervalMin = cli.StringFlag{
+		Name:  "engine-refresh-min-interval",
+		Value: "30s",
+		Usage: "set engine refresh minimum interval",
+	}
+	flRefreshIntervalMax = cli.StringFlag{
+		Name:  "engine-refresh-max-interval",
+		Value: "60s",
+		Usage: "set engine refresh maximum interval",
+	}
+	flRefreshRetry = cli.StringFlag{
+		Name:  "engine-refresh-retry",
+		Value: "3",
+		Usage: "set engine refresh retry count on failure",
+	}
 	flEnableCors = cli.BoolFlag{
 		Name:  "api-enable-cors, cors",
 		Usage: "enable CORS headers in the remote API",
@@ -117,7 +132,6 @@ var (
 		Usage: "discovery options",
 		Value: &cli.StringSlice{},
 	}
-
 	flLeaderElection = cli.BoolFlag{
 		Name:  "replication",
 		Usage: "Enable Swarm manager replication",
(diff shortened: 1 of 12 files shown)


cli/flags.go
```diff
@@ -61,6 +61,21 @@ var (
...
+		Name:  "engine-refresh-min-interval",
+		Value: "30s",
+		Usage: "set engine refresh minimum interval",
+	}
+	flRefreshIntervalMax = cli.StringFlag{
+		Name:  "engine-refresh-max-interval",
+		Value: "60s",
+		Usage: "set engine refresh maximum interval",
+	}
+	flRefreshRetry = cli.StringFlag{
```
jimmyxian: Maybe can use `IntFlag` here

dongluochen: Thanks @jimmyxian. I use `StringFlag` to be consistent with other duration flags like flTTL.

vieux: @dongluochen in that case I think the default value should be `3s` instead of `3` ?

vieux: @dongluochen but this one isn't really a duration, it's a count.

dongluochen: @jimmyxian @vieux you are right. Retry count should be int. I'll update.

abronan:
Design LGTM

@dong Can you please squash the last two commits? Thanks!

cluster/engine.go
```diff
@@ -52,10 +47,21 @@ func (d *delayer) Wait() <-chan time.Time {
...
+	waitPeriod := int64(d.rangeMin)
+	if delta := int64(d.rangeMax) - int64(d.rangeMin); delta > 0 {
+		// Int63n panics if the parameter is 0
+		waitPeriod += d.r.Int63n(delta)
+	}
 	return time.After(time.Duration(waitPeriod))
 }
 
+// EngineOpts represents the options for an engine
+type EngineOpts struct {
```
MHBauer: If we're going to add this type, do we want to add the other arguments to `NewEngine` to this type? This is very specific to timing rather than the engine.

dongluochen: @MHBauer Yes that's the plan @aluzzardi and I talked about. We want to aggregate engine arguments and cluster arguments. I should have described it in the pull request.

aluzzardi:
LGTM

aluzzardi:
I'd love to merge this, but our CI seems to fail.

@abronan Any idea why? I just merged your PR that was meant to fix the CI.

abronan:
@aluzzardi looking at the logs `docker ps --before` seems to be a legitimate failure. `reschedule with image affinity` seems to be legitimate too: image `busyboxabcde` seems to be inaccessible.

abronan:
Tested locally (will open a separate issue for unrelated CI failures, tests needs to be updated supposedly because of changes in docker master).

LGTM

Issue #1199, opened 2015-09-06
Authorization (ACLs) support in Swarm as a first step towards multi tenancy.


doronp:
## Overview

Swarm currently doesn't support multiple tenants or general authorization roles. It may be crucial for some users to be able to manage their docker cluster  using Swarm while being able to isolate between the different tenants and users. This means that while several tenants are using the same docker engines and other resources to run their containers they will have isolation in terms of the cluster management. (The scope of this issue is just the management. Run-time isolation in the engine level is out of scope for this issue)
## Very simple example

Suppose we want to isolate the users of different tenants.

Let there be tenant A and tenant B defined on some identity service.
Tenant A has user Alice and user Bob.
Tenant B has user Mike.
Say Alice creates a container on the Swarm cluster successfully. Call it C.
Mike now submits list containers requests - He gets empty list back.
Bob now submits list containers requests - He gets C back.

As you can see I assumed a 'policy' in which users on the same tenant can see each others containers. This does not have to be this way, those kind of rules as other fine grained ACLs showed be definable via the authorization plugin.
## Details

As Docker engine is likely to introduce [authorization support via a generic authorization plugin](https://github.com/docker/docker/issues/13697)
and as there is reason in effort & code reuse such as [uniform docker-swarm API](https://github.com/icecrime/docker-api) it may be sensible to have a similar mechanism in Swarm as it is 'one big aggregated engine'.
As Swarm does not necessarily contacts the engines in its cluster for all requests it receives (e.g. list containers) and as some of the API may behave differently in Swarm it may be necessary to have a bit different logic. However - Lots of shared abstractions remains.

This issue may encapsulate several issue for this context which are required in order to achieve the described goal amongst:
- [Docker like](https://github.com/docker/docker/blob/master/docs/extend/plugins.md) generic authorization plugin mechanism to enables multiple back-ends to implement enforcement of the authorization defined API in Swarm.
- Adding support for tenancy related data and entities depending on the chosen approach but may be similar to [Kubernetes approach](https://github.com/kubernetes/kubernetes/blob/release-1.0/docs/admin/authentication.md)
- Having some clean way to perform authorization related operation using the generic plugin (Pre/Post operations or more powerful Swarm handler wrapper)

At this point I'd not like to present a full blown solution, but to get feedback and let the proposal evolve.

Signed-off-by: Doron Podoleanu doronp@il.ibm.com

abronan:
[...]

PR #1365, opened 2015-11-01
ACLs in Swarm see #1199

doronp:
Authorization (ACLs) support in Swarm as a first step towards multi tenancy. #1199 

This PR includes (organized by Commits for convenience)

1 - Validation API & Default implementation- Approve/Disapprove/condition of 'Who wants to to what'?
2 - Handler API & Default implementation - Use labels to enforce decision of 1
3 - Feature flag --multiTenant to get this feature going (regular Swarm otherwise) 
4 - Sign off :-)


chanwit:
Thank your @doronp for this PR. Could you please sign the commits as suggested by Gordon?

test/authZ/simpleTest.sh
```diff
@@ -0,0 +1,91 @@
+SwarmNode="http://127.0.0.1:2379"
```
chanwit: We do testing with `bats`. Could you please convert this test to `bats`?

cli/commands.go
```diff
@@ -28,7 +28,8 @@ var (
 				flTLS, flTLSCaCert, flTLSCert, flTLSKey, flTLSVerify,
 				flHeartBeat,
 				flEnableCors,
-				flCluster, flDiscoveryOpt, flClusterOpt},
+				flCluster, flDiscoveryOpt, flClusterOpt,
+				multiTenant},
```
chanwit: We have a convention to declare this with the `fl` prefix.

api/primary.go
```diff
@@ -6,6 +6,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/swarm/cluster"
+	"github.com/docker/swarm/pkg/authZ"
```
chanwit: It's better to `godep` this package rather than putting it directly under `pkg/authZ`.

abronan:
Thanks @doronp! This is huge :)

This will not make it for 1.0 but we'll take time to review and get this merged for 1.1 if we all good.

In the meantime, can you please sign-off all your commits so that it passes the DCO check? Sorry about that!

api/primary.go
abronan: I guess it's fine to have it under `pkg` as this one is created for the purpose of the PR. Having it under Godep would mean having it under a separate repo, but we want to be able to fix things asap without having to update dependencies with Godeps at first (can be considered later on but for now it's best to keep it under `pkg` for now imho).

doronp:
Thanks for your comments. I've Opened PR again (for internal PM reasons) see #1366 
I've started working on the comments I've got - Some of them already in #1366 
Thanks
Doron P

PR #1366, opened 2015-11-02
ACLs in Swarm see #1199 (ReOpen for #1365)

doronp:
ACLs in Swarm see #1199 #1365 

Authorization (ACLs) support in Swarm as a first step towards multi tenancy. #1199

This PR includes (organized by Commits for convenience)

1 - Validation API & Default implementation- Approve/Disapprove/condition of 'Who wants to to what'?
2 - Handler API & Default implementation - Use labels to enforce decision of 1
3 - Feature flag --multiTenant to get this feature going (regular Swarm otherwise) 

(Re open PR for internal PM reasons)

### DIFF
--- a/api/primary.go
+++ b/api/primary.go
@@ -7,6 +7,8 @@ import (
 	"net/http/pprof"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/docker/docker/pkg/authorization"
+
 	"github.com/docker/swarm/cluster"
 	"github.com/gorilla/mux"
 )
@@ -19,6 +21,7 @@ type context struct {
 	debug         bool
 	tlsConfig     *tls.Config
 	apiVersion    string
+	authZPlugins  []authorization.Plugin
 }
 
 type handler func(c *context, w http.ResponseWriter, r *http.Request)
@@ -115,7 +118,7 @@ func profilerSetup(mainRouter *mux.Router, path string) {
 }
 
 // NewPrimary creates a new API router.
-func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHandler, debug, enableCors bool) *mux.Router {
+func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHandler, debug, enableCors bool, authorizationPlugins []string) *mux.Router {
 	// Register the API events handler in the cluster.
 	eventsHandler := newEventsHandler()
 	cluster.RegisterEventHandler(eventsHandler)
@@ -127,6 +130,11 @@ func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHan
 		tlsConfig:     tlsConfig,
 	}
 
+	if len(authorizationPlugins) > 0 {
+		context.authZPlugins = authorization.NewPlugins(authorizationPlugins)
+
+	}
+
 	r := mux.NewRouter()
 	setupPrimaryRouter(r, context, enableCors)
 
@@ -135,6 +143,39 @@ func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHan
 	}
 
 	return r
+}
+
+func newAuthorizationMiddleware(plugins []authorization.Plugin, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// FIXME: fill when authN gets in
+		// User and UserAuthNMethod are taken from AuthN plugins
+		// Currently tracked in https://github.com/docker/docker/pull/13994
+		user := ""
+		userAuthNMethod := ""
+		authCtx := authorization.NewCtx(plugins, user, userAuthNMethod, r.Method, r.RequestURI)
+
+		if err := authCtx.AuthZRequest(w, r); err != nil {
+			log.Errorf("#1 AuthZRequest for %s %s returned error: %s", r.Method, r.RequestURI, err)
+			http.Error(w, err.Error(), http.StatusUnauthorized)
+			return
+		}
+
+		rw := authorization.NewResponseModifier(w)
+
+		next.ServeHTTP(rw, r)
+
+		//		if err := handler(rw, r); err != nil {
+
+		//		}
+
+		if err := authCtx.AuthZResponse(rw, r); err != nil {
+			log.Errorf("#2 AuthZResponse for %s %s returned error: %s", r.Method, r.RequestURI, err)
+			http.Error(w, err.Error(), http.StatusUnauthorized)
+			return
+		}
+
+		//TODO - which cases pass via here missing
+	})
 }
 
 func setupPrimaryRouter(r *mux.Router, context *context, enableCors bool) {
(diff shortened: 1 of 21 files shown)


thaJeztah:
@vieux @aluzzardi should this be looked at together with https://github.com/docker/docker/issues/13697 and https://github.com/docker/docker/pull/17412 (@NathanMcCauley is currently working with them on that proposal)

doronp:
@abronan @chanwit Done with Most of the comments (except Shell tests should be converted to BATS I think)
I agree that moving to godep should happen at a later phase.
Kindly put the tags from the old PR to this one and lets continue our work here.

abronan:
@doronp done! :) But it seems that some of your commits are still not signed though :( Also would be better if you can squash some commits (like all those `Fix for golint` mostly). Especially if you plan to add some more on top.

@thaJeztah Thanks this is helpful! We'll take a look at it as this seems closely related.

tiborvass:
FYI there's also docker/docker#17412 and docker/docker#17687 dealing with this.

doronp:
@chanwit : Added BATS tests, however since the tests are run without the --multiTenant flag they fail and hence the build fails. Can you put the Flag on for those tests?
P.S. The tag milestone 1.1.0 did not move from the closed PR (#1365) if it matters.
P.P.S When squashing I older commit inside (not mine) - Should I override that or is that fine?
@tiborvass: Tnx.

abronan:
@doronp Digging into this one. We are seeing the first bits being merged in docker with docker/docker#15365 (for `authZ` support with plugins). I wonder how we can think about an integration and use the support straight out of docker maybe? This would simplify the code tremendously on the Swarm side. WDYT?

doronp:
@abronan. 

Indeed it would be neat to harness the docker side authZ support.
I am not completely sure what you had in mind but I can assume at least two additional approaches:

1 - Swarm will pass the requests to the engines in its cluster and those will communicate with the AuthZ back-end via the plugin - Swarm will then aggregate and handle the responses.

I suspect there will be lots of inefficacy in this approach - Swarm will have to access the engines directly every time - it would not be able to use the internal cluster data as to complete operation as it does today*. This will probably have tremendous performance impact especially on bigger cluster (which is a reasonable assumption if someone uses ACL at all)
for example:
	 ps -a - list all container will access all over the cluster - then filter out to show only the users  containers etc.

*We can keep the operations such as ps in Swarm and only make it do AuthZ  via the engines but it is still inefficient. 
		

2 - Take the AuthZ(+AuthN) plugin from docker and make swarm do that job itself - Swarm is just like an engine in that case

I am going on digging the possible implementation options.

let me know what you think.

doronp:
@vieux @abronan @aluzzardi commit 497e328 is the plugin package import - It does not interfere with functionality and passes all tests. It will help a lot to review and merge it ASAP - better not mix it with the AuthZ plugin code (701d09e commit WIP)  Thanks.

abronan:

doronp:
@vieux @dongluochen When can we start working on this (I think start with commit 497e328) ? 
I'd not want to upset Gordon;-)

MHBauer:
Since what's added in the `refactor 402ec3ee` commit is moved in the later `add middleware f827edad` commit, I would rather it just went away.

Can't help with the godeps. This needs a rebase.

api/server.go
```diff
@@ -34,9 +35,26 @@ func (d *dispatcher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 // Server is a Docker API server.
 type Server struct {
-	hosts      []string
-	tlsConfig  *tls.Config
-	dispatcher *dispatcher
+	cfg          *Config
```
MHBauer: I see this is where the `Config` struct below on L#50 is used, but where is this pointer used?

api/server.go
```diff
@@ -34,9 +35,26 @@ func (d *dispatcher) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	hosts        []string
+	tlsConfig    *tls.Config
+	dispatcher   *dispatcher
+	authZPlugins []authorization.Plugin
+}
+
+// versionMatcher defines a variable matcher to be parsed by the router
+// when a request is about to be served.
+const versionMatcher = "/v{version:[0-9.]+}"
```
MHBauer: Good idea to make this a constant, but I don't see it used. Looks like it should be used in `api/primary.go`.

api/primary.go
```diff
@@ -115,18 +119,23 @@ func profilerSetup(mainRouter *mux.Router, path string) {
...
 
 // NewPrimary creates a new API router.
-func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHandler, debug, enableCors bool) *mux.Router {
+func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHandler, debug, enableCors bool, authorizationPlugins []string) *mux.Router {
 	// Register the API events handler in the cluster.
 	eventsHandler := newEventsHandler()
 	cluster.RegisterEventHandler(eventsHandler)
 
-	context := &context{
+	context := &routerContext{
```
MHBauer: Why was this renamed? Makes a lot of clutter in the diff.

doronp:
@MHBauer There was a huge rebase maily  #1880  
Had lots of the same changes which were pre requisites for the AuthZ plugin usage.. 

That made separating Godeps and AuthZ plugin actual  usage less relevant.

About the next step of actually importing and using AuthZ plugin from the Engine:
I did try to test the authZ usage itself and I get conflict version of the discovery version used in it and in swarm:

cannot save github.com/docker/docker/pkg/authorization at revision 6eb38359954269103f2a3094283c4cb0a1cb63f9: already have github.com/docker/docker/pkg/discovery at revision b65fd8e879545e8c9b859ea9b6b825ac50c79e46.

Does that mean that docker authZ and swarm should decide on one revision? Its a bit of a blocker...

About the failing tests - I am not sure this is related to my changes.

api/primary.go
```diff
@@ -7,6 +7,8 @@ import (
 	"net/http/pprof"
 
 	log "github.com/Sirupsen/logrus"
+	//	"github.com/docker/docker/pkg/authorization"
```
ezrasilvera: Why don't you remove it for now instead of commenting out ?

doronp:
@ezrasilvera Good catch. The comment out is due to Godeps indeed. Without that code there is no commit really, with it the error you get is misleading since I could not update the Godeps yet. I will try to determine if I can persuade one side to use the others version to resolve that.

doronp:
So I changed the revision in the Swarm side it compiles now - Fails are not related to the change AFAIK.

doronp:
@MHBauer Ping

MHBauer:
Be forewarned, docker/docker#21897 is about to move the auth stuff around.

duglin:
@vieux @abronan what do you think about the design of this?

MHBauer:
I think the point of https://github.com/docker/swarm/pull/1366#issuecomment-164195664 was if we could just init the Middleware directly without copying the init into swarm. `newAuthorizationMiddleware` is already out of date due to changes in docker-engine. It appears to be just a copy of the initial version. If that was not the intent of the comment, it is still what I myself think.

MHBauer:
We'd have to reach directly into `docker/docker/api/server/middleware` to do that though. (Un)Fortunately that is moved by docker/docker#21897  into a `pkg` so it will be easy to grab in the future. 

I can see that was considered, because it is commented out in `primary.go`. Some general cleanup to remove comments would be nice.

doronp:
@MHBauer The reason newAuthorizationMiddleware is copied is that it is a bit different, Swarm handlers call chain does not support returning error whereas the dockers does.

In order to avoid that (almost a) copy there needs to be done another change of behaviour in Swarm which is that the handler being used should be able to return error and be like the one in Docker.

When that is done the behaviour could be the same as in Dockers middle wear. And then we can avoid that private function there and use Dockers.

I was planning to do that on another PR since it is not a small change.

doronp:
@dongluochen - @abronan advised me you can help with this. What are the next steps on my side? 10x.

hsluoyz:
Hi, may I ask what status of the ACL integration for Swarm is right now?

I have developed a Docker auth plugin called [casbin-authz-plugin](https://github.com/casbin/casbin-authz-plugin), and I want it to work on Docker Swarm too. What is the suggested way to do this? Thanks!

dongluochen:
@hsluoyz Is your plugin for Docker Swarm-mode? You can find the difference of classic Swarm and Swarm-mode [here](https://github.com/docker/swarm#swarm-disambiguation). You may open issue at  https://github.com/moby/moby for Swarm mode authentication.

hsluoyz:
Hi @dongluochen ,

> Is your plugin for Docker Swarm-mode?

No, my plugin is still a local Docker engine plugin base on: https://github.com/docker/go-plugins-helpers/tree/master/authorization

> You can find the difference of classic Swarm and Swarm-mode here.

So I think I should develop the auth plugin based on the new Swarm-mode on the moby repo.

> You may open issue at https://github.com/moby/moby for Swarm mode authentication.

I didn't find any information about how to develop an authorization plugin for Docker Swarm mode, I will open an issue in moby to ask about it as you suggested, thanks!

PR #1368, opened 2015-11-02
Add configurable discovery path option

abronan:
This PR allows to configure the discovery path using the
`--discovery-opt` flag (with `kv.path=path/to/nodes`). We
can point to `docker/nodes` and use the docker discovery.

If docker instances are advertising to the cluster using
the `--cluster-advertise` flag, the `swarm join` command
becomes unnecessary.

### DIFF
--- a/discovery/kv/kv.go
+++ b/discovery/kv/kv.go
@@ -17,7 +17,7 @@ import (
 )
 
 const (
-	discoveryPath = "docker/swarm/nodes"
+	defaultDiscoveryPath = "docker/swarm/nodes"
 )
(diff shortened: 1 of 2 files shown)


abronan:
Should add a simple test maybe..

/cc @icecrime @vieux

dongluochen:
LGTM.

abronan:
There must be something wrong in the CI env, failures are not related though.

vieux:
LGTM

abronan:
This is :green_apple:

PR #1376, opened 2015-11-03
Fix first sentence in docs

bfirsh:
"create and access to a pool" didn't make any sense, so I rewrote it to reflect some of the similar stuff we have on the website.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -10,11 +10,11 @@ parent="smn_workw_swarm"
 
 # Docker Swarm
 
-Docker Swarm is native clustering for Docker. It allows you create and access to
-a pool of Docker hosts using the full suite of Docker tools. Because Docker
-Swarm serves the standard Docker API, any tool that already communicates with a
-Docker daemon can use Swarm to transparently scale to multiple hosts. Supported
-tools include, but are not limited to, the following:
+Docker Swarm is native clustering for Docker. It turns a pool of Docker hosts
+into a single, virtual Docker host. Because Docker Swarm serves the standard
+Docker API, any tool that already communicates with a Docker daemon can use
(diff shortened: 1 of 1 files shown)


vieux:
ping @moxiegirl

moxiegirl:
LGTM

MHBauer:
The whole paragraph reads kind of funky to me, but it looks better with the change.

docs/index.md
```diff
@@ -10,11 +10,11 @@ parent="smn_workw_swarm"
 
 # Docker Swarm
 
-Docker Swarm is native clustering for Docker. It allows you create and access to
-a pool of Docker hosts using the full suite of Docker tools. Because Docker
-Swarm serves the standard Docker API, any tool that already communicates with a
-Docker daemon can use Swarm to transparently scale to multiple hosts. Supported
-tools include, but are not limited to, the following:
+Docker Swarm is native clustering for Docker. It turns a pool of Docker hosts
```
abronan: I would maybe replace slightly with: `Docker Swarm is a Docker native clustering tool. [...]`\

or `Docker Swarm is a native clustering solution for Docker containers`

Reads funky to me but I'm not a native speaker so.. :smile:

MHBauer: I like the second better. I think we can still use the world `tool` instead of `solution` in that sentence.

I feel like we don't define `native`. I'm pretty sure it refers to the `Swarm serves the Docker API` part, but it is never explicitly said.

aluzzardi:
LGTM - @bfirsh @MHBauer are we good to merge this?

MHBauer:
Go for it. Don't let me hold anything up.

Issue #1319, opened 2015-10-21
Godep update cmd remove the test file


jimmyxian:
The latest `godep` automatic remove the `_test` file.

I found that use `godep save -t` can save _test file.(https://github.com/tools/godep/commit/87fb5cdc4cbd6f0150f2e58087e235e843065304)
But `godep update` do not have `-t` option to save `_test` file(https://github.com/tools/godep/issues/298).

Now, which way should we go? 
ping @docker/swarm-maintainers

vieux:
I think we should remove the tests

jimmyxian:
@vieux OK, got it!

PR #1384, opened 2015-11-06
Add token pass-thru for Authconfig

dhiltgen:
Add token pass-thru for Authconfig
    
This augments the CreateContainer call to detect the AuthConfig header
and use any supplied auth for pull operations.  This will allow pulling
of protected image on to specific node during the create operation.
    
CLI usage example using username/password:

    # Calculate the header
    REPO_USER=yourusername
    read -s PASSWORD
    HEADER=$(echo "{\"username\":\"${REPO_USER}\",\"password\":\"${PASSWORD}\"}"|base64 -w 0 )
    unset PASSWORD
    echo HEADER=$HEADER

    # Then add the following to your ~/.docker/config.json
    "HttpHeaders": {
        "X-Registry-Auth": "<HEADER string from above>"
    }
    
    # Now run a private image against swarm:
    docker run --rm -it yourprivateimage:latest
    
CLI usage example using registry tokens: (Required engine 1.10 with new auth token support)
    
    REPO=yourrepo/yourimage
    REPO_USER=yourusername
    read -s PASSWORD
    AUTH_URL=https://auth.docker.io/token
    TOKEN=$(curl -s -u "${REPO_USER}:${PASSWORD}" "${AUTH_URL}?scope=repository:${REPO}:pull&service=registry.docker.io" |
        jq -r ".token")
    HEADER=$(echo "{\"registrytoken\":\"${TOKEN}\"}"|base64 -w 0 )
    echo HEADER=$HEADER
    
    # Update the docker config as above, but the token will expire quickly...
    
Signed-off-by: Daniel Hiltgen <daniel.hiltgen@docker.com>

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -503,7 +503,7 @@ func (e *Engine) TotalCpus() int64 {
 }
 
 // Create a new container
-func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool) (*Container, error) {
+func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool, authConfig *dockerclient.AuthConfig) (*Container, error) {
 	var (
 		err    error
 		id     string
@@ -521,17 +521,17 @@ func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool) (*
 	dockerConfig.CpuShares = int64(math.Ceil(float64(config.CpuShares*1024) / float64(e.Cpus)))
 	dockerConfig.HostConfig.CpuShares = dockerConfig.CpuShares
 
-	if id, err = client.CreateContainer(&dockerConfig, name); err != nil {
+	if id, err = client.CreateContainer(&dockerConfig, name, nil); err != nil {
 		// If the error is other than not found, abort immediately.
 		if err != dockerclient.ErrImageNotFound || !pullImage {
 			return nil, err
 		}
 		// Otherwise, try to pull the image...
-		if err = e.Pull(config.Image, nil); err != nil {
+		if err = e.Pull(config.Image, authConfig); err != nil {
 			return nil, err
 		}
 		// ...And try again.
-		if id, err = client.CreateContainer(&dockerConfig, name); err != nil {
+		if id, err = client.CreateContainer(&dockerConfig, name, nil); err != nil {
 			return nil, err
 		}
 	}
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -125,7 +125,7 @@
 		},
 		{
 			"ImportPath": "github.com/samalba/dockerclient",
-			"Rev": "9445e25ed06943cf2828bce5a92391d768513d21"
+			"Rev": "4656b1bc6cbc06b75d65983475e4809cbd53ebb5"
 		},
 		{
 			"ImportPath": "github.com/samuel/go-zookeeper/zk",
--- a/Godeps/_workspace/src/github.com/samalba/dockerclient/auth.go
+++ b/Godeps/_workspace/src/github.com/samalba/dockerclient/auth.go
@@ -8,9 +8,10 @@ import (
 
 // AuthConfig hold parameters for authenticating with the docker registry
 type AuthConfig struct {
-	Username string `json:"username,omitempty"`
-	Password string `json:"password,omitempty"`
-	Email    string `json:"email,omitempty"`
+	Username      string `json:"username,omitempty"`
+	Password      string `json:"password,omitempty"`
+	Email         string `json:"email,omitempty"`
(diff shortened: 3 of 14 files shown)


dhiltgen:
Related to https://github.com/docker/docker/pull/17741 and https://github.com/samalba/dockerclient/pull/184

MHBauer:
Looks a lot like #1350, but with a godep update.

aluzzardi:
@ezrasilvera @dhiltgen There seems to be another implementation of this feature in #1350.

I agree with the feature per se - we need this.

Could you two figure out if both PRs address your use case and which one we should merge? I'm happy with either.

Thanks!

ezrasilvera:
@dhiltgen I think there are two separate set of changes here:
1) Change Swarm to handle auth header during create()
2) Change the dockerclient to load the auth file and push the credential during run 
This PR tries to cover both (1) and (2).  Your changes for (1) are similar/identical to  #1350 - so I suggest we first just merge #1350 to get the Swarm support. Note that it can already be fully utilized without any change to the dockerclient using the --config flag)
If needed we can then have a separate unrelated PR to update the GODEP when the dockerclient is updated. (BTW, I actually have additional idea on changing the dockerclient that will save a lot of code for this ..)
What do you think ?

dhiltgen:
Now that the engine PR has merged, and dockerclient has support for the new registrytoken field in AuthConfig, I've updated this PR to include a proper vendoring of dockerclient.

I've also updated the example to show usage from the CLI.

If the maintainers want to split into two PRs that's OK with me.

cluster/engine.go
```diff
@@ -521,17 +521,17 @@ func (e *Engine) Create(config *ContainerConfig, name string, pullImage bool) (*
...
 		}
 		// Otherwise, try to pull the image...
-		if err = e.Pull(config.Image, nil); err != nil {
+		if err = e.Pull(config.Image, authConfig); err != nil {
 			return nil, err
 		}
 		// ...And try again.
-		if id, err = client.CreateContainer(&dockerConfig, name); err != nil {
+		if id, err = client.CreateContainer(&dockerConfig, name, nil); err != nil {
 			return nil, err
```
MHBauer: Do we want to use `authConfig` here when creating the container? Or up above? Otherwise, what is the point of the godep update?

MHBauer:
I like that there are examples provided. I don't want to lose them in the commit history. Is there a place in the documentation an example could be placed, would it be better placed in base docker documentation?

I'm not familiar with the godep update process. I'll leave others to comment on that. I do like that it is done in it's own commit.

It would be nice to separate the 'create' support of authconfig from the `push/pull`  support of authconfig, at least by commits. That way we can be sure everything is right in separation.

ezrasilvera:
@dhiltgen if we change dockerclient we might as well add the auth into dockerclient.ContainerConfig  and then we would be able to eliminate 99% of the changes in swarm. Basically most of the changes are just passing this additional info that as I said can be avoided.
We would need just a very small change in the handlers.go to populate that structure

cluster/engine.go
dhiltgen: The engine itself currently does not pull on CreateContainer API calls.  Swarm is the only place where this logic is implemented.

The reason I updated dockerclient was to make sure that when the incoming AuthConfig is unmarshaled at https://github.com/docker/swarm/pull/1384/files#diff-379151c15066f365334797e959e6ea94R435 that the registrytoken isn't dropped.  I suppose an alternative would have been to pass the base64 encoded header around, but that felt like a bigger deviation from the current flow.

dhiltgen:
@ezrasilvera I believe the intention in the dockerclient is to keep ContainerConfig consistent with the paylod passed in to http://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#create-a-container which does not include Auth information.  Auth information is passed in through the header.  The other prior examples in the client were pull/push, which passed this as a top-level argument, so I followed that pattern.

dhiltgen:
I've added the examples to the swarm-api docs.  If there's a better place for this, please let me know.

dongluochen:
LGTM. The example is helpful. I think we may merge this one. @ezrasilvera can make changes from it. ping @docker/swarm-maintainers.

jimmyxian:
@dhiltgen Thanks for doing this.
Can you upgrade `godep` cmd and then update dockerclient. 
The latest `godep` will remove `_test` file.
See this : https://github.com/docker/swarm/issues/1319

dhiltgen:
Vendor change fixed to remove the _test files

abronan:
LGTM

vieux:
LGTM

pdevine:
awesome possums!

Issue #1379, opened 2015-11-04
Does spread strategy will  consider  the status of container ?


chengjingtao:
I am using the spread strategy and it is very great .

But I have this scene:

```
Containers:  15
Nodes: 2
 docker01: ***.***.***.***:2376
  └ Containers: 10                                 //    only 1 is running   
  └ Reserved CPUs: 0 / 4
  └ Reserved Memory:2.44GB /  4.052 GiB           //   10*250M
 docker02: ***.***.***.***:2376
  └ Containers: 5                                   //   4  is running 
  └ Reserved CPUs: 0 / 4
  └ Reserved Memory: 1.22GB  /  4.052 GiB            //  5 * 250M

```

I prefer to create new container on `docker01` ,because it has less running containers.
Is there any way to  solve the question from the viewpoint of strategy ?

vieux:
I think it's a good idea, @aluzzardi @abronan  ?

abronan:
Agreed, we should be able to restrict the strategy to be applied only to running containers on each nodes.

dongluochen:
Similar to network load balancing, these are 2 different strategies. The first one is round robin which distributes tasks to every node (sequentially). The second one is least connections where the one with least active connections is selected. I think both are useful.

aluzzardi:
I'm not sure - stopped containers may come back to life at any point.

In reality, I think they won't in most cases.

ezrasilvera:
@aluzzardi how about doing two stage scheduling. We first bring all the Node candidates and take into account all containers (stopped and running), we then go over that list and prioritized it according to free resources (e.g., not including stopped containers). 
BTW, we can extend this to inspect actual usage on the second loop so we can even take into account underutilized Nodes.

amitshukla:
@aluzzardi I also suggest that restarting a container should actually reschedule a container rather than restart it on the same node it was stopped on.

nishanttotla:
Closing due to lack of activity. Please reopen if you wish to continue discussing it.

PR #1397, opened 2015-11-10
Suggested implementation for  #1379

doronp:
Added optimistic flag which means weight of nodes will be calculated only with running container for CPU share and memory count thus achieving the desired effect.

I recon extensive testing and review is required.

Thanks 
Doron Podoleanu

### DIFF
--- a/scheduler/strategy/weighted_node.go
+++ b/scheduler/strategy/weighted_node.go
@@ -1,6 +1,8 @@
 package strategy
 
 import (
+	"math"
+
 	"github.com/docker/swarm/cluster"
 	"github.com/docker/swarm/scheduler/node"
 )
@@ -24,16 +26,33 @@ func (n weightedNodeList) Swap(i, j int) {
 }
 
 func (n weightedNodeList) Less(i, j int) bool {
+
 	var (
-		ip = n[i]
-		jp = n[j]
+		ip        = n[i]
+		jp        = n[j]
+		threshold = 11.0
 	)
 
 	// If the nodes have the same weight sort them out by number of containers.
 	if ip.Weight == jp.Weight {
+		delta := math.Abs(float64(len(ip.Node.Containers) - len(jp.Node.Containers)))
+		if delta < threshold {
+			return countRunningContainers(ip.Node) < countRunningContainers(jp.Node)
+		}
 		return len(ip.Node.Containers) < len(jp.Node.Containers)
 	}
 	return ip.Weight < jp.Weight
+}
+
+func countRunningContainers(node *node.Node) int {
+	var c = 0
+	for _, container := range node.Containers {
+		if container.Status == "running" {
+			c++
+		}
+	}
+	return c
+
 }
 
 func weighNodes(config *cluster.ContainerConfig, nodes []*node.Node, healthinessFactor int64) (weightedNodeList, error) {
--- a/scheduler/strategy/binpack_test.go
+++ b/scheduler/strategy/binpack_test.go
@@ -67,8 +67,8 @@ func TestPlaceEqualWeight(t *testing.T) {
 	assert.Equal(t, node.UsedMemory, int64(3*1024*1024*1024))
 
 	// check that the last container ended on the node with the highest number of containers
-	assert.Equal(t, node.ID, nodes[1].ID)
-	assert.NotEqual(t, len(nodes[0].Containers), len(nodes[1].Containers))
+	assert.Equal(t, node.ID, nodes[0].ID)
+	//	assert.NotEqual(t, len(nodes[0].Containers), len(nodes[1].Containers))
 
 }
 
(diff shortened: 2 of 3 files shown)


cli/flags.go
```diff
@@ -90,6 +90,10 @@ var (
 		Usage: "placement strategy to use [" + strings.Join(strategy.List(), ", ") + "]",
 		Value: strategy.List()[0],
 	}
+	flOptimistic = cli.BoolFlag{
+		Name:  "optimistic",
```
dongluochen: nit: I prefer a straight-forward name, like  `ignore-inactive`.

doronp:
Ping @dongluochen. Changed the name according to your comment. Had a  messy merge/re-base :-( . Please let me know what you need so we can wrap this up.

cli/flags.go
```diff
@@ -110,6 +110,10 @@ var (
 		Usage: "placement strategy to use [" + strings.Join(strategy.List(), ", ") + "]",
 		Value: strategy.List()[0],
 	}
+	flIgnoreInactive = cli.BoolFlag{
+		Name:  " ignoreInactive",
```
dongluochen: Any reason there is an empty space in the flag name " ignoreInactive"?

cli/flags.go
```diff
@@ -110,6 +110,10 @@ var (
...
+		Usage: "Do not account for non running containers when scheduling",
```
dongluochen: This may need clarification. Swarm always accounts for node port occupation, reserved memory/CPU, label, etc.

scheduler/strategy/weighted_node.go
```diff
@@ -71,3 +73,58 @@ func weighNodes(config *cluster.ContainerConfig, nodes []*node.Node) (weightedNo
...
+
+		// Skip nodes that are smaller than the requested resources.
+		if nodeMemory < int64(config.Memory) || nodeCpus < config.CpuShares {
+			continue
+		}
+
+		var (
+			cpuScore         int64 = 100
+			memoryScore      int64 = 100
+			notRunningSahres int64
```
dongluochen: `notRunningSahres` -> `notRunningShares`

scheduler/strategy/weighted_node.go
```diff
@@ -71,3 +73,58 @@ func weighNodes(config *cluster.ContainerConfig, nodes []*node.Node) (weightedNo
...
+				notRunningSahres += container.Config.CpuShares
+				notRunningMemory += container.Config.Memory
+
+				//				notRunningSahres += container.Info.Config.CpuShares
+				//				notRunningMemory += container.Info.Config.Memory
+			}
+
+		}
+
+		log.Debug("Deducing cpu shres", notRunningSahres)
```
dongluochen: unnecessary log

scheduler/strategy/weighted_node.go
```diff
@@ -71,3 +73,58 @@ func weighNodes(config *cluster.ContainerConfig, nodes []*node.Node) (weightedNo
...
+		log.Debug("Deducing memory shres", notRunningMemory)
+
+		if config.CpuShares > 0 {
+			cpuScore = (node.UsedCpus + config.CpuShares - notRunningSahres) * 100 / nodeCpus
```
dongluochen: This causes over-commit problem. That means the stopped containers cannot start (at least with the resource reservation). We need some consensus on this.

dongluochen:
The current implementation could lead to overcommit issue. For example, it may schedule a container with 1GB memory requirement on a node where all its memory has been assigned to stopped containers.  I don't think Engine would accept such request. Even it can, that means stopped containers may not start with resource guarantee. 

A slight change is to allow such scheduling where there is no overcommit. 

/cc @docker/swarm-maintainers.

doronp:
Fixed the typos and unnecessary log.
Regarding the over commit issue and the clarification - According to:
https://github.com/docker/swarm/issues/1379

I interpreted that:
A- It should be well documented what this flag means.
B- One should use it if he think that in his cluster there are slim to none chance of stopped containers getting back to running state.  So kind of use at your own risk. 

WDYT?

vieux:
I agree with @dongluochen and I think it could lead to many issues and I won't see myself using it.
@abronan @jimmyxian @chanwit what do you guys think ?

doronp:
@vieux 
I can change it to @ezrasilvera s suggestion in https://github.com/docker/swarm/issues/1379
That will eliminate your concerns I think?

doronp:
@dongluochen cc

doronp:
Changed approach: 

As long as the number of containers are about the same in two nodes (delta 11 for now) favor the node with less running containers.

(No flags and the change is only in Less method in 
scheduler/strategy/weighted_node.go)

Later the fixed threshold could be derived and amended during the cluster lifetime   

WDYT?

@vieux @dongluochen @abronan @jimmyxian @chanwit

doronp:
Undesired approach - See: https://github.com/docker/swarm/issues/2015

PR #1409, opened 2015-11-12
typos in swarm networking docs

philwhln:


docs/networking.md
```diff
@@ -55,8 +55,8 @@ scope driver.
 As you can see here, the ID is the same on the two nodes, because it's the same
 network.
 
-If you want to want to create a local scope network (for example with the bridge
-driver) you should use `<node>/<name>` otherwise your network will be created on a
+If you want to create a local scope network (for example with the bridge
```
MHBauer: 'locally scoped network'? 'network with local scope'? Do either of those read better for anyone? Maybe I'm parsing it wrong, but it sounds weird to me.

MHBauer: maybe fix 'global stop driver' up above in the same fashion as this if it is deemed an acceptable change. I'm still not sure if the phrase is odd or not.

abronan:
Thank you @philwhln 

Agreed with @MHBauer, I think `network with local scope` fits better. Same for `global`. If you can take this into consideration that would be great.

abronan:
I took your commit in #1473 to add some more improvements and details. It's now merged. Thanks for the contribution @philwhln!

Issue #1353, opened 2015-10-29
swarm join heartbeat should be an interval


aluzzardi:
When starting a bunch of machines at the same time it hits hard the store.

dongluochen:
The problem is synchronized registration when nodes are added in batch mode. Adding a random delay at `swarm join` should reduce the spike to store.

abronan:
Closed by #1410

PR #1410, opened 2015-11-13
Add a random delay to avoid synchronized registration at swarm join. Simple fix for #1353

dongluochen:

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -36,7 +36,7 @@ var (
 			Name:      "join",
 			ShortName: "j",
 			Usage:     "join a docker cluster",
-			Flags:     []cli.Flag{flJoinAdvertise, flHeartBeat, flTTL, flDiscoveryOpt},
+			Flags:     []cli.Flag{flJoinAdvertise, flHeartBeat, flTTL, flJoinRandomDelay, flDiscoveryOpt},
 			Action:    join,
 		},
 	}
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -32,6 +32,11 @@ var (
 		Usage:  "Address of the Docker Engine joining the cluster. Swarm manager(s) MUST be able to reach the Docker Engine at this address.",
 		EnvVar: "SWARM_ADVERTISE",
 	}
+	flJoinRandomDelay = cli.StringFlag{
+		Name:  "joindelay",
+		Value: "0s",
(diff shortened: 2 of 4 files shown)


MHBauer:
I feel like this change merits a warning somewhere saying that registration might not occur until after `heartbeat` amount of time. This could be surprising if the heartbeat is minutes or longer. (I do admit to not knowing what any typical heartbeat is in actual deployments.)

dongluochen:
@MHBauer I was thinking about it in the first commit.

vieux:
LGTM

ping @aluzzardi 

@dongluochen can you please rebase ?

aluzzardi:
I agree with @MHBauer - this can be surprising, especially with longer HBs.

Thoughts?

cpuguy83:
Wouldn't it be better to handle this in the manager so that it's not overwhelmed when the thundering herd comes in?

cpuguy83:
Either way there is still a delay in registration, but at least then it's only happening during high volume instead of all the time.

dongluochen:
@cpuguy83 `swarm join` is not controlled by swarm manager. For example, a script starting 1k nodes, or a synchronized recovery from failure would cause spikes to registry (hub or KV store).  

@aluzzardi Maybe we can take the minimum of 10s and heartbeat to avoid a long wait. The random delay would be between 0 and min(10s, heartbeat).

cpuguy83:
@dongluochen Oh, I see.

aluzzardi:
What if we have a different set of flags to control that, which default to `0s` (no wait)?

dongluochen:
@aluzzardi Right that would give user control. I'll update the PR.

vieux:
@aluzzardi pling

aluzzardi:
Since this is a flag of `join`, maybe just `--delay`?

As in, `join --joindelay` seems like too many joins

vieux:
LGTM

abronan:
@dongluochen Can you update the PR to add a warning or a docs section to explain why we would need `--delay` as suggested by @MHBauer? Otherwise LGTM.

abronan:
Tracked by #1353

dongluochen:
Thanks @abronan. I updated docs/discovery.md.

abronan:
LGTM

dongluochen:
ping @docker/swarm-maintainers.

PR #1414, opened 2015-11-14
fix event deletion under lock

vieux:

### DIFF
--- a/api/events.go
+++ b/api/events.go
@@ -90,16 +90,18 @@ func (eh *eventsHandler) Handle(e *cluster.Event) error {
 			f.Flush()
 		}
 	}
-
 	eh.RUnlock()
-
+	eh.Lock()
 	if len(failed) > 0 {
 		for _, key := range failed {
 			if ch, ok := eh.cs[key]; ok {
 				close(ch)
 			}
+			delete(eh.cs, key)
 		}
 	}
+
+	eh.Unlock()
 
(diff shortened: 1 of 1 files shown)


MHBauer:
What does this do that https://github.com/docker/swarm/pull/1414/files#diff-66d73e49acd1e88edc3f9f514d0a8f4cR58 doesn't? (I hope that link came out right, but L#58 in the same file.)

aluzzardi:
@abronan @dongluochen Hey, can you guys try to review this PR?

It's quite tricky but we think this solves a panic.

abronan:
@MHBauer It just fixes a very rare race when an event channel can be deleted before we attempt to close it (but go panics when you try to close an already closed channel).

LGTM (sorry for the delay, the *DO NOT MERGE* didn't helped)

The fixes a panic but not sure #1426 is quite the same issue. 

This can be safely included though.

MHBauer:
I think I see it now. There's nothing that says we're not already in the middle of `handle`ing an event before we `handle` the next one. Concurrent programming is hard.

abronan:
ping @dongluochen

dongluochen:
LGTM.

jimmyxian:
LGTM

agsmorodin:
Hi,

In which swarm release can I expect to see this fix?  thanks.

Issue #1402, opened 2015-11-11
Network connect multihost env | invalid container <nil> :


cerias:
If i try to connect an container to an network it's works only on the same host as the master is. When i connect to an other host direct to docker i can connect the container to the network and everything works as excepted.

Steps to reproduce:
```
# eval "$(docker-machine env --swarm docker-app1)"
# docker ps
CONTAINER ID        IMAGE                 COMMAND                  CREATED             STATUS              PORTS               NAMES
243268f57f74        elasticsearch:2.0.0   "/docker-entrypoint.s"   48 minutes ago      Up 48 minutes                           docker-app1/es3
19b319ccfd75        elasticsearch:2.0.0   "/docker-entrypoint.s"   49 minutes ago      Up 49 minutes                           docker-app2/es1
b9d355311060        elasticsearch:2.0.0   "/docker-entrypoint.s"   49 minutes ago      Up 49 minutes                           docker-app2/es2

# network ls
NETWORK ID          NAME                 DRIVER
b85370d450ec        docker-app1/host     host                
fa3f0425c192        test                 overlay             
cc1f7df9d179        docker-app2/bridge   bridge              
ceddf908f39d        docker-app2/none     null                
ef2d1f54becb        docker-app2/host     host                
15d19af8965b        docker-app1/bridge   bridge              
2e13fa81d306        docker-app1/none     null   

# docker network connect test es3
# docker network connect test es2
Error response from daemon: invalid container <nil> : nosuchcontainer: no such id: es2
# eval "$(docker-machine env  docker-app2)"
# docker network connect test es2

```

Inspecting the Network delivers only for the current host the connected containers.
```
[
    {
        "Name": "test",
        "Id": "fa3f0425c19243cc84c6f0ffd73a754f6f86ef8380e519f466eac7259b4dfcf2",
        "Scope": "global",
        "Driver": "overlay",
        "IPAM": {
            "Driver": "default",
            "Config": [
                {}
            ]
        },
        "Containers": {
            "b9d3553110600e558f7963fddcd8d737a8d694cb8d6d6e68196c9a2969fff716": {
                "EndpointID": "f3a82c84c451018d365fd567c5767ea62d14d3151f108edcd4e3f1d48f239e52",
                "MacAddress": "02:42:0a:00:01:03",
                "IPv4Address": "10.0.1.3/24",
                "IPv6Address": ""
            }
        },
        "Options": {}
    }
]

```

All machines was installed by docker-machine. 

System informations:
docker info
[...]

Issue #889, opened 2015-06-02
Network integration tests


aluzzardi:
We currently have network integration tests (`test/integration/port-filter.bats`) that verify host mode networking (contributed by @chanwit).

We should add tests to verify `bridge` mode as well

chanwit:
@aluzzardi let me know if you've got something else needed to test.

PR #1438, opened 2015-11-20
Route network connect/disconnect request to the right engine.

dongluochen:
In swarm, `docker network connect overlaynetwork *container*` should be routed to the engine with  *container*. This issue is tracked by #1402. 

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -24,6 +24,16 @@ import (
 // APIVERSION is the API version supported by swarm manager
 const APIVERSION = "1.21"
 
+// A simple buffer reader to copy http.Request.Body stream
+type bufReader struct {
+	*bytes.Buffer
+}
+
+// Implements io.ReadCloser interface
+func (bf bufReader) Close() error {
+	return nil
+}
+
 // GET /info
 func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
 	info := dockerclient.Info{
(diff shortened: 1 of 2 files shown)


api/handlers.go
```diff
@@ -24,6 +24,16 @@ import (
 // APIVERSION is the API version supported by swarm manager
 const APIVERSION = "1.21"
 
+// A simple buffer reader to copy http.Request.Body stream
+type bufReader struct {
+	*bytes.Buffer
+}
+
+// Implements io.ReadCloser interface
+func (bf bufReader) Close() error {
```
MHBauer: This is to satisfy the golang http.Request body, which is an io.ReadCloser.

See https://golang.org/pkg/net/http/#Request

api/handlers.go
```diff
@@ -782,6 +792,41 @@ func proxyVolume(c *context, w http.ResponseWriter, r *http.Request) {
...
+	if network == nil {
+		httpError(w, fmt.Sprintf("No such network: %s", networkid), http.StatusNotFound)
+		return
+	}
+	// Set the network ID in the proxied URL path.
+	r.URL.Path = strings.Replace(r.URL.Path, networkid, network.ID, 1)
+
+	// make a copy of r.Body
+	buf, _ := ioutil.ReadAll(r.Body)
+	copy1 := bufReader{bytes.NewBuffer(buf)}
```
MHBauer: We don't have a copy2 or any other variable with copy. This var name can drop the number. Maybe `bodyCopy`?

MHBauer:
I don't like the way the whole `bufReader` concept feels, but I do not know any other way to do it. Golang seems to be complicating the issue a bit.

The container is pulled out of the request, so it should go to the right place, instead of being scheduled normally.

dongluochen:
@MHBauer Thank you! I agree with you it's not very clean. Another way is to not `proxy` such requests, but reconstruct it. It'd require more functions to handle different operations. 
We may have more cases like this. @docker/swarm-maintainers, any recommendation for this problem?

dongluochen:
@MHBauer ioutil.NopCloser can be used to make copy of http.request.Body. There is no need to create  `bufReader`. Thanks for your comment!

MHBauer:
@dongluochen Very cool! I am happy to learn something about go. The diff and the updated code both look _nicer_ and _cleaner_ to me. It's not any kind of technical measure, but I like it better.

abronan:
We might want to update the `connect/disconnect` test in the Integration tests to make sure we don't see any regression in the future (like running a container on node-02 and make sure the connect request is routed to this node)

@dongluochen Can you take care of that? Thanks!

Other than that LGTM.

dongluochen:
@abronan Thanks for the feedback! I'll update `network connect/disconnect` integration test.

aluzzardi:
Should we merge this and make the overlay network testing a separate issue?

I think that is already tracked by #889

abronan:
Yeah let's merge it, I'll handle the tests.

@dongluochen Can you please rebase? (new github behavior is really annoying :sob:)

abronan:
Thanks @dongluochen,

CI behavior is still super erratic though... Tested on a local cluster instead.

@docker/swarm-maintainers PTAL

Issue #1442, opened 2015-11-24
Swarm HTTP API and CORS not working for other verbs than GET


Bahaika:
Hi,

It seems there is a bug in the implementation of the --cors option. I'm currently working on a web interface to manage a swarm cluster, and I consume the HTTP API using AngularJS.
I started the swarm manager with the `--cors` option, and I can use it for all the GET requests. But when it comes to the POST requests, nothing works anymore :

```
XMLHttpRequest cannot load http://172.31.8.142:3375/networks/create.
Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.
Origin 'http://localhost:9000' is therefore not allowed access. The response had HTTP status code 404.
```

The thing is, it didn't do a POST request, but an OPTIONS request.
So, I started digging a little. When in a CORS case, if you try to request with something else than GET, the browser will always asks the server if it can do the request using a OPTIONS request. The server is supposed to respond with headers to allow the browser to emit the real request :

```
Access-Control-Allow-Origin: *
Access-Control-Allow-Methods: POST, PUT, DELETE, PATCH
```

It would be nice to implement this in the swam manager.

MHBauer:
I'm going to guess this line, https://github.com/docker/swarm/blob/master/api/primary.go#L93 needs to be `"/"` or `"/*"`. I am going to try to write a test.

PR #1446, opened 2015-11-24
OPTIONS response when cors is enabled

MHBauer:
I am not a CORS expert, but I think this might help with #1442. From my limited understanding, there needs to be an OPTIONS response for every method/endpoint combo. I don't know how to get gorilla mux to just respond to everything, so I added an options response for every registered method. Probably overkill, but I think it might work.

I couldn't get any response to the OPTIONS method that was already defined, so I removed it.

I pulled out the setup of the router to make it easier to unit test directly. 

Added a super basic unit test with print statements to see how it works if you run `test -v`.

### DIFF
--- a/api/primary.go
+++ b/api/primary.go
@@ -91,9 +91,6 @@ var routes = map[string]map[string]handler{
 		"/networks/{networkid:.*}": deleteNetworks,
 		"/volumes/{name:.*}":       deleteVolumes,
 	},
-	"OPTIONS": {
-		"": optionsHandler,
-	},
 }
 
 func writeCorsHeaders(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


aluzzardi:
Awesome - thanks @MHBauer !

@Bahaika Could you try this PR and see if it fixes your problems?

Thanks

Bahaika:
Thank you @MHBauer. That was fast.
I'll try it as soon as possible and let you know if it works.

vieux:
@Bahaika were you able to try to see if it's fixes your issue ?

vieux:
@MHBauer please rebase :)

vieux:
LGTM

abronan:
LGTM

PR #1449, opened 2015-11-25
Improving error output for bad swarm mesos user

jimenez:
Signed-off-by: Isabel Jimenez <contact@isabeljimenez.com>

### DIFF
--- a/cluster/mesos/task.go
+++ b/cluster/mesos/task.go
@@ -184,7 +184,11 @@ func (t *task) monitor() (bool, []byte, error) {
 	case mesosproto.TaskState_TASK_FINISHED:
 		return true, taskStatus.Data, nil
 	case mesosproto.TaskState_TASK_FAILED:
-		return true, nil, errors.New(taskStatus.GetMessage())
+		errorMessage := taskStatus.GetMessage()
+		if errorMessage == "Abnormal executor termination" {
+			errorMessage += " : please verify your SWARM_MESOS_USER is correctly set"
+		}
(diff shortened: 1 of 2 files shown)


cluster/mesos/task.go
```diff
@@ -184,7 +184,11 @@ func (t *task) monitor() (bool, []byte, error) {
 	case mesosproto.TaskState_TASK_FINISHED:
 		return true, taskStatus.Data, nil
 	case mesosproto.TaskState_TASK_FAILED:
-		return true, nil, errors.New(taskStatus.GetMessage())
+		errorMessage := taskStatus.GetMessage()
+		if errorMessage == "Abnormal executor termination" {
+			errorMessage += " : please verify your SWARM_MESOS_USER is correctly set"
```
tnachen: This seems a bit too generic for me. Users will get this error message for a variety of reasons, not just when a user cannot be used.

jimenez: I had the same thought but this is a quick improvement for a most known occurring error. Also this practice can be found in other places in Docker: https://github.com/docker/docker/blob/master/api/client/utils.go#L104

jimenez:
@vieux ? thoughts?

vieux:
LGTM

sorry about the delay @jimenez

cluster/mesos/task.go
tnachen: Please add a comment here at least then so that we know this is a blanket assumption we have and is actually masking other possible errors, also a todo to fix this.

Issue #1452, opened 2015-11-25
Fix CI env with latest docker master


abronan:
We test against `docker master` so changes are happening really fast. There are a few tests failing that just needs to be updated or need some attention (I see those running the tests locally):

- [x] `docker import - check error code`
- [x] `docker ps --before`
- [x] `reschedule with soft-image-affinity and node constraint`

Please make sure to pull the latest fix with the updated `swarm-test-env` for the CI first.

ping @docker/swarm-maintainers

abronan:
Closed by #1453

PR #1453, opened 2015-11-26
Fix CI

jimmyxian:
Related to: https://github.com/docker/swarm/issues/1452

### DIFF
--- a/test/integration/api/import.bats
+++ b/test/integration/api/import.bats
@@ -47,6 +47,7 @@ function teardown() {
 	swarm_manage
 
 	temp_file=$(mktemp)
+	echo abc > $temp_file
 
 	run docker_swarm import - < $temp_file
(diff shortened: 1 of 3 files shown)


abronan:
Was wondering what was going on with the `ps --before`, somehow my test env didn't report the deprecation warning...

Thanks for investigating @jimmyxian 

First two failures are unrelated, seems like there is a race in the discovery where we issue a request before the nodes are getting registered thus we get a `No healthy node available`.

jimmyxian:
@abronan I download the docker master binary. The result of `ps --before` include a Warning messsage:
`Warning: '--before' is deprecated, it will be removed soon. See usage`

abronan:
Ok CI looks way better now. Will look at the discovery thing separately.

LGTM

Thanks @jimmyxian

PR #1460, opened 2015-11-28
fix list networks with filter {name: <overlay_network_name>}

BSWANG:
fix list networks with filter {name: {overlay_network_name}} fail
such as: 
<code>curl --globoff  http://host:port/v1.21/networks\?filters\="{\"name\":[\"multi-host-name\"]}"</code>

fix fail to inspect {node}/{overlay_network_name}
such as:
<code>docker inspect node1/multi-host-network</code>

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -282,7 +282,11 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Apply filters.
-		if !filters.Match("name", strings.TrimPrefix(container.Names[0], "/")) {
+		if len(container.Names) > 0 {
+			if !filters.Match("name", strings.TrimPrefix(container.Names[0], "/")) {
+				continue
+			}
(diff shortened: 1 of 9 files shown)


vieux:
@BSWANG can you please add a test ?

BSWANG:
@vieux another bug (network filter) is rest API, I don't know how do test for rest API with bats.

vieux:
@BSWANG after reading more carefully your comment, I think it's the expected behavior.

global scope networks aren't attached to a particular host, the are on the whole cluster, so to me it make no sens to support <node>/<network> with global scope networks 

@abronan ?

BSWANG:
@vieux If it's a network on the whole cluster, the network's containers information should be combine of the nodes', but now is just containers of one node.

vieux:
@BSWANG the fix for inspect is here: https://github.com/docker/swarm/pull/1547

PR #1463, opened 2015-11-29
Ensure CloseWrite is called for hijacked TLS conns

### LINKED ISSUE #1430: stdin close event is not transferred to a docker node on TLS
Swarm does not transfer "stdin close" event as TCP half close to docker node daemon when TLS is enabled.
The reason is what "crypto/tls" does not have CloseWrite() method to send half close.
"docker" package solved the issue by implementing their own CloseWrite method(and "tlsDial", "dial")
   https://github.com/docker/docker/blob/master/api/client/hijack.go
How about using the same way to handle the issue ?

```
$ eval $(docker-machine env master)
$ echo aaa | docker run -i ubuntu cat |grep aaa
aaa
$ eval $(docker-machine env --swarm master)
$ echo aaa | docker run -i ubuntu cat |grep aaa
aaa
^C^C^C^C
^C^C^C^C
^C^C^C
Connection to docker.paiza.io closed.
```

cpuguy83:
Golang's `*tls.Conn` does not support `CloseWrite`, this means that
connections using TLS will not be able to properly close on hijacked
connections.

This copies Go's tls.Dial and instead returns an internal
`tlsClientConn` type that does store the raw net.Conn and implements
`CloseWrite`.
Implementation is mostly copied from
`github.com/docker/docker/api/client/hijack.go`

Test for this is difficult -- at least without introducing more hacks than this already is.
Verified the functionality manually by inserting a Debug message in the new `CloseWrite` function.
Note that there is a closed Golang issues: golang/go#8579
I pinged @rsc on it to see if we can do something about this, may open a separate issue.

Fixes #1430

### DIFF
--- a/api/utils.go
+++ b/api/utils.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"time"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/swarm/cluster"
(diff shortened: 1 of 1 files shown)


cpuguy83:
Something wrong with token discovery?

vieux:
@cpuguy83 yes

vieux:
LGTM ping @aluzzardi

MHBauer:
Is there a way we can reuse the code from the base docker engine? I see that the hijacker and dialer aren't exposed (being in api/client). Should this kind of functionality be made into a pkg? Or is it best to wait for a golang fix?

aluzzardi:
@MHBauer Yes, it makes total sense to reuse the code from the engine.

The engine maintainers are working on splitting up the API from the rest of the engine, so it will make it much easier to re-use here.

aluzzardi:
Thanks @cpuguy83 !

LGTM

aluzzardi:
@cpuguy83 Could you please rebase?

cpuguy83:
Rebased

Issue #985, opened 2015-06-20
Unstarted container causes ps -a to panic


bfirsh:
Running `docker ps -a` against my Swarm causes a panic:

```
2015/06/20 22:28:37 http: panic serving 162.245.20.143:27042: runtime error: index out of range
goroutine 117 [running]:
net/http.func·011()
	/usr/src/go/src/pkg/net/http/server.go:1100 +0xb7
runtime.panic(0x963280, 0xcd6a3c)
	/usr/src/go/src/pkg/runtime/panic.c:248 +0x18d
github.com/docker/swarm/api.getContainersJSON(0xc2080b0f00, 0x7f7ee13bfc18, 0xc208496500, 0xc208161ee0)
	/go/src/github.com/docker/swarm/api/handlers.go:199 +0x16e7
github.com/docker/swarm/api.func·008(0x7f7ee13bfc18, 0xc208496500, 0xc208161ee0)
	/go/src/github.com/docker/swarm/api/primary.go:111 +0x2a0
net/http.HandlerFunc.ServeHTTP(0xc20803d300, 0x7f7ee13bfc18, 0xc208496500, 0xc208161ee0)
	/usr/src/go/src/pkg/net/http/server.go:1235 +0x40
github.com/gorilla/mux.(*Router).ServeHTTP(0xc2080add60, 0x7f7ee13bfc18, 0xc208496500, 0xc208161ee0)
	/go/src/github.com/docker/swarm/Godeps/_workspace/src/github.com/gorilla/mux/mux.go:98 +0x292
github.com/docker/swarm/api.(*dispatcher).ServeHTTP(0xc2080be910, 0x7f7ee13bfc18, 0xc208496500, 0xc208161ee0)
	/go/src/github.com/docker/swarm/api/server.go:32 +0x9d
net/http.serverHandler.ServeHTTP(0xc2081372c0, 0x7f7ee13bfc18, 0xc208496500, 0xc208161ee0)
	/usr/src/go/src/pkg/net/http/server.go:1673 +0x19f
net/http.(*conn).serve(0xc208459780)
	/usr/src/go/src/pkg/net/http/server.go:1174 +0xa7e
created by net/http.(*Server).Serve
	/usr/src/go/src/pkg/net/http/server.go:1721 +0x313
2015/06/20 22:28:49 http: panic serving 162.245.20.143:23937: runtime error: index out of range
goroutine 119 [running]:
net/http.func·011()
[...]

PR #1465, opened 2015-11-30
fix the panic when "ps -a" with dead containers

### LINKED ISSUE #1421: Dead containers break docker ps -a output
I have the exact same issue as mentioned in https://github.com/docker/swarm/issues/985#issuecomment-156476578

If there's a `DEAD` container in the swarm I can no longer use `docker-compose` or `docker ps -a`.

To work around this issue, I need to login to every swarm machine and find the `DEAD` containers:

```
root@tim-a1:~# docker ps -a
CONTAINER ID        IMAGE                                                         COMMAND                   CREATED             STATUS                        PORTS               NAMES
f1be81586920        tutum/haproxy:latest                                          "python /haproxy/main"    42 minutes ago      Exited (137) 13 minutes ago                       feuerwehrblue_khfhaproxy_1
[....]
7061ef7b78d5        phundament/php:5.6-fpm-4.4.1-dev                              "/bin/sh -c 'echo \"up"   56 minutes ago      Dead                                              
677f84c889e9        swarm                                                         "/swarm --debug join "    About an hour ago   Up About an hour              2375/tcp            swarm-agent
```

Container removal throws an error...

```
root@tim-a1:~# docker rm -fv 7061
Error response from daemon: Driver aufs failed to remove init filesystem 7061ef7b78d573583126e92e814f414b6dcfedec5d4fa930b84b168cf3708a06-init: rename /var/lib/docker/aufs/mnt/7061ef7b78d573583126e92e814f414b6dcfedec5d4fa930b84b168cf3708a06-init /var/lib/docker/aufs/mnt/7061ef7b78d573583126e92e814f414b6dcfedec5d4fa930b84b168cf3708a06-init-removing: device or resource busy
Error: failed to remove containers: [7061]
```

... but looks like it's gone now

```
root@tim-a1:~# docker rm -fv 7061
Error response from daemon: no such id: 7061
Error: failed to remove containers: [7061]
```

... and my swarm is working as expected afterwards.

Running swarm 1.0.0 and docker 1.9.0

### LINKED ISSUE #1436: Return EOF when ps -a
"docker ps" and "docker ps -a" will be OK, "docker -H swarm-manager:2376 ps" is OK also, but if input "docker -H swarm-manager:2376 ps -a", it will report an error:
An error occurred trying to connect; Get http://swarm-manager:2376/v1.21/containers/json?all=1; EOF

mountkin:
Fixes #1436 
Fixes #1421
If it fails after https://github.com/docker/docker/blob/master/daemon/delete.go#L118 on container deletion, the `container.Names` will be empty because the container list API will query the graphdb at https://github.com/docker/docker/blob/master/daemon/list.go#L200

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -282,7 +282,11 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		}
 
 		// Apply filters.
-		if !filters.Match("name", strings.TrimPrefix(container.Names[0], "/")) {
+		if len(container.Names) > 0 {
+			if !filters.Match("name", strings.TrimPrefix(container.Names[0], "/")) {
+				continue
+			}
+		} else if len(filters["name"]) > 0 {
 			continue
 		}
 		if !filters.Match("id", container.Id) {


jimmyxian:
@mountkin Thanks

Related : https://github.com/docker/swarm/issues/985

chanwit:
Yep, there's a case of `len(Names)` being 0 when a container's status is `Dead`.
Is it possible to start a container with `Dead` status? This is quite tricky for integration test, IMO.

mountkin:
@chanwit I think it's quite tricky to add the integration test because we have to manage to create a `Dead` container first :crying_cat_face: 
I opened another PR in the Docker project to reduce the probability of the empty Names problem. https://github.com/docker/docker/pull/18308

abronan:
Thanks @mountkin, also thanks for trying to mitigate the issue on the docker side as well as we shouldn't get an empty container name. I assume it's better to merge and check it on our side to avoid this panic even though we shouldn't get it in the first place. As a safety measure.

And indeed this is going to be tricky for the integration tests, no way to reliably get a `Dead` container AFAIK (same problems applies in docker).

LGTM

vieux:
LGTM

MHBauer:
I know this is merged already, but seeing as this is a bad case, I think it would be good to log some data saying that this case was reached.

abronan:
@MHBauer Do you mean: when using `--debug`, logging that a container has reached the `dead` state?

PR #1466, opened 2015-11-30
Integration: Wait for nodes to be registered and discovered before testing

abronan:
There seems to be a race on the CI only (tests are now fine locally) that involves a race in the discovery. We issue a command before the Manager is ready and registers the Agents correctly. Thus we see random `No Healthy Node available in the cluster` errors on random tests.

Might be a side effect of limited resources on a Jenkins host (only a limited subset of machines seems to be affected).

### DIFF
--- a/test/integration/discovery/consul.bats
+++ b/test/integration/discovery/consul.bats
@@ -99,7 +99,7 @@ function teardown() {
 	swarm_join "$DISCOVERY"
 
 	# Start a manager. It should keep retrying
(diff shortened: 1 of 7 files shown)


abronan:
CI is :green_apple: 

@vieux any reason for disabling discovery? :confused:

abronan:
LGTM

vieux:
LGTMMM :fireworks:

PR #1476, opened 2015-12-02
Zk tests

jimenez:
Updating mesos-go dependency to fix: https://github.com/mesos/mesos-go/issues/201 and added a test.


vieux:
That's a LOT of changes, but it seems to fix the zookeeper issue. 

LGTM

PR #1478, opened 2015-12-02
Add MESOS_OFFER_REFUSE_SECONDS environment configuration

klaus1982:
I'd like to configure offer refuse seconds when launch task. Environment variables is an option to avoid long parameters.


vieux:
ping @jimenez

jimenez:
@klaus1982 I think you're referring here to the already existing offer timeout flag/environment variable. 
Can you please confirm? 
Please see: https://github.com/docker/swarm/blob/master/cli/help.go#L49

klaus1982:
@jimenez nop :). For now, Swarm use all available resources in slave to launch tasks; it'll return un-used resources to mesos, e.g. running one task required on 1 CPU in a slave with 2 CPU, the other CPU will return to Mesos and try to re-offer in next allocation cycle. In allocation, the returned resources has a refused filtered, default value in Mesos-GO is 5s. This ticket is used to provide the parameter for refused filtered when Swarm return unused resources.

cluster/mesos/cluster.go
```diff
@@ -468,7 +469,21 @@ func (c *Cluster) scheduleTask(t *task) bool {
 
 	t.build(n.ID, c.slaves[n.ID].offers)
 
-	if _, err := c.driver.LaunchTasks(offerIDs, []*mesosproto.TaskInfo{&t.TaskInfo}, &mesosproto.Filters{}); err != nil {
+	// Set Mesos refuse seconds by environment variables.
+	var offerFilters *mesosproto.Filters;
+	var refuseSecondsStr string;
+
+	offerFilters = &mesosproto.Filters{};
```
jimenez: Could you remove the var and just use `:=` here please?

jimenez:
My bad, thanks for adding this. I read the PR and made a small comment but I think that more generally, this should be treated as the other flags, since they all can be environment variables. We shouldn't split configuration between flags and environment variables and rather let users decide. We should provide as we've been doing, both options. Could you adapt this PR to make this offer filter part of the flags?

klaus1982:
sure, my pleasure :). I'll follow the practice just as we did for other options; and will address your comments together.

jimenez:
Thanks!

jimenez:
@klaus1982 Could you please try to fix this? I think a rebase should do the trick.

klaus1982:
Addressed comments :).

klaus1982:
@jimenez would you help to review code? If any comments, please let me know :).

jimenez:
Thanks @klaus1982. The rebase is not very clean yet, there are a lot of commits. I'm carrying your commits for the original purpose of this PR (refuse filter) into [1585](https://github.com/docker/swarm/pull/1585). 
Could you please open a second PR for the roles part? meanwhile I'm closing this one.

klaus1982:
@jimenez, Sure, thanks very much :).

PR #1489, opened 2015-12-03
Integration: fix --kernel-memory value set too low

abronan:
Related to docker/docker#18272

`--kernel-memory` value was set too low in the CI, test was passing because setting `--kernel-memory` had no effect until now. As it is fixed in `docker:master`, the test would fail either with `cannot start container: broken pipe` or would crash the Jenkins slave process on the host.


chanwit:
LGTM

jimmyxian:
@abronan Thanks.
```
ok 105 docker run with resources
```
CI works now.
LGTM

chanwit:
:+1:

PR #1490, opened 2015-12-03
Fix retry time when check disconnect events

jimmyxian:
Sometime `check disconnect events` failed because the retry time is short.


abronan:
LGTM thanks @jimmyxian :+1:

abronan:
`ok 36 etcd discovery: check engine connect/disconnect events` :+1:

PR #1497, opened 2015-12-04
emphasize the hosted discovery service is not meant for production

vieux:

### DIFF
--- a/docs/discovery.md
+++ b/docs/discovery.md
@@ -24,10 +24,10 @@ can use.
 
 ### Hosted Discovery with Docker Hub
 
-#####The Hosted Discovery Service is not recomended for production use.
+#####The Hosted Discovery Service is not recommended for production use.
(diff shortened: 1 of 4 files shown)


abronan:
LGTM

dmp42:
LGTM

cc @londoncalling @moxiegirl

docs/discovery.md
```diff
@@ -24,6 +24,11 @@ can use.
 
 ### Hosted Discovery with Docker Hub
 
+#####The Hosted Discovery Service is not recomended for production use.
+#####It's intended to be used for testing/development.
```
dongluochen: Maybe we can add a sentence since testing may DoS the site. `Please do not send high volume requests to this discovery service.`

samalba:
Should we also give an emphasize on one of the k/v store? I am sure people will be confused by which one to go with.

PR #1500, opened 2015-12-05
[DO NOT MERGE] WIP: Retry container create at failure.

dongluochen:
When container create fails, swarm manager should retry create according to preference. This PR has dependency on https://github.com/samalba/dockerclient/pull/194. 

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -151,32 +151,45 @@ func (c *Cluster) createContainer(config *cluster.ContainerConfig, name string, 
 	}
 
 	nodes, err := c.scheduler.SelectNodesForContainer(c.listNodes(), configTemp)
+	c.scheduler.Unlock()
+
 	if err != nil {
-		c.scheduler.Unlock()
 		return nil, err
 	}
-	n := nodes[0]
-	engine, ok := c.engines[n.ID]
-	if !ok {
+
+	for _, n := range nodes {
+		engine, ok := c.engines[n.ID]
+		if !ok {
+			continue
+		}
+
+		c.scheduler.Lock()
+		c.pendingContainers[swarmID] = &pendingContainer{
+			Name:   name,
+			Config: config,
+			Engine: engine,
+		}
 		c.scheduler.Unlock()
-		return nil, fmt.Errorf("error creating container")
+
+		container, err := engine.Create(config, name, true)
+
+		// pendingContainer need to be clear
+		c.scheduler.Lock()
+		delete(c.pendingContainers, swarmID)
+		c.scheduler.Unlock()
+
+		if err == nil {
+			return container, err
+		}
+		log.WithFields(log.Fields{"name": "swarm"}).Infof("container create failed: %v", err.Error())
+
+		if err == dockerclient.ErrConnectionRefused {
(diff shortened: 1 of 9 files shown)


abronan:
@dongluochen Needs a golint check :bowtie:

abronan:
@dongluochen Also this needs a `Godep update` to pick the change in dockerclient :)

cluster/swarm/cluster.go
```diff
@@ -151,32 +151,45 @@ func (c *Cluster) createContainer(config *cluster.ContainerConfig, name string,
...
+
 	if err != nil {
-		c.scheduler.Unlock()
 		return nil, err
 	}
-	n := nodes[0]
-	engine, ok := c.engines[n.ID]
-	if !ok {
+
+	for _, n := range nodes {
```
aluzzardi: There should be a way to limit this I think - we might end up retrying on every node on the cluster, not sure this is what we want

cluster/swarm/cluster.go
```diff
@@ -151,32 +151,45 @@ func (c *Cluster) createContainer(config *cluster.ContainerConfig, name string,
...
+			return container, err
+		}
+		log.WithFields(log.Fields{"name": "swarm"}).Infof("container create failed: %v", err.Error())
 
-	c.scheduler.Lock()
-	delete(c.pendingContainers, swarmID)
-	c.scheduler.Unlock()
+		if err == dockerclient.ErrConnectionRefused {
+			// Mark node as unhealthy
+			engine.SetEngineHealth(false)
```
aluzzardi: I don't think we want to do that here - we should probably let the Engine continue with its normal heartbeating process.

This might be a transient error that occurred only once - `CreateContainer` is not equipped with retries and so on.

A better approach would be to increment the number of failures in `engine.Create` - although that might be overkill?

cluster/swarm/cluster.go
```diff
@@ -151,32 +151,45 @@ func (c *Cluster) createContainer(config *cluster.ContainerConfig, name string,
...
-	container, err := engine.Create(config, name, true)
+		if err == nil {
+			return container, err
+		}
+		log.WithFields(log.Fields{"name": "swarm"}).Infof("container create failed: %v", err.Error())
 
-	c.scheduler.Lock()
-	delete(c.pendingContainers, swarmID)
-	c.scheduler.Unlock()
+		if err == dockerclient.ErrConnectionRefused {
```
aluzzardi: I think there are other classes of errors than `ErrConnectionRefused`.

From what I've seen in production, we might get simple `i/o errors` and so on.

`dockerclient` is able to distinguish between a docker generated error (as in, the error comes from the HTTP **response**) vs a communication error (the error comes from the HTTP client while making the request)

dongluochen:
@aluzzardi @abronan and I had a discussion on this reschedule problem. We decide to use response from dockerclient to fail node fast. I'll put this PR on hold while I work on that.

abronan:
@dongluochen Any news on that PR? Should we go along with that?

If so this needs a rebase :bowtie:

dongluochen:
@abronan Part of the work is done in #1600. I'll close this PR for now.

Issue #1486, opened 2015-12-03
Proposal: Improved Node Management


aluzzardi:
When Swarm fails to connect to a node for the first time, it gives up and never retries.

This was designed to avoid continuously retrying to connect to invalid nodes.

This causes multiple problems (#1185 #1331) such as:
- If the manager is started before a node then it will never see the node
- It's very hard to debug for the user

## Proposal

To fix those problems we could redesign the way Swarm does node management.

This is a possible approach I think we could explore:

- **Always**  add discovered nodes to the cluster

- Do not attempt to synchronously connect to *engines* from *discovery* at all (this will make the discovery code much simpler since we won't need the goroutine anymore)

- Let the Engine *refresh loop*, which currently handles retrying on failure, handle initial connection as well. This means that if an engine is not reachable at start, the refresh loop will try to reconnect after a random delay within range.

- Make sure the rest of the code base can handle unhealthy nodes. Until we make first connection, we do not know the `ID` of the Engine which will be a problem since we index Engines by ID.

- Perhaps we could do something more fancy than the random delay retry such as exponential back-off.

### Reporting

Debugging node registration issues is really difficult for users and requires fiddling with logs. A proposal was already made (#1136) and I think it could be built on top of the improved node management.

Since we'd have a list of **all** nodes (even if we failed to connect to them), we could report all of them in `docker info`, along with their status and error message if any:

```
$ docker info
Nodes: 1
 dev: 10.0.0.123:4242
  └ Status: Down
  └ Error: 10.0.0.123: No route to host
```

This could include extra information such as:
- Last seen (last successful heartbeat)
- Number of retries
- Scheduled next retry

Reporting must also take care of **duplicate IDs**: the problem confuses users very often (#1467). At the very least we should report it in `docker info`.

### Existing Attempts

There have been a couple of attempts to fix this issue (#1044 #1195), however both approaches try to solve the problem at the discovery level rather than the engine level.

The downsides are:
- It duplicates the reconnection problem in both engine and discovery. For instance, now that `Engine` retries to reconnect at a random interval we would have to reimplement that for discovery as well
- It makes the discovery code path more complicated. It's already a beast. If we implement this proposal however, we would actually make it simpler since we don't need to connect to engines in goroutines anymore
- It doesn't provide a way to report the status to the user, since the `Cluster` is not aware of those nodes.

/cc @docker/swarm-maintainers @MHBauer @lokikiller @ahmetalpbalkan @riuvshin

[...]

PR #1501, opened 2015-12-05
Increase default TTL and heartbeat value

### LINKED ISSUE #1485: Revisit default discovery TTL/heartbeat values
Current default discovery TTL values are pretty low and the load is heavy on the kv discovery/token service.

Those values are configurable anyway so we can consider increasing to a higher default value. To be defined.

/cc @docker/swarm-maintainers

abronan:
Increases the default TTL and heartbeat value for discovery.
Because the node will still be listed for a long period on
`docker info`, there is now a Status to know if a node is
in the healthy or unhealthy state.

Fixes #1485 

I think a default 3 minutes TTL and 60 seconds heartbeat is fair for now. We can reconsider and increase that default value when we'll have Node Management proposed in #1486.

Depends on #1500 .Without it, there is a probability that the scheduler will block subsequent requests for up to 60 seconds. Because of the probability to pick the downed node before it being marked as Unhealthy by the refresh loop which can take up to `max_refresh_time`.

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -48,12 +48,12 @@ var (
 	}
 	flHeartBeat = cli.StringFlag{
 		Name:  "heartbeat",
-		Value: "20s",
+		Value: "60s",
 		Usage: "period between each heartbeat",
 	}
 	flTTL = cli.StringFlag{
 		Name:  "ttl",
-		Value: "60s",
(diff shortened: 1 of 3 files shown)


jimmyxian:
LGTM

vieux:
LGTM

PR #1517, opened 2015-12-09
Use failureCount as a secondary health indicator.

dongluochen:
We will use failureCount to record connection failure or other engine unhealthy errors. When failureCount reaches threshold, engine is marked as unhealthy. This can fail engine fast, comparing with normal engine refresh loop. Ideally, failureCount can also be used to lower the priority of a engine in node selection so container creation has a higher success rate. This is under discussion/debate. 

This change is to implement fail/recover fast for Node Management #1486.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -24,9 +24,6 @@ const (
 
 	// Minimum docker engine version supported by swarm.
 	minSupportedVersion = version.Version("1.6.0")
-
-	// Engine failureCount threshold
-	engineFailureCountThreshold = 3
 )
 
 // delayer offers a simple API to random delay within a given time range.
@@ -62,7 +59,7 @@ func (d *delayer) Wait() <-chan time.Time {
 type EngineOpts struct {
 	RefreshMinInterval time.Duration
 	RefreshMaxInterval time.Duration
-	RefreshRetry       int
+	FailureRetry       int
 }
 
 // Engine represents a docker engine
@@ -86,7 +83,7 @@ type Engine struct {
 	client          dockerclient.Client
 	eventHandler    EventHandler
 	healthy         bool
-	failureCount    int64
+	failureCount    int
 	overcommitRatio int64
 	opts            *EngineOpts
 }
@@ -188,6 +185,17 @@ func (e *Engine) IsHealthy() bool {
 	return e.healthy
 }
 
+// SetHealthy sets engine healthy state
+func (e *Engine) SetHealthy(state bool) {
+	e.Lock()
+	e.healthy = state
+	// if engine is healthy, clear failureCount
+	if state {
+		e.failureCount = 0
+	}
+	e.Unlock()
+}
+
 // Status returns the health status of the Engine: Healthy or Unhealthy
 func (e *Engine) Status() string {
 	if e.healthy {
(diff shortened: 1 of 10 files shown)


dongluochen:
ping @abronan @aluzzardi .

cluster/engine.go
```diff
@@ -434,7 +459,7 @@ func (e *Engine) refreshLoop() {
 			failedAttempts++
```
aluzzardi: I think `failureCount` and `failedAttempts` here should be the same things.

- The refresh loop will contribute in increasing the failed count (so after 2 failed refresh loops, we just need to make a single failing call to mark the node as unhealthy)
- It will be configurable by the same `e.opts.RefreshRetry` - maybe the name is wrong now?

cluster/engine.go
```diff
@@ -192,6 +196,27 @@ func (e *Engine) Status() string {
...
+	e.Lock()
+	e.failureCount++
+	if e.healthy && e.failureCount >= engineFailureCountThreshold {
+		e.healthy = false
+	}
+	e.Unlock()
+}
+
+// SetEngineHealth sets engine healthy state
+func (e *Engine) SetEngineHealth(state bool) {
```
aluzzardi: This should map with `IsHealthy` to look more like a getter/setter.

Perhaps `IsHealthy()` and `SetHealthy()`?

cluster/engine.go
```diff
@@ -192,6 +196,27 @@ func (e *Engine) Status() string {
 	return "Unhealthy"
 }
 
+// IncFailureCount increases engine's failure count, and set engine as unhealthy if threshold is crossed
+func (e *Engine) IncFailureCount() {
```
aluzzardi: I think this should be private (`incFailureCount`?).

No-one outside should be calling this - I think internally we should call it whenever one of the dockerclient calls fails.

You might do a helper, like:

```go
foo, err := e.client.ListContainers()
e.checkTransportError(err)
[... rest of the code as usual ...]

func (e *Engine) checkTransportErr(err error) {
    if (check if it's a transport error from docker client) {
        e.increaseFailureCount()
    } else {
        e.resetFailureCount()
   }
}
```

Or something like that.

We would call `checkTransportError` **every** time we use `dockerclient`, meaning that every call to the engine would contribute to the failure count.

Thoughts?

dongluochen: Totally agree with your comments on `checkTransportError`. Using feedback from dockerclient can fail nodes much faster and improve system efficiency.

dongluochen: `incFailureCount` cannot be made private because dockerclient are used in `package swarm`.

cluster/engine.go
dongluochen: You are right. I merge this logic and update `e.opts.RefreshRetry` to `e.opts.FailureRetry`.

cluster/engine.go
dongluochen: Updated. Thanks!

vieux:
We did release 1.0.1 with the `engine-refresh-retry` flag. Is the flag name change really necessary ?

dongluochen:
@vieux Thanks for the comment. I think we need to change it for long term good. With the fail fast PR, this flag will a key setting.

vieux:
@dongluochen then we need an upgrade machinisme, we cannot break ppl that will rely on `engine-refresh-retry` from 1.0.1

dongluochen:
ping @vieux @abronan @aluzzardi

dongluochen:
I talked with @aluzzardi about multiple clients in swarm. We think a refactor on client is necessary, tracked by #1549. It'd be done outside this PR.  

Ping @docker/swarm-maintainers.

jimmyxian:
LGTM

vieux:
LGTM

Issue #1357, opened 2015-10-30
Re-direct Swarm's log to a file instead of stdout


ghost:
One requirement of mine is to re-direct Swarm's log to a file instead of stdout; ">" is not the option as I'd like to run swarm in docker and retrieve the log by file. If this feature is here, I'd suggest to update it in `Dockerfile`

MorganBauer:
Just throwing this out there. Given the work put into the implementation in the docker-engine (multiple backends, etc), do you think it's feasible to turn that into a pkg and reuse it?

MHBauer:
Just throwing this out there. Given the work put into the implementation in the docker-engine (multiple backends, etc), do you think it's feasible to turn that into a pkg and reuse it?

klaus1982:
I think it's each component's work to log to a file; for example, zookeeper/marathon are using log4j which can be configured to append log message to a file.

But, I not sure this issue is to Swarm or log tool.

aluzzardi:
That should be easy to implement - we are using http://github.com/Sirupsen/logrus which supports many logging backends, including file.

klaus1982:
@aluzzardi , thanks for your feedback. That'll be great if we pass conf to logrus in Swarm :).

jimmyxian:
@klaus1982 Do you have anyhave  interest in implementing this feature. :)

klaus1982:
@jimmyxian , sure, my pleasure :). Any mentor/shepherd to guide me on this one?

jimmyxian:
@klaus1982 Cool.
* First: setup a swarm development environment(https://github.com/docker/swarm/blob/master/README.md)
* Second: contribute by this guide(http://docs.docker.com/opensource/code/)

Also, you can join our IRC #docker-swarm on freenode. Any questiones you can ping me there.

aluzzardi:
Some pointers:

`cli.go` is the main entry point for the CLI and this is where we handle global logging options (currently `--log-level`) and set up the logger:

https://github.com/docker/swarm/blob/master/cli/cli.go

What you probably want to do is:
- Add a new flag, such as `--log-file`
- When setting up the logger, open or create the log file and give the handle to `log.SetOutput` (https://godoc.org/github.com/Sirupsen/logrus#SetOutput)

aluzzardi:
Although as @MHBauer pointed out, Docker Engine recently added log drivers and perhaps we should use that instead.

/cc @docker/swarm-maintainers

jimmyxian:
[...]

PR #1536, opened 2015-12-12
Log Redirection

CiaranCostello:
The purpose of this pull request is to allow for log redirection in the docker swarm client. 

We have editted the flags.go and cli.go files in the cli folder to allow a file name to be selected that the log will be written to instead of standard out.

We did this as a fix for this suggested issue https://github.com/docker/swarm/issues/1357.

### DIFF
--- a/scheduler/filter/affinity.go
+++ b/scheduler/filter/affinity.go
@@ -34,7 +34,9 @@ func (f *AffinityFilter) Filter(config *cluster.ContainerConfig, nodes []*node.N
 			case "container":
 				containers := []string{}
 				for _, container := range node.Containers {
-					containers = append(containers, container.Id, strings.TrimPrefix(container.Names[0], "/"))
+					if len(container.Names) > 0 {
+						containers = append(containers, container.Id, strings.TrimPrefix(container.Names[0], "/"))
+					}
 				}
 				if affinity.Match(containers...) {
 					candidates = append(candidates, node)


abronan:
Hi @CiaranCostello, thanks for the contribution. Can you please use `gofmt` for the indentation issue and also squash your commits? You also need to sign the commit (see Gordon's suggestion above :). Thanks!

cli/cli.go
```diff
@@ -1,59 +1,60 @@
...
 			EnvVar: "DEBUG",
 		},
-
 		cli.StringFlag{
 			Name:  "log-level, l",
 			Value: "info",
 			Usage: fmt.Sprintf("Log level (options: debug, info, warn, error, fatal, panic)"),
 		},
+		cli.StringFlag{
+			Name:	"file",
```
abronan: Maybe `log-file` instead of just `file` as suggested in #1357

cli/cli.go
```diff
@@ -1,59 +1,60 @@
...
+		},
 	}
-
 	// logs
 	app.Before = func(c *cli.Context) error {
-		log.SetOutput(os.Stderr)
+        if c.IsSet("file") {
+            logDir := c.String("file")
+            fmt.Print(logDir)
+            file, err := os.OpenFile(logDir, os.O_RDWR, 0777)
```
jimmyxian: Should check `err` here

PR #1539, opened 2015-12-14
Log Redirection

CiaranCostello:
Editing cli/cli.go and cli/flags.go to allow logging to a file instead of to stdout.

Apologies for the new pull request. I accidentally closed the pull request when rebasing and amending.

Signed-off-by: CiaranCostello costelci@tcd.ie

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -2,13 +2,12 @@ package cli
 
 import (
 	"fmt"
-	"os"
-	"path"
-
 	log "github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
 	"github.com/docker/swarm/experimental"
 	"github.com/docker/swarm/version"
+	"os"
+	"path"
 )
 
 // Run the Swarm CLI.
@@ -27,7 +26,6 @@ func Run() {
 			Usage:  "debug mode",
 			EnvVar: "DEBUG",
 		},
-
 		cli.StringFlag{
 			Name:  "log-level, l",
 			Value: "info",
@@ -38,11 +36,24 @@ func Run() {
 			Name:  "experimental",
 			Usage: "enable experimental features",
 		},
+		cli.StringFlag{
+			Name:  "log-file",
+			Usage: "Logging the output to a file instead of stdout",
+		},
 	}
-
 	// logs
 	app.Before = func(c *cli.Context) error {
 		log.SetOutput(os.Stderr)
+		if c.IsSet("log-file") {
+			logDir := c.String("log-file")
+			file, err := os.OpenFile(logDir, os.O_RDWR, 0777)
+			if err != nil {
+				//log.Warn("Error creating log file: falling back to default method")
+				log.Fatalf(err.Error())
+			} else {
+				log.SetOutput(file)
+			}
+		}
 		level, err := log.ParseLevel(c.String("log-level"))
 		if err != nil {
 			log.Fatalf(err.Error())


CiaranCostello:
Here is the link to the previous pull request that I accidentally closed, https://github.com/docker/swarm/pull/1536

vieux:
LGTM, once this PR gets merged, @jimenez please to a PR to reflect this change in mesos.

cli/cli.go
```diff
@@ -26,17 +26,27 @@ func Run() {
...
+		},
 	}
-
 	// logs
 	app.Before = func(c *cli.Context) error {
-		log.SetOutput(os.Stderr)
+		if c.IsSet("log-file") {
+			logDir := c.String("log-file")
+			fmt.Print(logDir)
+			file, err := os.OpenFile(logDir, os.O_RDWR, 0777)
```
abronan: We should check for the error here opening the file, if it fails: try to create the file and the directory hierarchy. If there is a failure creating the file or directory, then fall-back to default log mode (with `os.Stderr`).

cli/flags.go
```diff
@@ -141,4 +141,8 @@ var (
 		Value: "30s",
 		Usage: "Leader lock release time on failure",
 	}
+	logFile = cli.StringFlag{
```
abronan: This portion is redundant, you already declared it in `cli/cli.go`.

cli/cli.go
```diff
@@ -26,17 +26,27 @@ func Run() {
...
+			fmt.Print(err)
```
aluzzardi: Instead of printing errors to the console, we should `log.SetOutput(os.Stderr)` before the `if c.IsSet("log-file") {` and use `log.Fatalf` when issuing `--log-file` related errors.

Also, no need to have a `else {` in that case

vieux:
@CiaranCostello can you please update your PR to address the comments above ?

cli/cli.go
CiaranCostello: Fatalf will exit the program, should this happen or should it fall back to os.Stderr like abronan suggested above?

abronan: @CiaranCostello Imho, falling back to the default way to do logging should be the behavior (but we should probably log it as a **Warning**: `Error creating log file: falling back to default method`).

It's not necessary to fail hard on this one as this will cause much confusion for users running the Swarm Image in a container and don't have a direct access to logs.

daniel-garcia:
How will swarm handle log rotation with file output?

CiaranCostello:
Log rotation is not provided with logrus.
image

vieux:
@CiaranCostello can you please rebase this PR ?

Thanks.

cli/cli.go
```diff
@@ -26,17 +25,30 @@ func Run() {
...
+			Usage: "Logging the output to a file instead of stdout",
+		},
 	}
-
 	// logs
 	app.Before = func(c *cli.Context) error {
 		log.SetOutput(os.Stderr)
+		if c.IsSet("log-file") {
+			logDir := c.String("log-file")
+			fmt.Print(logDir)
```
vieux: please remove this print.

CiaranCostello: Rebased and print statement removed

cli/cli.go
```diff
@@ -38,11 +36,24 @@ func Run() {
...
+			Usage: "Logging the output to a file instead of stdout",
+		},
 	}
-
 	// logs
 	app.Before = func(c *cli.Context) error {
 		log.SetOutput(os.Stderr)
+		if c.IsSet("log-file") {
+			logDir := c.String("log-file")
+			file, err := os.OpenFile(logDir, os.O_RDWR, 0777)
```
dongluochen: Is there a way to close file when swarm exits?

vieux: yes I think it would be nice to close it in `app.After`
Manage never exit anyway, but it would be nice to do that for other subcommands.

Also, (in another PR) we might catch signals and do a proper cleanup

CiaranCostello: The other option is a ```defer file.Close()```

vieux: does it works ? because it's going to close the file even before the app starts

vieux: @CiaranCostello ping for my comment

vieux: add you add a close in `app.After` ?

allencloud:
I think it is a great job to implement log redirection?
While swarm master glows a lot, can you rebase your code and solve the conflict here? @CiaranCostello

nishanttotla:
@CiaranCostello since this PR hasn't moved forward in a long time, we'll close it. If you still wish to work on it, feel free to reopen.

Issue #1550, opened 2015-12-16
go get github.com/docker/swarm is not working


jmccormick2001:
I get the following error when doing a build today of swarm:

tater jeffmc~/swarmproject]: go get github.com/docker/swarm
package github.com/docker/swarm
	imports github.com/Sirupsen/logrus
	imports github.com/codegangsta/cli
	imports github.com/docker/docker/pkg/ioutils
	imports golang.org/x/net/context
	imports github.com/docker/docker/pkg/parsers
	imports github.com/docker/docker/pkg/parsers/filters
	imports github.com/docker/docker/pkg/parsers/filters
	imports github.com/docker/docker/pkg/parsers/filters: cannot find package "github.com/docker/docker/pkg/parsers/filters" in any of:
	/usr/lib/golang/src/github.com/docker/docker/pkg/parsers/filters (from $GOROOT)
	/home/jeffmc/swarmproject/src/github.com/docker/docker/pkg/parsers/filters (from $GOPATH)

my environment is centos 7.2.1511, docker Version:      1.8.2-el7.centos

jmccormick2001:
I also get this error when I try to build with v1.0.1:

tater jeffmc~/swarmproject/src/github.com/docker/swarm]: git checkout v1.0.1
Note: checking out 'v1.0.1'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at 744e3a3... Merge pull request #1522 from dongluochen/bump-1.0.1
tater jeffmc~/swarmproject/src/github.com/docker/swarm]: godep go install
# github.com/docker/swarm/api
api/primary.go:108: undefined: "net/http/pprof".Trace
godep: go exit status 2

jmccormick2001:
it builds if I switch to the v1.0.0 branch

dongluochen:
@jmccormick2001 Could be a go version issue. See https://github.com/revel/revel/issues/937. Please check.

jmccormick2001:
I'm using the stock centos 7.2 version
of golang

go version go1.4.2 linux/amd64

dongluochen:
[...]

PR #1492, opened 2015-12-03
Enable profiling over HTTP in debug mode

aluzzardi:
Signed-off-by: Andrea Luzzardi <aluzzardi@gmail.com>

PR #1551, opened 2015-12-16
update readme with req for golang 1.5.2

MHBauer:
- 1.5.2 requirement from 4748bb14ddddb0343d2344e16e4fb82e2f918842

Signed-off-by: Morgan Bauer <mbauer@us.ibm.com>

I feel kind of bad about asking about trace in #1492. Forcing a requirement of 1.5.2 for that is probably a better argument against it. What's done is done. Que sera.

@vieux @aluzzardi @moxiegirl  to check please.

Also, to avoid things like #1550

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,11 +34,11 @@ using the Docker image.
 
 ### Prerequisites
 
-1. Beginning with Swarm 0.4 golang 1.4.x or later is required for building Swarm. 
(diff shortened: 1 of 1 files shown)


vieux:
@MHBauer why not udpate https://github.com/docker/swarm/pull/1190 ?

vieux:
@MHBauer and we require 1.5.x I think

MHBauer:
merging into #1190

PR #1552, opened 2015-12-16
Fix Consul Lock TTL with store failure

### LINKED ISSUE #1545: Leader Election Recovery with Consul fails sometimes
There are cases where the Swarm Leader Election process cannot recover from the store failure using Consul.

It does not affect *etcd* and *zookeeper* though.

As the leader election process relies on a Distributed Lock and Watch events, this might be related to #1413 (race and missed event on new leader key which would prevent the switch from replica to primary role).

Primary sees itself as the Leader with its own IP, it's just the switch from the Replica role back to Primary/leader that does not happen (related to the **candidate** goroutine).

abronan:
Fix #1545 

If using the Lock TTL feature with Consul, the code
path in libkv is issuing a Put in the background through
the PeriodicRenewal call. The error is then eaten up and
ignored on the candidate loop. This would lead to the
candidate and followers being stuck in their candidate
loop. Consequence would be that they would not retry to
take the lock ending in a state with no Leader.

This patch restores an explicit error check instead of
wrongfully passing on the error to the channel before
giving it back to the caller.

### DIFF
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -138,7 +138,7 @@ var (
 	}
 	flLeaderTTL = cli.StringFlag{
 		Name:  "replication-ttl",
-		Value: "30s",
+		Value: "15s",
 		Usage: "Leader lock release time on failure",
 	}
 )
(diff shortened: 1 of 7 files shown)


abronan:
Needs a test update.

abronan:
@vieux there seems to be an issue on the inspect test

dongluochen:
LGTM

vieux:
LGTM

dtepeneu:
Thank you Alex and the entire swarm team for such a quick fix!

Issue #1495, opened 2015-12-03
Performance issues with mesos - due to Global lock in Mesos scheduleTask() ?


ezrasilvera:
When using native Swarm cluster we can increase the create_container bandwidth (how many containers/sec the cluster can handle) by adding more engines.
However when using Mesos cluster no matter how many concurrent requests we use the bandwidth when using N engines is the same as using a single engine which mean such cluster can't scale at all !!

I believe the issue is coming from the global lock on scheduleTask(). For some reason it seems as if incoming tasks are waiting on this lock until previous task is completed/in RUNNING state ...

func (c *Cluster) scheduleTask(t *task) bool {
	c.scheduler.Lock()
	defer c.scheduler.Unlock()

To reproduce - just use a cluster of N engines and sent N create() or run() requests in parallel . The N requests will be completed as if there is only single engine.

ezrasilvera:
I think the problem is coming from  "finished, data, err := t.monitor()"  which waits till change in status sent by Mesos.

ezrasilvera:
I have changed the scheduler lock and this indeed solve the issue. Hopefully it's not causing other issues. 
You can see in the graph bellow that now multiple concurrent requests can be handled
![throughput_graph_mesos](https://cloud.githubusercontent.com/assets/13567561/11874203/2c8a7f2e-a4e7-11e5-9fcf-0edd30db401c.png)

abronan:
ping @vieux @jimenez

jimenez:
closed by https://github.com/docker/swarm/pull/1554

PR #1554, opened 2015-12-17
Change the scheduler lock in Mesos cluster

ezrasilvera:
Addressing the issue in  #1495 . Need to change the scheduler lock , otherwise it blocks all incoming tasks and doesn't allow concurrent usage

Signed-off-by: Ezra Silvera <ezra@il.ibm.com>

### DIFF
--- a/cluster/mesos/cluster.go
+++ b/cluster/mesos/cluster.go
@@ -444,16 +444,18 @@ func (c *Cluster) removeOffer(offer *mesosproto.Offer) bool {
 
 func (c *Cluster) scheduleTask(t *task) bool {
 	c.scheduler.Lock()
-	defer c.scheduler.Unlock()
+	//change to explicit lock defer c.scheduler.Unlock()
 
 	nodes, err := c.scheduler.SelectNodesForContainer(c.listNodes(), t.config)
 	if err != nil {
+		c.scheduler.Unlock()
 		return false
 	}
 	n := nodes[0]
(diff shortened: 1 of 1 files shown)


MHBauer:
I am not familiar with the code.

Looking at it, this is quite an aggressive advancement of the lock. Looking at the code I would think we could unlock right after `SelectNodesForContainer` on L#449. I don't see any other uses of the scheduler or the cluster after that look like they could be effected.

Alternatively, and I'd have to do more research, we could pull the lock inside `SelectNodesForContainer`.

ezrasilvera:
@MHBauer See the issue I opened for clarifications. The original lock was using defer to unlock blocking till the function completes. The problem is that the in that fuction there is a call to monitor() which was blocking till the container was in RUNNING state. this means that you must wait for the previous request to change to RUNNING before new request can be handled which as you can see in the graph I attached make 0 concurrency. 
I wanted to avoid side effects, as there are other functions that use the scheduler as well, so I pushed the unlock as far as I could which is just before the blocking call to monitor()

abronan:
ping @vieux @jimenez

tnachen:
This mostly looks good to me. We are refactoring this path (cc @jimenez ) as it currently only able to launch one task per offer but we want to be able to pack as much tasks per offer.

jimenez:
LGTM Thanks @ezrasilvera

PR #1556, opened 2015-12-18
Fix registry auth header pass-thru for build API

hoonmin:
Hi.

The request header for /build API should be X-Registry-Config instead of
X-Registry-Auth.

https://docs.docker.com/engine/reference/api/docker_remote_api_v1.21/

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1055,7 +1055,7 @@ func postBuild(c *context, w http.ResponseWriter, r *http.Request) {
 		json.Unmarshal([]byte(buildArgsJSON), &buildImage.BuildArgs)
 	}
 
-	if buf, err := decodeBase64UrlEncodedHeader(r, "X-Registry-Auth"); err == nil {
+	if buf, err := decodeBase64UrlEncodedHeader(r, "X-Registry-Config"); err == nil {
 		json.Unmarshal(buf, &buildImage.Config)
 	}
 
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -437,8 +437,7 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 
 	// Pass auth information along if present
 	var authConfig *dockerclient.AuthConfig
-	buf, err := base64.URLEncoding.DecodeString(r.Header.Get("X-Registry-Auth"))
-	if err == nil {
+	if buf, err := decodeBase64UrlEncodedHeader(r, "X-Registry-Auth"); err == nil {
 		authConfig = &dockerclient.AuthConfig{}
 		json.Unmarshal(buf, authConfig)
 	}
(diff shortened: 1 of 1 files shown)


api/handlers.go
```diff
@@ -1124,3 +1121,16 @@ func notImplementedHandler(c *context, w http.ResponseWriter, r *http.Request) {
 func optionsHandler(c *context, w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 }
+
+// Decode a base64-url-safe-encoded Registry Auth Config header.
+func decodeRegistryAuthHeader(r *http.Request, header string) []byte {
```
MHBauer: This could technically decode any header that's given to it, right? Let's give it a more general name and hope for reuse.

api/handlers.go
```diff
@@ -1124,3 +1121,16 @@ func notImplementedHandler(c *context, w http.ResponseWriter, r *http.Request) {
...
+	auth := r.Header.Get(header)
+	if auth == "" {
+		return nil
+	}
+	buf, err := base64.URLEncoding.DecodeString(auth)
+	if err != nil {
+		return nil
```
MHBauer: I do not like silently ignoring errors.

MHBauer:
This has a refactor mixed in with a change in behavior. I would rather see this done in two commits, to show the distinct change in header, separately from the change in code structure.

hoonmin:
@MHBauer Sure. I'll change it. Thank you for your comments.

hoonmin:
@MHBauer Fixed the commits. Would you please review this?

MHBauer:
My opinion is the code looks good.

I'm thinking now that we probably don't have a test for that particular functionality. If we give docker the wrong header this is obviously and catastrophically wrong. This seems like something we should have caught with an integration test of some sort.

vieux:
ping @aaronlehmann can you confirm it's the right headers ?

aaronlehmann:
The API doc seems to say `X-Registry-Config` is correct, but this would be an API change. I have no idea if anything relies on the existing behavior.

It seems strange for build to name the header `X-Registry-Config` and create to name the header `X-Registry-Auth`. It seems like a mistake.

Unless this is a regression, maybe it makes more sense to correct the API doc.

vieux:
@aaronlehmann `X-Registry-Auth` is basically one auth for one server,
when you do docker pull `localhost:5000/test` the cli parses the config file and send you just the auth for `localhost:5000`

For a build it's different since the CLI doesn't parse the Dockerfile, it doesn't know which server you are going to to talk, so it sends all the `auths` it has.
To me it makes sense that it uses a different header (`X-Registry-Config`)

aaronlehmann:
@vieux: This makes sense, thanks for explaining.

The API document doesn't make this clear.

For `X-Registry-Auth` it says: `Include a base64-encoded AuthConfig. object`
For `X-Registry-Config` it says: `A base64-url-safe-encoded Registry Auth Config JSON object with the following structure:`

It would be great if a separate PR could clarify this and explain that one refers to a specific server and the other does not.

amitshukla:
@hoonmin can you update us on the status of this PR?

vieux:
@amitshukla I replaced this PR by https://github.com/docker/swarm/pull/1806

vieux:
Fixed by https://github.com/docker/swarm/pull/2042

Issue #1540, opened 2015-12-14
docker run with image_id fails randomly


databus23:
I'm having trouble running docker containers by image id that I created on the swarm cluster with `docker build`.

Its working sometimes but not always.
When its not working I get errors like this:
```
docker build -f Dockerfile.build .
Sending build context to Docker daemon 9.403 MB
[...] (Build steps omitted)
Successfully built 9007a85d2b52
docker run --rm 9007a85d2b52 make test
Error response from daemon: Error: image library/9007a85d2b52:latest not found
```
I also tried using the full image instead of the abbreviated one but this also only works sometimes and when its not working I get the following error:
```
docker build -f Dockerfile.build .
Sending build context to Docker daemon 3.153 MB
[...] (Build steps omitted)
Successfully built 9cebfa102536
docker run --rm $(docker images -a -q --no-trunc |grep 9cebfa102536) make test
Error response from daemon: Invalid repository name (9cebfa1025365797ffb88e729912dd1c60d5025e962fd008c81e4c316e53110f), cannot specify 64-byte hexadecimal strings
```

So it looks like that swarm sometimes is not aware of the image id that was build in the cluster and interprets the image_id as an image name which it tries to lookup in the registry

aluzzardi:
@vieux @jimmyxian Could you look into this? Both of you worked in image mapping.

jimmyxian:
@databus23 Thanks for reporting this. 
I will try to reproduce and see what's wrong. :)

anandkumarpatel:
+1 when trying to run an image from a step from a docker build 
```
Magikarp@10.0.0.1:/test # docker build --no-cache .
Sending build context to Docker daemon 2.048 kB
Step 1 : FROM busybox
 ---> ac6a7980c6c2
Step 2 : RUN echo hi
 ---> Running in 986466089fcd
hi
 ---> 43515763c8d3
Removing intermediate container 986466089fcd
Step 3 : RUN echo again
 ---> Running in 5b34088030d6
again
 ---> 5bbf6a0ac54c
Removing intermediate container 5b34088030d6
Successfully built 5bbf6a0ac54c

Magikarp@10.0.0.1:/test $ sudo docker run -it 43515763c8d3 bash
``` 
^ above works

when trying the same thing pointing swarm:
```
Gyarados: docker run -it 43515763c8d3 bash
Error response from daemon: Error: image library/43515763c8d3:latest not found
```

and I verified that swarm knows about this host

[...]

PR #1565, opened 2015-12-23
fresh image when receive commit event

jimmyxian:
Fix: https://github.com/docker/swarm/issues/1540

### DIFF
--- a/test/integration/api/images.bats
+++ b/test/integration/api/images.bats
@@ -68,3 +68,13 @@ function teardown() {
 	[ "${#lines[@]}" -eq 2 ]
 	[[ "${lines[1]}" == *"testimage"* ]]
 }
+
+@test "docker images - after commit on engine side" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	docker -H ${HOSTS[0]} run -d --name test_container busybox sleep 500
+	docker -H ${HOSTS[0]} commit test_container testimage
+
+	retry 5 1 eval "docker_swarm images | grep 'testimage'"
+}
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -773,7 +773,7 @@ func (e *Engine) String() string {
 func (e *Engine) handler(ev *dockerclient.Event, _ chan error, args ...interface{}) {
 	// Something changed - refresh our internal state.
 	switch ev.Status {
-	case "pull", "untag", "delete":
+	case "pull", "untag", "delete", "commit":
 		// These events refer to images so there's no need to update
 		// containers.
 		e.RefreshImages()


test/integration/api/images.bats
```diff
@@ -68,3 +68,15 @@ function teardown() {
...
+
+@test "docker images - after build on engine side" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	docker -H ${HOSTS[0]} build --no-cache -t testimage:latest $TESTDATA/build
+
+	run docker_swarm images testimage
+	[ "$status" -eq 0 ]
+	[ "${#lines[@]}" -eq 2 ]
```
anandkumarpatel: question: the `2` here means that `2` images present here? one for `testimage:latest` and one `busybox`?

jimmyxian: It will only show testimage, such as(2 = header + image)
```
root@SZX1000041894:/home# docker images centos
REPOSITORY          TAG                 IMAGE ID            CREATED             VIRTUAL SIZE
centos              latest              14dab3d40372        2 weeks ago         194.7 MB
```

But this PR still have problem(CI is not OK), It will continue working on this.

vieux:
@jimmyxian engine-master still have issues, but engine-stable should be good, as long as this one is green, we can merge.

jimmyxian:
@vieux Yeah. I will check the integration test (a little busy recently)   :)

vieux:
@jimmyxian even is the refresh images is triggered right after a commit, you might need to wait for it to finish, before calling `docker images`

jimmyxian:
@vieux 
Image that we build can not be refreshed , even we can receive `commit` event when building image on engine side.
Receiving `commit` event do not mean "build finished"
So, this can not fix: https://github.com/docker/swarm/issues/1540
I have rewritten the test
`docker build` on engine side ---->  `docker commit` on engine side)
I think It's OK to merge this PR.  :)

anandkumarpatel:
@vieux or @jimmyxian I might be wrong but it seems like this would fix #1540 (or at least me issue I think)

I try to start an intermediate image in from a build. Once the `commit` event is emitted, and I run `docker images`, the image I want shows up in the list.

so if `e.RefreshImages()` does what I think it does (runs `docker images` after getting the `commit` event) this should fix the issue I was having :)

jimmyxian:
@anandkumarpatel 
Ehm. this PR can solve `start an intermediate image`.
Can you try this and see if it works. Thanks.

vieux:
LGTM ping @abronan 

@jimmyxian any plan to fix build ?

abronan:
LGTM

jimmyxian:
@vieux 
Yeah, try to find the way to solve. :)

PR #1568, opened 2015-12-29
docs: add note for the case of the daemon being started through ssh

### LINKED ISSUE #1343: Swarm node disconnects after SSH terminal closes
Hi,

  I set up a swarm cluster, when I run on the host\node:

  docker daemon -H tcp://0.0.0.0:2375 I get the daemon output to the console and everything works fine but when I close the terminal or the SSH connection disconnects I get:

  Error response from daemon: dial tcp <node IP>:2375: connection refused

  when runing commands on the cluster manager.

  It seems the deamon is not listing to the port anymore… running on the host netstat -an|grep :2375 shows nothing is listening to 2375 .

  Isn’t a daemon supposed to listen after terminal closes and SSH session disconnects ?

Thanks,
 Doron

abronan:
Add a simple note in the case the daemon is started through SSH for the manual install process.

Fixes #1343 

/cc @moxiegirl 

### DIFF
--- a/docs/install-manual.md
+++ b/docs/install-manual.md
@@ -36,6 +36,13 @@ Docker engine as registered. To workaround this problem, you can generate a new
 ID for each node with affected by this issue. To do this stop the daemon on a node,
 delete its `/etc/docker/key.json` file, and restart the daemon.
 
+If you use `ssh` to start the Engine `daemon`, make sure the daemon runs in the
+background. You can do this using `disown` or `nohup` in combination with the `&`
+(ampersand). Or you could launch the `daemon` through `systemd` or any other
+Linux-distribution-specific method. Regardless of which launch method you
+choose, make sure the daemon process does not stop when your session on a node
+ends.
+
 You can run Docker Swarm on Linux 64-bit architectures. You can also install and
 run it on 64-bit Windows and Max OSX but these architectures are *not* regularly
 tested for compatibility.


docs/install-manual.md
```diff
@@ -23,6 +23,7 @@ your network.  Each node in your swarm must:
 
 * be accessible by the swarm manager across your network
```
moxiegirl: @abronan I'd move this to another location.  Also, need to clean up some antecedent issues in the content. I think the `this` in your original paragraph meant the daemon. Check my rewrite please...

```markdown
* be accessible by the swarm manager across your network
* have Docker Engine 1.6.0+ installed
* open a TCP port to listen for the manager
* *do not* install on a VM or from an image created through cloning

Docker generates a unique ID for the Engine that is located in the
`/etc/docker/key.json` file. If a VM is cloned from an instance where a
Docker daemon was previously pre-installed, Swarm will be unable to differentiate
among the remote Docker engines. This is because the cloning process copied the
the identical ID to each image and the ID is no longer unique.

If you forget this restriction and create a node anyway, Swarm displays a single
Docker engine as registered. To workaround this problem, you can generate a new
ID for each node with affected by this issue. To do this stop the daemon on a node,
delete its `/etc/docker/key.json` file, and restart the daemon.

> **Note**: If you use `ssh` to start the Engine `daemon`, make sure the daemon runs in the background. You can do this using `disown` or `nohup` in combination with the `&` (ampersand). Or you could launch the `daemon` through `systemd` or any other Linux-distribution-specific method. Regardless of which launch method you choose, make sure the daemon process does not stop when your session on a node ends.
```

abronan:
@moxiegirl PTAL, sorry for the delay :) included your suggestion (thanks!).

moxiegirl:
LGTM

vieux:
LGTM

Issue #1508, opened 2015-12-07
Is there a way to temporarily disable scheduling to a node (eg. for maintainance) ?


ikreymer:
I am running a swarm cluster where the containers are distributed using the default spread scheduling.

However, at times I would like to be able to temporarily stop scheduling to a particular node, eg. for example, to gradually drain it of containers so that it can be taken down for maintenance.

Currently I just stop the `swarm-agent` container on the node, but this is less than ideal (cleanup tasks may not run), is there any other way to do this?

dongluochen:
@ikreymer You can use constraints to exclude the node under maintenance from scheduler, e.g., `docker -H swarm_ip:swarm_port run -e constraint:node!=#NODE_UNDER_MAINTENANCE# hello-world`. Built-in support may be available in the future for node exclusion.

ikreymer:
If I understand https://docs.docker.com/swarm/scheduler/filter/ correctly

> To tag a node with a specific set of key/value pairs, one must pass a list of --label options at docker startup time

It looks like labels are set statically on startup. The idea is to be able to dynamically set the node for maintenance while its running, eg. 'take it out of the pool' so this won't help. I am using swarm as a kind of load balancer and would like to disable a node (for example, to make updates on it, etc..) and then put it back in.

If labels could be added dynamically, that could work.

abronan:
@ikreymer Thanks for the feedback. I think this is the scope of #1486 which aims to improve Node Management. Draining a node that goes down for maintenance should fall into the scope of this proposal. So we'll make sure we address that scenario as part of the changes.

Using labels for this is impractical (and as you outlined, they cannot be added dynamically, you have to stop the Engine and restart it).

ikreymer:
Thanks for the quick response, looking at the proposal, I think this is also similar to #1341

aluzzardi:
In the meantime, there's a hack you can do.

You can get around the missing functionality of dynamically changing node labels by running containers with a specific set of labels on drained machines and applying affinities to never co-schedule containers next to them.

Example:

1.  Create a container named "node-123-drain" on the machine "node-123" and label it as "drained=true".

```
docker create --name node-123-drain --label drained=true -e constraint:node==node-123 busybox
```

Note that I'm using docker create rather than docker run. The container needs not to be running and take resources, it just needs to exist in the machine. I'm using "busybox" as it's a tiny image, but you could use anything else, preferably one that is already on the machine so there's no need to pull new content.

[...]

PR #1569, opened 2015-12-30
Improve node management.

dongluochen:
This change improves swarm node management according to proposal #1486. 
1. Introduce pending state. Pending nodes need validation before moving to healthy state. Resolve issues of duplicate ID and no join retry for unreachable node issues.
2. Expose error and last update time in docker info.
3. Use connect success/failure to drive state transition between healthy and unhealthy.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -208,15 +208,11 @@ func (e *Engine) IsHealthy() bool {
 	return e.state == stateHealthy
 }
 
-// setState sets engine healthy state
+// setState sets engine state
 func (e *Engine) setState(state engineState) {
 	e.Lock()
 	defer e.Unlock()
 	e.state = state
-	// if engine is healthy, clear failureCount
-	if state == stateHealthy {
-		e.failureCount = 0
-	}
 }
 
 // TimeToValidate returns true if a pending node is up for validation
@@ -292,10 +288,17 @@ func (e *Engine) UpdatedAt() time.Time {
 	return e.updatedAt
 }
 
+func (e *Engine) resetFailureCount() {
+	e.Lock()
+	defer e.Unlock()
+	e.failureCount = 0
+}
+
 // CheckConnectionErr checks error from client response and adjusts engine healthy indicators
 func (e *Engine) CheckConnectionErr(err error) {
 	if err == nil {
 		e.setErrMsg("")
+		e.resetFailureCount()
 		// If current state is unhealthy, change it to healthy
 		if e.state == stateUnhealthy {
 			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Infof("Engine came back to life after %d retries. Hooray!", e.failureCount)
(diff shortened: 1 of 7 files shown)


abronan:
Design LGTM. Will review the code and test shortly.

ping @vieux for design approval.

@dongluochen This might need a set of integration tests (or improve the existing ones) to validate that the state machine works and that a node transitions correctly from a state to another. At least for `Pending` to `Healthy` and from `Healthy` to `Unhealthy`.

dongluochen:
@abronan thanks Alex! I'll update test cases soon.

vieux:
@dongluochen small comment, but could we please put `└ Error: <none>` or `└ Error: (none)` instead of blanc ?

dongluochen:
@vieux I updated no error to `└ Error: (none) `. Thanks for the comment.

vieux:
LGTM

dongluochen:
ping @docker/swarm-maintainers.

cluster/engine.go
```diff
@@ -26,6 +26,27 @@ const (
...
+type engineState int
+
+const (
+	// pending means an engine added to cluster has not been validated
+	statePending engineState = iota
+	// unhealthy means an engine is unreachable
+	stateUnhealthy
+	// healthy means an engine is reachable
+	stateHealthy
+	// maintenance means an engine is under maintenance.
```
abronan: @dongluochen This is just for convenience but can we add a TODO and reference to the issue (#1508) here? Because the comment is there anyway :)

cluster/engine.go
```diff
@@ -182,45 +203,114 @@ func (e *Engine) isConnected() bool {
...
-	if state {
-		e.failureCount = 0
+	defer e.Unlock()
+	if e.state != statePending {
+		return false
 	}
-	e.Unlock()
+	sinceLastUpdate := time.Since(e.updatedAt)
+	// Increase check interval for a pending engine according to failureCount and cap it at 4 hours
+	if sinceLastUpdate > 4*time.Hour || sinceLastUpdate > time.Duration(e.failureCount)*30*time.Second {
```
abronan: small nit but I feel like `4*time.Hour` and `30*time.Second` should be constants here.

Suggestions:

`validationLimit -> 4*time.hour`
`retryFactor -> 30*time.Second`

cluster/swarm/cluster.go
```diff
@@ -288,6 +325,26 @@ func (c *Cluster) monitorDiscovery(ch <-chan discovery.Entries, errCh <-chan err
 	}
 }
 
+// monitorPendingEngines checks if some previous unreachable/invalid engines have been fixed
+func (c *Cluster) monitorPendingEngines() {
+	for {
+		// Don't need to do it frequently
+		time.Sleep(10 * time.Second)
```
abronan: probably here as well.

abronan:
Tested and seems to work as expected, just very small nits but LGTM

dongluochen:
Thanks @abronan! I've updated code accordingly.

abronan:
Sweet! Thanks @dongluochen!

abronan:
:tada:

Issue #1488, opened 2015-12-03
Proposal: Container Rescheduling


aluzzardi:
## Background

The goal of this proposal is to reschedule containers automatically in case of node failure.

This is currently one of the top requested feature for Swarm.
## Configuration

The behavior should be user controllable and **disabled** by default since rescheduling can have nasty effects on stateful containers.

The user can select the policy at `run` time use the `reschedule` environment variable:

```
docker run -e reschedule:on-node-failure redis
```

Possible values for `reschedule` are:
- **no**: Never reschedule the container (**default**).
- **on-node-failure**: Reschedule the container whenever the node fails.

The reason this is more complicated than `yes`/`no` is in the future we might have more complicated rescheduling policies (for instance, we might want reschedule containers to re-spread or re-pack them). **Open question: Is this really necessary?**

Rescheduling policies will be stored as a container label: `com.docker.swarm.reschedule-policy`
## Persistance

Ideally, Swarm would store all containers (at least those that should be rescheduled) persistently. That way, the manager can figure out which containers are down and take action.

Unfortunately, we currently don't have a shared state and this feature has been postponed because of that for a long time.

Since this is one of the top requested feature, I propose we take a different approach until we have shared state (that feature has been postponed for usability concerns - we don't want to make a kv store a dependency for Swarm).

By storing the _rescheduling policy_ as a container label, we are able to reconstruct the desired state at startup time.

Since we are already storing **constraints**,  **affinities** etc as container labels (exactly for this reason), the manager will have all the information it needs to perform rescheduling.

This means we can restart the manager as much as we want and it will resume rescheduling as expected.

However, the problem arises when a node goes **down** while the manager is not running: in that case, we won't "remember" that container even existed when the manager is started again.

This situation can be counter-balanced by using `replication`. The `rescheduler` would be running on the **primary** manager and, upon failure, the **replica** that gets elected **primary** would be taking over rescheduling responsibilities.

Since every manager is aware of the cluster state (containers & rescheduling policy), it means that as long as at least one manager is still running we won't forget about containers.
## Failure detection

This functionality is already provided by `cluster/engine`.

`Engine` actively heartbeats nodes in the cluster every `X` seconds. After `Y` failures, the node is marked as _unhealthy_.

Rescheduling can rely on the health status already available.
## Resurrection

Eventually, a node may come back to life and re-join the cluster. If the node has containers that were rescheduled, we will end up with duplicates.

[...]

PR #1578, opened 2016-01-04
[experimental] Simple container rescheduling on node failure

vieux:
See #1488 for details

When a node goes down, swarm will try to reschedule container on another machine.

Depends on https://github.com/docker/docker/pull/19001 for IP stability.
Depends on https://github.com/docker/swarm/pull/1569 for proper cleanup of old containers.

### DIFF
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -186,3 +186,12 @@ func (c *ContainerConfig) HaveNodeConstraint() bool {
 	}
 	return false
 }
+
+// Affinities returns all the affinities from the ContainerConfig
+func (c *ContainerConfig) ReschedulePolicy() string {
+	policy, ok := c.Labels[SwarmLabelNamespace+".reschedule-policy"]
+	if !ok {
+		return "no"
+	}
+	return policy
+}
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -2,6 +2,8 @@ package cluster
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"strings"
 
 	"github.com/samalba/dockerclient"
@@ -63,9 +65,10 @@ func consolidateResourceFields(c *dockerclient.ContainerConfig) {
 // BuildContainerConfig creates a cluster.ContainerConfig from a dockerclient.ContainerConfig
 func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
 	var (
-		affinities  []string
-		constraints []string
-		env         []string
+		affinities         []string
+		constraints        []string
+		reschedulePolicies []string
+		env                []string
 	)
 
 	// only for tests
@@ -83,12 +86,19 @@ func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
 		json.Unmarshal([]byte(labels), &constraints)
 	}
 
-	// parse affinities/constraints from env (ex. docker run -e affinity:container==redis -e affinity:image==nginx -e constraint:region==us-east -e constraint:storage==ssd)
+	// parse reschedule policy from labels (ex. docker run --label 'com.docker.swarm.reschedule-policies=on-node-failure')
+	if labels, ok := c.Labels[SwarmLabelNamespace+".reschedule-policies"]; ok {
+		json.Unmarshal([]byte(labels), &reschedulePolicies)
+	}
+
+	// parse affinities/constraints/reschedule policies from env (ex. docker run -e affinity:container==redis -e affinity:image==nginx -e constraint:region==us-east -e constraint:storage==ssd -e reschedule:off)
 	for _, e := range c.Env {
 		if ok, key, value := parseEnv(e); ok && key == "affinity" {
 			affinities = append(affinities, value)
 		} else if ok && key == "constraint" {
 			constraints = append(constraints, value)
+		} else if ok && key == "reschedule" {
+			reschedulePolicies = append(reschedulePolicies, value)
 		} else {
 			env = append(env, e)
 		}
@@ -108,6 +118,13 @@ func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
 	if len(constraints) > 0 {
 		if labels, err := json.Marshal(constraints); err == nil {
 			c.Labels[SwarmLabelNamespace+".constraints"] = string(labels)
+		}
+	}
+
+	// store reschedule policies in labels
+	if len(reschedulePolicies) > 0 {
+		if labels, err := json.Marshal(reschedulePolicies); err == nil {
+			c.Labels[SwarmLabelNamespace+".reschedule-policies"] = string(labels)
 		}
 	}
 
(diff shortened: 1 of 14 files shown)


cluster/config.go
```diff
@@ -83,12 +84,24 @@ func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
 		json.Unmarshal([]byte(labels), &constraints)
 	}
 
+	// parse reschedule policy from labels (ex. docker run --label 'com.docker.swarm.reschedule-policy=on-node-failure')
+	if label, ok := c.Labels[SwarmLabelNamespace+".reschedule-policy"]; ok {
+		reschedulePolicy = label
+	}
+
 	// parse affinities/constraints from env (ex. docker run -e affinity:container==redis -e affinity:image==nginx -e constraint:region==us-east -e constraint:storage==ssd)
```
dongluochen: nit: update comment.

docs/scheduler/rescheduling.md
```diff
@@ -0,0 +1,46 @@
...
+[menu.main]
+parent="smn_workw_swarm"
+weight=5
++++
+<![end-metadata]-->
+
+# Rescheduling
+
+The Docker Swarm scheduler is able to detect node failure and
+restart it's container on another node.
```
dongluochen: `it's container` => `its containers`

docs/scheduler/rescheduling.md
```diff
@@ -0,0 +1,46 @@
...
+
+When you start a container, use the env var `reschedule` or the
+label `com.docker.swarm.reschedule-policy` to specify the policy to
+apply to the container.
+
+```
+# do not reschedule (default)
+$ docker run -d -e reschedule:off redis
+# or
+$ docker run -d -l com.docker.swarm.reschedule-policy=off redis
```
dongluochen: Why do we provide 2 different options for the same function?

test/integration/rescheduling.bats
```diff
@@ -0,0 +1,240 @@
...
+	# Get c1 swarm id
+	swarm_id=$(docker_swarm inspect -f '{{ index .Config.Labels "com.docker.swarm.id" }}' c1)
+
+	# Stop node-0
+	docker_host stop ${DOCKER_CONTAINERS[0]}
+
+	# Wait for Swarm to detect the node failure.
+	retry 5 1 eval "docker_swarm info | grep -q 'Unhealthy'"
+
+	# c1 should have been rescheduled from node-0 to node-1
```
dongluochen: It could fail if rescheduling is slow.

test/integration/rescheduling.bats
```diff
@@ -0,0 +1,240 @@
...
+	run docker_swarm inspect c3
+	[ "$status" -eq 0 ]
+	[[ "${output}" == *'"Name": "node-1"'* ]]
+}
+
+@test "reschedule conflict" {
+	start_docker_with_busybox 2
+	swarm_manage
+
+	docker_swarm run --name c1 -dit -e reschedule:on-node-failure --label com.docker.swarm.reschedule-policy=off busybox sh
```
dongluochen: It'd be better to fail this command explicitly.

cluster/watchdog.go
```diff
@@ -0,0 +1,101 @@
...
+	for _, container := range e.Containers() {
+		if w.cluster.Container(container.Config.SwarmID()) != nil {
+			// container already exists in the cluster, destroy it
+			e.RemoveContainer(container, true, true)
+		}
+	}
+}
+
+// Try to reschedule containers if possible
+func (w *Watchdog) reschedulePendingContainers() {
```
dongluochen: There is a type `pendingContainer` in cluster/swarm/cluster.go. I think it better to differentiate the 2 concepts.

cluster/config.go
```diff
@@ -83,12 +84,24 @@ func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
...
 	for _, e := range c.Env {
 		if ok, key, value := parseEnv(e); ok && key == "affinity" {
 			affinities = append(affinities, value)
 		} else if ok && key == "constraint" {
 			constraints = append(constraints, value)
+		} else if ok && key == "reschedule" {
+			if reschedulePolicy == "" {
+				reschedulePolicy = value
+			} else if reschedulePolicy != value {
```
dongluochen: I prefer failing explicitly.

cluster/config.go
```diff
@@ -111,6 +124,11 @@ func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
 		}
 	}
 
+	// store reschedule policy in labels
+	if reschedulePolicy != "" {
+		c.Labels[SwarmLabelNamespace+".reschedule-policy"] = reschedulePolicy
```
dongluochen: If there is only a very limited number of reschedulePolicy, I think we should fail it explicitly. For example, user may input `-e reschedule:On-node-failure` with a capital `O`, or `-e reschedule:node-failure`. It's better to let user know it's wrong, than failing it later implicitly.

docs/scheduler/rescheduling.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+```
+# rechedule on node failure
```
claudiu-coman: Hope you don't mind me bugging into your work, just trying to help. Typo: "reschedule"

docs/scheduler/rescheduling.md
vieux: It's the same with affinities and contraints:

1) in swarm < 0.3 we used env because we had only this.
2) in swarm > 0.4 we started storing them in labels because it's cleaner.
but we kept the `env` as compatibility.
3) after a few release, many people still use `env` because it's much more simpler to use.

Since it's stored in labels, there is no way to remove `-l com.docker.swarm.reschedule-policy=*`
we could remove `-e reschedule:*` but that's much more easier to use.
I'd like to keep both.

@abronan ?

docs/scheduler/rescheduling.md
vieux: @claudiu-coman not at all, thanks!

test/integration/rescheduling.bats
vieux: I don't get it

dongluochen: When a node becomes 'Unhealthy', its containers will be rescheduled but it may take time. If you inspect container right away, rescheduling could have finished or not.

docs/scheduler/rescheduling.md
```diff
@@ -0,0 +1,46 @@
...
+# Rescheduling
+
+The Docker Swarm scheduler is able to detect node failure and
+restart it's container on another node.
+
+## Rescheduling policies
+
+The rescheduling policies are:
+
+* `on-node-faillure`
```
abronan: `faillure` -> `failure`

docs/scheduler/rescheduling.md
abronan: I'm good with both too, I think it goes along with what we've always been doing with affinities and constraints.

cluster/config.go
vieux: @dongluochen agreed, do you have a suggestion for the name ?

cluster/config.go
vieux: @dongluochen agreed, do you have a suggestion for the name ?

cluster/watchdog.go
vieux: @dongluochen agreed, do you have a suggestion for the name ?

cluster/watchdog.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Handle cluster callbacks
+func (w *Watchdog) Handle(e *Event) error {
+	switch e.Status {
+	case "engine_reconnect":
+		go w.removeDuplicateContainers(e.Engine)
+	case "engine_disconnect":
+		go w.rescheduleContainers(e.Engine)
+	case "die", "destroy", "kill", "oom", "start", "stop", "rename":
+		go w.reschedulePendingContainers()
```
dongluochen: Any reason this trigger whole cluster reschedule, instead of just this engine? I think there could be a lot of events for a big cluster.

cluster/watchdog.go
dongluochen: See my other comment.

mavenugo:
@vieux we are working on the forced cleanup fix in https://github.com/docker/libnetwork/pull/862.
I have a private branch in docker/docker, where this is integrated fully with docker engine API which swarm can use to reschedule container reclaiming the locked resources. Would you like to give it a try in parallel as we make progress in getting this PRs merged in multiple repos ?

experimental/README.md~
```diff
@@ -0,0 +1,81 @@
+# Docker Experimental Features
```
moxiegirl: @vieux This looks like it shouldn't be here it is a backup file

moxiegirl:
@vieux One of the files appears to be a backup file, it has a ~ (tilde).  Otherwise, these LGTM

vieux:
@moxiegirl thanks fixed.

moxiegirl:
LGTM

dongluochen:
LGTM.

abronan:
LGTM

PR #1587, opened 2016-01-06
 do not save image affinity on reschedule

vieux:
it was the original intent, but wasn't working correctly /cc @jimmyxian 
Thanks @dongluochen for catching it.

using `configTemp := *config` and using `configTemp` doesn't do the trick since the embed `Label` field in a map (pointer)
So I decided to go toward the `RemoveAffinity` approach.

### DIFF
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -159,6 +159,22 @@ func (c *ContainerConfig) AddAffinity(affinity string) error {
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


dongluochen:
LGTM.

jimmyxian:
Nice catch, Thanks @dongluochen @vieux 
LGTM

PR #1595, opened 2016-01-08
docs: Typo in scheduler/filter.md

MiLk:
Small typo I found while reading the documentation.

### DIFF
--- a/docs/scheduler/filter.md
+++ b/docs/scheduler/filter.md
@@ -291,7 +291,7 @@ CONTAINER ID        IMAGE               COMMAND             CREATED             
 87c4376856a8        nginx:latest        "nginx"             Less than a second ago   running             192.168.0.42:80->80/tcp         node-1      trusting_yonath
 ```
 
-Then, use `-e affnity:com.example.type==frontend` to schedule a container next
+Then, use `-e affinity:com.example.type==frontend` to schedule a container next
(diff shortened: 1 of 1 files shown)


abronan:
Thank you @MiLk

/cc @thaJeztah Just to bring this to docs maintainer attention, seems like there is an issue with docs CI env.

`FATAL: 2016/01/08 mkdir /docs/public/opensource/kitematic/images/: no such file or directory`

Is it necessary to fail the docs tests if related to another project? I'm unfamiliar with the docs CI so just wondering :smile:

thaJeztah:
@abronan yeah, there's something funky causing Hugo to sometimes fail, see https://github.com/spf13/hugo/issues/1754

Reason if checks the other docs, is that the test needs all docs, to (e.g.) checks for duplicate menu-id's  (which can result in docs not being visible, or Hugo not generating the docs).

If you have an account for Jenkins, you can manually restart the test (I don't have an account currently, but I think Jess can provide one).

In this case, I think it's okay to ignore the failing docs-build, because no links were added, or pages renamed/menu-items added so LGTM

feel free to merge if you're okay with that! :+1:

abronan:
@thaJeztah thanks for the clarification!

MHBauer:
Good catch. Thanks @MiLk!

Issue #1593, opened 2016-01-07
Include docker plugin mechanism in Swarm to enable AuthZ plugin for AuthZ in Swarm 


doronp:
ACLs in Swarm see #1199 (ReOpen for #1365) #1366 - Seems to be better implemented using the Second option.

In order to achieve that Swarm should be able to load AuthZ plugin as in Docker engine:
https://github.com/docker/docker/pull/15365

Once done Swarm will load AuthZ plugin as the engine does.

Since the plugin mechanism is standalone I thought it should be a separate issue/pr

Thanks,
Doron Podoleanu

PR #1599, opened 2016-01-10
Add Plugin code from docker 

doronp:
Fulfill  #1593
Signed-off-by: Doron Podoleanu <doronp@il.ibm.com>


doronp:
go test -v -race ./... is successful locally, Can you please advice why it is failing ?
Thanks

abronan:
@doronp sorry for the delay, It seems that there was something wrong with a rebase, some of your commits are interleaved with other commits that were included in the meantime :disappointed: 

For the CI, it seems that this is a dependency error and it cannot find a package, but I think that fixing that rebase issue should do the trick.

doronp:
@abronan See:
https://github.com/docker/swarm/pull/1646
(Fresh clone and PR gets the same result after fixing the first two missing libs)

doronp:
Continue on #1646

PR #1600, opened 2016-01-11
Scheduler prefers nodes without connection failures

dongluochen:
Current swarm scheduler may keep trying the same node until its state changes to 'Unhealthy'. This change makes scheduler aware of connection failures and prefer nodes without failures. 

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -209,6 +209,18 @@ func (e *Engine) IsHealthy() bool {
 	return e.state == stateHealthy
 }
 
+// HealthIndicator returns degree of healthiness between 0 and 100.
+// 0 means node is not healthy (unhealthy, pending), 100 means last connectivity was successful
+// other values indicate recent failures but haven't moved engine out of healthy state
+func (e *Engine) HealthIndicator() int {
+	e.RLock()
+	e.RUnlock()
+	if e.state != stateHealthy || e.failureCount >= e.opts.FailureRetry {
+		return 0
+	}
+	return 100 - e.failureCount*100/e.opts.FailureRetry
+}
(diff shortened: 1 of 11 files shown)


cluster/engine.go
```diff
@@ -209,6 +209,18 @@ func (e *Engine) IsHealthy() bool {
 	return e.state == stateHealthy
 }
 
+// HealthIndicator returns degree of healthiness between 0 and 100.
+// 0 means node is not healthy (unhealthy, pending), 100 means last connectivity was successful
+// other values indicate recent failures but haven't moved engine out of healthy state
+func (e *Engine) HealthIndicator() int {
```
vieux: I'm curious, why not int64 here ?

cluster/engine.go
```diff
@@ -209,6 +209,18 @@ func (e *Engine) IsHealthy() bool {
...
+	e.RLock()
+	e.RUnlock()
```
aaronlehmann: Is this statement meant to be deferred?

cluster/engine.go
dongluochen: You are right. It's better to do the conversion here.

cluster/engine.go
dongluochen: Good catch. Thanks!

vieux:
LGTM

ping @abronan

abronan:
@dongluochen This may not be practical to do it (and we might hit timing issues on the CI), but that would be cool if we had the integration tests testing 3 or more engines with different health scores to test the impact of the health score on the scheduling more thoroughly.

LGTM though this should not block the merge.

PR #1624, opened 2016-01-12
add experimental- prefix on env/label

vieux:


vieux:
ping @abronan @dongluochen let's be explicit for now

PR #1626, opened 2016-01-12
add -experimental to enable rescheduling

vieux:

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -7,6 +7,7 @@ import (
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/codegangsta/cli"
+	"github.com/docker/swarm/experimental"
 	"github.com/docker/swarm/version"
 )
 
@@ -32,6 +33,11 @@ func Run() {
 			Value: "info",
 			Usage: fmt.Sprintf("Log level (options: debug, info, warn, error, fatal, panic)"),
 		},
+
+		cli.BoolFlag{
+			Name:  "experimental",
+			Usage: "enable experimental features",
+		},
 	}
 
 	// logs
@@ -48,6 +54,8 @@ func Run() {
 		if !c.IsSet("log-level") && !c.IsSet("l") && c.Bool("debug") {
 			log.SetLevel(log.DebugLevel)
 		}
+
+		experimental.ENABLED = c.Bool("experimental")
 
 		return nil
 	}
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/docker/swarm/experimental"
 	"github.com/samalba/dockerclient"
 )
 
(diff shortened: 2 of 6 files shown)


vieux:
ping @docker/swarm-maintainers what do you think of this ?

cluster/config.go
```diff
@@ -86,25 +87,27 @@ func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
...
-			affinities = append(affinities, value)
-		} else if ok && key == "constraint" {
-			constraints = append(constraints, value)
-		} else if ok && key == "reschedule" {
-			reschedulePolicies = append(reschedulePolicies, value)
-		} else {
-			env = append(env, e)
+		// parse affinities/constraints/reschedule policies from env (ex. docker run -e affinity:container==redis -e affinity:image==nginx -e constraint:region==us-east -e constraint:storage==ssd -e reschedule:off)
+		for _, e := range c.Env {
+			if ok, key, value := parseEnv(e); ok && key == "affinity" {
```
dongluochen: affinity and constraint are not experimental.

abronan:
@docker/swarm-maintainers So the context is that I expressed my concerns for #1624 because I though that adding `experimental` on the restart flag was an issue when we'll change the flag back when it's out of experimental (for example users scripting their deployments, etc. this would be a breaking change).

With this change it's actually a flag on the Manager which I think is better because not breaking existing scripts and deployments.

cli/cli.go
```diff
@@ -32,6 +33,11 @@ func Run() {
 			Value: "info",
 			Usage: fmt.Sprintf("Log level (options: debug, info, warn, error, fatal, panic)"),
 		},
+
+		cli.BoolFlag{
+			Name:  "experimental",
```
dongluochen: Why not use a flag like `autoreschdule`? You can state experimental in flag description. `experimental` is not one feature. Over time this flag would be used for other features. It's not backward compatible.

cluster/config.go
vieux: right

cli/cli.go
vieux: Yes that was my point,
docker engine for example has one regular build and one experimental build. not one build per experimental feature.

Going forward, ppl who wants to test the bleeding edge stuff just enable "-experimental" and that's it, they don't have to remove/add flags for every release.

vieux:
I'll add `Experimental: true` in the Info API response once this one gets merged (if it does) because it required yet another godep update.

vieux:
@abronan actually I put the flag on swarm itself, not the manager, so if one day we have experimental features on join on list, it will make things easier :)

abronan:
@vieux even better imho :+1:

dongluochen:
@vieux There is conflict please rebase.

vieux:
@dongluochen done. sometime I'm disappointed by git, this was was a no brainer, it could have resolved it.

dongluochen:
LGTM.

aluzzardi:
LGTM

moxiegirl:
docs LGTM

PR #1627, opened 2016-01-13
add buildtime, kernelversion and experimental to API version

vieux:
follow up of #1626

It makes me very happy as we introduce the `engine-api` dependency,
we can slowly start using it instead of `samalba/dockerclient` (at least for all the types)

/cc @calavera @icecrime 

(I used it because, as usual, `samalba/dockerclient `has it's types outdated (no Experimental, BuildTime or KernelVersion in the Version struct))


vieux:
ping @abronan @dongluochen, not mandatory for 1.10, but would be nice

abronan:
LGTM :tada: :balloon:

dongluochen:
LGTM. Thanks @vieux.

icecrime:
:tada:

PR #1629, opened 2016-01-13
Added role info when register Swarm to Mesos.

klaus1982:
Signed-off-by: Klaus Ma <klaus1982.cn@gmail.com>

### DIFF
--- a/cluster/mesos/cluster.go
+++ b/cluster/mesos/cluster.go
@@ -39,6 +39,7 @@ type Cluster struct {
 	taskCreationTimeout time.Duration
 	pendingTasks        *queue.Queue
 	engineOpts          *cluster.EngineOpts
+	role                string
 }
 
 const (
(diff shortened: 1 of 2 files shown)


cluster/mesos/cluster.go
```diff
@@ -82,13 +83,19 @@ func NewCluster(scheduler *scheduler.Scheduler, TLSConfig *tls.Config, master st
 	// Empty string is accepted by the scheduler.
 	user, _ := options.String("mesos.user", "SWARM_MESOS_USER")
 
+	var found bool
+
+	if cluster.role, found = options.String("mesos.role", "SWARM_MESOS_ROLE"); !found {
```
jimenez: What do you think if we put our `var found bool` inlined with the `if` as follow: 
```
if role, found := options.String("mesos.role", "SWARM_MESOS_ROLE"); found {
 cluster.role = role
}
``` 

And make the default `"*"` part of the initialization of Cluster line 78.

jimenez:
Thanks @klaus1982 ! Could you also add an integration test for this?

klaus1982:
Sure, I'll add some integration test for it.

jimenez:
Hi @klaus1982 , it'd be great if you could rebase this :)

vieux:
@klaus1982 please ping use to reopen once you rebased.

PR #1633, opened 2016-01-13
New network functions for mesos cluster: CreateNetwork(), RemoveNetwork(), and Networks()

ezrasilvera:
Implement the missing network functions for mesos cluster.  
These functions allow us to support the new docker bridge and overlay networks by 
1) Create a network 2) create a container with net=none 3) connect the container to the network created in (1)

We are also working on adding the new network capabilities as part of createContainer (support --net=userNetworkName but this require some changes to the mesos executer (that we are starting to push as well)

### DIFF
--- a/cluster/mesos/cluster.go
+++ b/cluster/mesos/cluster.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"os"
 	"sort"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


vieux:
ping @jimenez

cluster/mesos/cluster.go
```diff
@@ -242,7 +243,47 @@ func (c *Cluster) RemoveImages(name string, force bool) ([]*dockerclient.ImageDe
...
+	}
+
+	c.scheduler.Lock()
+	nodes, err := c.scheduler.SelectNodesForContainer(c.listNodes(), config)
+	c.scheduler.Unlock()
+	if err != nil {
+		return nil, err
+	}
+	if nodes == nil {
+		return nil, nil
```
dongluochen: I think it better to return error here.

ezrasilvera:
@dongluochen you are probably right but I tried to stay with the exact same errors that are returned in the case of the native swarm (without mesos) cluster. 
Maybe it would be better tat after this is merged I can have another PR to change the returned value on both the swarm and mesos cases?

ezrasilvera:
@dongluochen @jimenez I have fixed the error messages.

dongluochen:
LGTM. Thanks @ezrasilvera.

ezrasilvera:
@dongluochen @jimenez - any idea when it's going to be merged   ?

jimenez:
@ezrasilvera Thanks for adding this! this LGTM but before merging we need to have some integration tests. They should be very similar to those for Swarm stand alone. (It's impossible to test multihost network in the Swarm CI, but at least you can try creation and removal of a bridge network)

ezrasilvera:
@jimenez Thanks. I have added an integration test for all the network functions for mesos cluster. All passed..
ok 142 mesos - docker network ls
ok 143 mesos - docker network inspect
ok 144 mesos - docker network create
ok 145 mesos - docker network rm
ok 146 mesos - docker network disconnect connect

jimenez:
LGTM, Thanks @ezrasilvera !

PR #1642, opened 2016-01-15
Tcp user timeout

dongluochen:
Swarm implements TCP user timeout for dead peer detection. A sender can break the connection within 20 seconds, comparing with Linux default 15 minutes. We don't want to introduce cgo in swarm so it makes compromise on golang experiment package.

### DIFF
--- a/cluster/utils_windows.go
+++ b/cluster/utils_windows.go
@@ -1,0 +1,16 @@
+// +build !linux,!darwin
+
+package cluster
+
+import (
+	"errors"
+	"net"
+	"time"
+)
+
+// setTCPUserTimeout doesn't work under Windows because Go doesn't support
+// the option and swarm doesn't support cgo
+// This is a usability enhancement. Service shouldn't fail on this error.
+func setTCPUserTimeout(conn *net.TCPConn, uto time.Duration) error {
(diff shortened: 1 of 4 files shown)


cluster/utils_windows.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"errors"
+	"net"
+	"time"
+)
+
+// SetTCPUserTimeout doesn't support Windows because Go doesn't support
+// the option and swarm doesn't support cgo
+func setTCPUserTimeout(conn *net.TCPConn, uto time.Duration) error {
+	return errors.New("Go doesn't have native support for TCP_USER_TIMEOUT for this platform")
```
abronan: I'm concerned about this, does this mean that a Manager can't be started on a Windows Host because it does not support `TCP_USER_TIMEOUT`?

dongluochen: Thanks for bringing it up. TCP_USER_TIMEOUT is a good to have feature. Dockerclient doesn't fail if this setting is not supported by a system. I  will add a comment for this.

dongluochen:
Rebase and update use case comment.

abronan:
LGTM

ping @vieux

vieux:
LGTM !

vieux:
thanks @dongluochen

PR #1644, opened 2016-01-15
Allow specifying default resource constraints when running swarm manager

clinta:
Allow specifying default resource constraints with cluster options. This is very useful when using other tools interacting with swarm cannot create containers because they do not expose options for --cpu-shares or --memory.


nishanttotla:

PR #1646, opened 2016-01-17
add plugins PLEASE DO NOT REVIEW BEFORE looking at the other version at #1366  

doronp:
This also address (by importing packages from docker)#1593


doronp:
@abronan 

I have the next commit (AuthZ plugin infra based on this generic plugin infra) ready and passing race -v test locally and here:

https://github.com/doronp/swarm/tree/authZPlugin

That commit should conclude AuthZ plugin support for Swarm (will be like in docker)

I think We should first resolve the test issue here before I push it.

doronp:
ping @abronan

doronp:
version with importing the generic plugin and duplicating the authz plugin available at #1366

abronan:

doronp:
Will continue with #1366

Issue #1648, opened 2016-01-18
swarm in spread problem


yutian1224:
docker version: 1.9.1
swarm version: 1.01
nodes1: 25c 64g
nodes2: 25c 64g
nodes3: 42c 128g
nodes4: 42c 128g

I use swarm in spread strategy, but the memory of nodes are different.
The number of containers in node isn't grow up accoring to the specifications.

![image](https://cloud.githubusercontent.com/assets/13673671/12384503/a6081e0a-bded-11e5-912a-2fbe8a65e9c1.png)

alu:
If you need the spread distribute by resource, you have to specify memory limit (by `--memory`) or cpu limit (by `--cpuset-cpus` or `--cpu-shares`). Because swarm can get resource information just at the parameters.

yutian1224:
Thanks.
Is the "--memory-reservation" works too?

alu:
So far, it doesn't work for that.

schmunk42:
I noticed something similar, when images are not pulled beforehand on every node.

vieux:
hi @yutian1224, as @alu explained this is the expected behavior.
If you don't use any CPU (-c) or Memory (-m) resource, swarm is going to simply use the number of containers to spread evenly, that's why the number of containers in the same on every node.

I added a few tests in this PRs https://github.com/docker/swarm/pull/1657 to check that it's working as expected, I added a test were a node has twice as much CPU/Memory as the other once and Swarm places twice as much containers on the bigger node, as expected.

@schmunk42 it it's a private image it's expected since swarm applies an image affinity to make the run succeed (since swarm can pull private images during docker run, only in docker pull)

schmunk42:
@vieux Is there an issue about this to track? I looked around, but found nothing specific; also `docker-compose` related in my case.

vieux:
@schmunk42 since we don't want to store user's credentials in swarm, we nothing much we can do about it :/

amitshukla:
@yutian1224 does the above answer your question?

yutian1224:
yes,thanks for all

PR #1657, opened 2016-01-19
Add more unit tests for spread strategy

vieux:
Related to #1648

### DIFF
--- a/scheduler/strategy/spread_test.go
+++ b/scheduler/strategy/spread_test.go
@@ -8,6 +8,44 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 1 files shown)


MHBauer:
LGTM as it is.

Is there any concept of a memory heavy or cpu heavy balancing that would need to be tested? (i.e. same number of cpus, but one with twice is much memory, or one with same amount of memory, but twice as much cpus?) Not sure if relevant, just want to either cover the space or discount it.

abronan:
@vieux Should we add a test for memory too? Also I'm wondering the same thing than @MHBauer, maybe adding a more intricate test with varying CPUs, RAM in a completely unbalanced way could be interesting even though outside of the scope of this PR.

LGTM as is though

PR #1662, opened 2016-01-21
Enforce minimum backoff to avoid simultaneous validation on one engine

dongluochen:
`validatePendingEngine` can be called at first join, or from a monitoring thread. Current code allows both happen on the same engine simultaneously. The monitoring thread has a logic to check if it's the right time to do validation. It shouldn't validate an engine when it's just joined. This change is to enforce that.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -239,7 +239,9 @@ func (e *Engine) TimeToValidate() bool {
 	}
 	sinceLastUpdate := time.Since(e.updatedAt)
 	// Increase check interval for a pending engine according to failureCount and cap it at a limit
-	if sinceLastUpdate > validationLimit || sinceLastUpdate > time.Duration(e.failureCount)*failureBackoff {
+	// '+1' would enforce a minimum backoff because e.failureCount could be 0 at first join, or
+	// the engine has a duplicate ID
+	if sinceLastUpdate > validationLimit || sinceLastUpdate > time.Duration(e.failureCount+1)*failureBackoff {
 		return true
 	}
 	return false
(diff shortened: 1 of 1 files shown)


dongluochen:
I'm working on integration tests but stopping/restarting engines seems quite fragile in current integration tests.

abronan:
LGTM as is

Although I think we should make the backoff exponential sooner or later, to better find an acceptable rate of validation for the Engine. Even though this requires a slight change in `cluster.go` which checks engines at regular intervals. It will check often when an engine first appears and decrease the rate of checking in an non monotonic way.

It's mostly a problem when an engine first appears and an issue occurred, maybe it's more acceptable to have a rate of retry which would be `3-6-12-24-48-96`, etc. rather than `30-60-90-120`, etc. because you then have to wait `30s` to have this Engine validated and ready to be used.

ping @vieux @nishanttotla

dongluochen:
@abronan Thanks for the comment. Exponential backoff is the original idea. But I met 2 issues so I have this linear backoff first. I'm trying to find a better solution to have exponential backoff.  
- 2 consecutive validation shouldn't overlap. So 0, 3, 6 seconds might get into that space. In previous checkin duplicate validation result is discard so it'd be safe though. 
- The duplicate ID case, which is half the problem of the `Pending` state, would not backoff because failCount is 0. `failureCount` counts the number of connection failures. In duplicate ID case connection is successful so failureCount is 0.     
A comprise is backoff = 2 ^ failureCount + 30. WDYT?

nishanttotla:
@dongluochen I think `2^failureCount + 30` is a reasonable way to do it, white not letting 2 consecutive validations overlap. If the first validation fails though, the minimum is still 30, so is it possible to have the exponential backoff start as soon as we the first validation failed? Essentially something like `2^failureCount + firstValidationFailureTime`?

vieux:
LGTM

PR #1669, opened 2016-01-22
Support 'docker update' command

dongluochen:

### DIFF
--- a/api/primary.go
+++ b/api/primary.go
@@ -70,6 +70,7 @@ var routes = map[string]map[string]handler{
 		"/containers/{name:.*}/restart":       proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/start":         proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/stop":          proxyContainerAndForceRefresh,
+		"/containers/{name:.*}/update":        proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/wait":          proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/resize":        proxyContainer,
 		"/containers/{name:.*}/attach":        proxyHijack,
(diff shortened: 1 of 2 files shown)


vieux:
LGTM !

vieux:
@dongluochen I would do a `docker help | grep -q update` and skip the test if the update command in unknown.

abronan:
@dongluochen I'm probably missing something obvious but I'm a bit concerned about the scheduler.

Can't we fall into a race where an `update` is issued for a container on a given node, but we try to start another container with what should be an invalid amount of RAM (which will overcommit) on the same node? `UsedCPUs` can be triggered at a moment where the container specs update is not yet processed thus the information might be stale at the scheduler level.

Supposedly because the lock on container `Refresh` is bounded to the `container` object while `UsedCPUs` is locked on the `engine` object.

For example on a node with **8GB** start a container with **4GB** reservation, update it to **6GB** reservation and make sure you can't start another container on the same node with **4GB** reservation. The second request should fail in this case.

Schedule and update containers very quickly and there is a chance I think to allow the second container to be scheduled and started even though it shouldn't. In this case using `update` could be "dangerous".

Maybe worth adding a test on `docker info` after container updates?

Unless this falls under the scope of another PR?

/cc @vieux

dongluochen:
@abronan Thanks for the comment. I added docker info to reflect swarm's resource management. 

On the other hand, Engine is not rejecting requests over its resource limit. To make it work end to end, Engine may need to support that.

dongluochen:
@vieux Thanks for the suggestion. I skip docker version without `update` support.

vieux:
I agree with @dongluochen I think it should be fixed upstream in engine. It's good enough for now.

abronan:
thanks @dongluochen. At least we should support `update` so LGTM, mentioned this issue just so we can keep this in mind.

ping @vieux

abronan:
Woop done already haha :tada:

PR #1670, opened 2016-01-22
enable cross compile in travis again

vieux:
Let's enable it again since with go *1.5* it *so much simpler* than before (https://github.com/docker/swarm/pull/210)

ping @tianon


abronan:
OH shiny! LGTM

dongluochen:
LGTM. Thanks Victor!

tianon:
:metal:

It's also simper on the Travis side now, with "gimme": https://github.com/travis-ci/gimme#travisyml (if you wanted to consider going back to a matrix, but what you've got here is nice and simple :smile:)

Issue #1671, opened 2016-01-22
/info doesn't show container category correctly


dongluochen:
From swarm output containers in `Running`, `Paused` and `Stopped` are not counted. 
```
dchen@vm4:~/go/src/github.com/docker/swarm/test/integration$ docker -H vm2:2372 info
Containers: 33
 Running: 0
 Paused: 0
 Stopped: 0
Images: 19
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 1
 vm7: 192.168.56.207:4444
  └ Status: Healthy
  └ Containers: 33
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 30 MiB / 4.053 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.19.0-25-generic, operatingsystem=Ubuntu 14.04.3 LTS, storagedriver=aufs
  └ Error: (none)
  └ UpdatedAt: 2016-01-22T22:24:10Z
Plugins: 
 Volume: 
 Network: 
CPUs: 2
Total Memory: 4.053 GiB
Name: vm2
WARNING: No oom kill disable support
WARNING: No cpu cfs quota support
WARNING: No cpu cfs period support
WARNING: No cpu shares support
WARNING: No cpuset support
```
Here is the output from Docker daemon. 
```
dchen@vm4:~/go/src/github.com/docker/swarm/test/integration$ docker -H 192.168.56.207:4444 info
Containers: 33
 Running: 0
 Paused: 0
 Stopped: 33
Images: 322
Server Version: 1.10.0-rc1
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 392
 Dirperm1 Supported: true
Execution Driver: native-0.2
Logging Driver: json-file
Plugins: 
 Volume: local
 Network: bridge null host
Kernel Version: 3.19.0-25-generic
Operating System: Ubuntu 14.04.3 LTS
OSType: linux
Architecture: x86_64
CPUs: 2
Total Memory: 3.86 GiB
Name: vm7
ID: BDUL:67B7:IYN2:YAS6:7PIB:DTMA:KBGU:YSQN:TXP6:BO25:ZZIQ:SS7H
WARNING: No swap limit support
```

cc @vieux

PR #1684, opened 2016-01-26
Update info

vieux:
fix https://github.com/docker/swarm/issues/1671

ping @dongluochen

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -29,23 +29,45 @@ const APIVERSION = "1.21"
 
 // GET /info
 func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
-	info := dockerclient.Info{
-		Containers:        int64(len(c.cluster.Containers())),
-		Images:            int64(len(c.cluster.Images().Filter(cluster.ImageFilterOptions{}))),
+	info := apitypes.Info{
+		Images:            len(c.cluster.Images().Filter(cluster.ImageFilterOptions{})),
 		DriverStatus:      c.statusHandler.Status(),
-		NEventsListener:   int64(c.eventsHandler.Size()),
+		NEventsListener:   c.eventsHandler.Size(),
 		Debug:             c.debug,
 		MemoryLimit:       true,
 		SwapLimit:         true,
+		CPUCfsPeriod:      true,
+		CPUCfsQuota:       true,
+		CPUShares:         true,
+		CPUSet:            true,
 		IPv4Forwarding:    true,
 		BridgeNfIptables:  true,
-		BridgeNfIp6tables: true,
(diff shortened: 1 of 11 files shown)


MHBauer:
wow. I just realized that swarm could be running on windows and absolutely nothing would change. 

blows my mind :open_mouth:

nishanttotla:
It seems like you're setting a default value for `KernelVersion` (`"<unknown>"`) for the `/info` endpoint, but for the `/version` endpoint there's no such default. Can we make that consistent?

Apart from that, LGTM.

vieux:
@nishanttotla thanks, I updated

nishanttotla:
Great, LGTM :)

dongluochen:
Is stable docker API version not showing container categories like 'Running', 'Stopped'? In that case we may need to skip the test for stable.

nishanttotla:
@dongluochen I think `engine-api` was going to be vendored in at the end of the week (as per @calavera). Should that be enough?

vieux:
@nishanttotla stable means engine 1.9 and this feature is only present in 1.10

vieux:
thanks @dongluochen I'll update

vieux:
test failure are unrelated, they will probably be fixed by https://github.com/docker/swarm/pull/1690

@dongluochen can you please review again ?

dongluochen:
LGTM.

api/handlers.go
```diff
@@ -29,23 +29,45 @@ const APIVERSION = "1.21"
...
 		NoProxy:           os.Getenv("no_proxy"),
 		SystemTime:        time.Now().Format(time.RFC3339Nano),
+		ExperimentalBuild: experimental.ENABLED,
+	}
+
+	if kernelVersion, err := kernel.GetKernelVersion(); err == nil {
+		info.KernelVersion = kernelVersion.String()
+	}
+
+	for _, c := range c.cluster.Containers() {
```
abronan: @vieux I don't think this is the most efficient way to do it, we are looping across all the containers in the cluster, if we have many tens of thousands of containers, this is a bit overkill (especially knowing that `Containers()` also loops through all the engine's container entries first). `info` provides the number of `Paused` and `Running` containers now thanks to changes in the Engine API, might be better to use those values instead I think.

Knowing that the rate of creation of containers is pretty stable after going through the scheduling process, calculating this from the cached values on `info` for every engine will be more efficient and still provide with meaningful values (unless you run `docker info` every second :smile:).

vieux: ok if you don't mind, merge this one and I'll start working on this on another PR

abronan: Sounds perfect :) thanks @vieux!

PR #1685, opened 2016-01-26
Adding .DS_Store to .gitignore for Mac developers

nishanttotla:
Mac filesystem creates `.DS_Store` files to keep track of folder attributes, and we don't want these to be committed to the repo. Many developers don't have global git settings to ignore these files. `docker/docker` explicitly ignores them too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 /swarm
+.DS_Store


chanwit:
LGTM
Test failed is unrelated to this change as it's trivial and not touch the code.

I'm not a Mac user actually. If other maintainers think it's useful please consider merging.

MHBauer:
sure why not. Somebody would have to be doing something very weird to have this happen (and on not-OSX), but we might as well be proactive in preventing it.

vieux:
NOT LGTM on my side,
I think you should use the global git ignore.

if we merge this one, we will end up merging `*~` and plenty of other stuff

I think it's better to teach the very few people that submit changes with .DS_Store committed (I don't think it ever happened here)

abronan:
I understand the intent but I'm leaning towards @vieux's side, I never saw a single case of someone committing `.DS_Store` and this should be done through the global git ignore instead as a good practice.

So NOT LGTM.

@dongluochen WDYT?

nishanttotla:
My reasoning was that if someone committed `.DS_Store` files using `git add .` and not checking what they were adding, it just slows down the workflow since we have to ask them to update the PR. In particular, some code plugins use these files (as I found out yesterday), so it's a weird corner case as @MHBauer pointed out. Of course, experienced programmers will know to look out for this in the first place, but perhaps not everyone.

That said, I get @vieux's reasoning too (the `.gitignore` in `docker/docker` is very long for example), so if everyone thinks that this hasn't been an issue in the past, then it's not needed and I'm happy to close the PR.

dongluochen:
I understand the intention of PR. But I think platform/IDE related files shouldn't get into code base, same reasons @vieux @abronan stated.

nishanttotla:
Thanks for the comments everyone. I'm going to close this PR, but if anyone has suggestions for things to add in `.gitignore`, I can reopen.

Issue #1655, opened 2016-01-19
Swarm and mesos resource consolidation


Spritekin:
This is the other side of this issue I reported here:
https://github.com/docker/swarm/issues/1588

Summary:
Essentially, I set the offer timeout to a higher value (120s) in order to reduce the overload to the servers and subsequent crash and restart of the Swarm service. Unfortunately this leads to Swarm acquiring the resources for too long so Mesos is unable to reintegrate the resources.

Description:
In my cluster, in order to simulate CPU Shares (currently not available in Swarm) I declared each CPU as 100 shares in mesos (i.e. for a machine with 4 CPUs I declared resources=cpus(*):400). 

That allows the developers to execute a job in, say, 10 CPUs which would be 10 shares. It works fine and devs are happy.

But now we are noticing other problems. After a while I get many fragments.. like 33 shares and 25 shares in the same server and they don't seem to consolidate back (or are just taking too long) which leads to a resource fragmentation.

So I reduced the offer timeout to 60s. This allows mesos to recover resources faster allowing to reintegrate them. After a while I get this list where you can see the reintegration happening. You can see that there are 66 and 99 CPUs which we don't have, so those are essentially merged jobs which previously used 33CPUs.

**…a770-089ebb649af5-O4606  swarm 172.31.43.230 99  1.3 GB**
…a770-089ebb649af5-O4604  swarm 172.31.39.40  400 14.7 GB
…a770-089ebb649af5-O4603  swarm 172.31.38.170 400 14.7 GB
**…a770-089ebb649af5-O4602  swarm 172.31.40.133 66  900 MB**
…a770-089ebb649af5-O4600  swarm 172.31.41.72  33  450 MB
…a770-089ebb649af5-O4599  swarm 172.31.32.23  400 14.7 GB
…a770-089ebb649af5-O4598  swarm 172.31.47.54  268 11.4 GB
…a770-089ebb649af5-O4594  swarm 172.31.46.157 33  450 MB
…a770-089ebb649af5-O4593  swarm 172.31.35.94  4 6.3 GB
**…a770-089ebb649af5-O4592  swarm 172.31.40.133 66  900 MB**
…a770-089ebb649af5-O4591  swarm 172.31.35.94  33  450 MB
…a770-089ebb649af5-O4590  swarm 172.31.37.136 400 14.7 GB
**…a770-089ebb649af5-O4588  swarm 172.31.40.133 4 6.3 GB**
[...]

Issue #1588, opened 2016-01-07
Swarm on Mesos overloads slave docker daemons


Spritekin:
Hi,
I have noticed a few irregularities since I installed Swarm 1.0.1. It does not mean its a 1.0.1 problem. It does mean I just noticed this.
In summary the docker daemons in the nodes start gaining more and more loaded even if no job is sent at all. 
Eventually they get over 100% CPU usage and they become so slow an image pull can take about 5 mins instead of the 30 secs it usually takes. All docker commands are irresponsive, a 'docker image' can take 1 min to respond directly in the slave node.

Setup
Running on CoreOS 835.8.
Docker 1.8.3 (part of CoreOS 835.8)
Mesos master and two slaves. (mesos 0.25)
Swarm 1.0.1

I start swarm as:
docker run --net=host \
    --name=swarm-manager \
    -e SWARM_MESOS_USER=root \
    -e SWARM_MESOS_TASK_TIMEOUT=180s \
    -e SWARM_MESOS_OFFER_TIMEOUT=5s \
    swarm manage \
        --strategy "binpack" \
        -c mesos-experimental \
        --cluster-opt mesos.address=${COREOS_PRIVATE_IPV4} \
        --cluster-opt mesos.port=3375 \
        mesos-master.aws.cmcrc.com:5050

The same happens when I connect directly to the mesos master or ZK. I started swarm 10 mins ago and look at the CPU usage. And its slowly building. 
And when I kill Swarm the CPU usage in the slaves goes down to 0.

top - 00:04:04 up 22:13,  1 user,  load average: 0.11, 0.50, 0.55
Tasks:  90 total,   1 running,  89 sleeping,   0 stopped,   0 zombie
%Cpu(s):  5.1 us,  1.0 sy,  0.0 ni, 93.7 id,  0.0 wa,  0.1 hi,  0.0 si,  0.1 st
KiB Mem:  16436532 total,  4821520 used, 11615012 free,   660512 buffers
KiB Swap:        0 total,        0 used,        0 free.  2806856 cached Mem

  PID USER      PR  NI    VIRT    RES    SHR S  %CPU %MEM     TIME+ COMMAND                                                       
  828 root      20   0 1654296 481688  18676 S  <b>15.3</b>  2.9 954:42.82 docker                                                        

I have also tried with a SWARM_MESOS_OFFER_TIMEOUT of 15s and its the same case. I know 1.0.1 has this param as a default of 30s.

[...]

PR #1688, opened 2016-01-27
Fixes swarm overloading slaves when using the mesos drivers.

Spritekin:
The problem is described here:
https://github.com/docker/swarm/issues/1588

This fix also allows reducing the offer timeout which reduces problems with mesos resource consolidation described here:
https://github.com/docker/swarm/issues/1655

Close https://github.com/docker/swarm/issues/1588
Close https://github.com/docker/swarm/issues/1655


Spritekin:
The error is in a container not properly removed at the end of the process.

jimenez:
LGTM, Thanks @Spritekin  for the thorough report.

vieux:
LGTM

PR #1690, opened 2016-01-27
swarm_manage should confirm nodes pass validation

dongluochen:
With node management change, node starts with Pending state. `swarm_manage` should confirm nodes pass validation, which is the intention of this function. 

Thank @vieux for finding this. 


vieux:
LGTM

chanwit:
LGTM

Issue #1625, opened 2016-01-12
Swarm API /info endpoint has \b chars


nishanttotla:
This was done to make the output human readable, by adding these extra characters and piggybacking on top of `DriverStatus` field in the Docker Engine API ([here](https://github.com/docker/docker/blob/c1582f20cc6ff1abe576e5aabb60c73309055556/vendor/src/github.com/docker/engine-api/types/types.go)).

The proposal is to add a new field in the Docker Engine API for Swarm to use so that the `\b` can be removed.

Some relevant discussion on this was previously at #735, #1214, and #1147.

Corresponding issue in `docker/docker`: https://github.com/docker/docker/issues/19277

cc @vieux

PR #1696, opened 2016-01-28
Extracting apiversion from HTTP request

nishanttotla:
We need the API version supplied by the HTTP request for #1625. This also improves the regex that defines allowed version values. Previously malformed versions like `v1.2323.535` etc were allowed, now they'll be well formed decimals.

### DIFF
--- a/api/primary.go
+++ b/api/primary.go
@@ -18,6 +18,7 @@ type context struct {
 	statusHandler StatusHandler
 	debug         bool
 	tlsConfig     *tls.Config
+	apiVersion    string
 }
 
 type handler func(c *context, w http.ResponseWriter, r *http.Request)
@@ -149,11 +150,12 @@ func setupPrimaryRouter(r *mux.Router, context *context, enableCors bool) {
 				if enableCors {
 					writeCorsHeaders(w, r)
 				}
+				context.apiVersion = mux.Vars(r)["version"]
 				localFct(context, w, r)
 			}
 			localMethod := method
 
-			r.Path("/v{version:[0-9.]+}" + localRoute).Methods(localMethod).HandlerFunc(wrap)
+			r.Path("/v{version:[0-9]+.[0-9]+}" + localRoute).Methods(localMethod).HandlerFunc(wrap)
 			r.Path(localRoute).Methods(localMethod).HandlerFunc(wrap)
 
 			if enableCors {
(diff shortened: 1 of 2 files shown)


nishanttotla:
cc @vieux

nishanttotla:
@vieux fixed the test, it should succeed now.

vieux:
LGTM

api/primary.go
```diff
@@ -18,6 +18,7 @@ type context struct {
 	statusHandler StatusHandler
 	debug         bool
 	tlsConfig     *tls.Config
+	apiversion    string
```
dongluochen: apiversion => apiVersion

api/primary.go
```diff
@@ -149,11 +150,12 @@ func setupPrimaryRouter(r *mux.Router, context *context, enableCors bool) {
 				if enableCors {
 					writeCorsHeaders(w, r)
 				}
+				context.apiversion = mux.Vars(r)["version"]
 				localFct(context, w, r)
 			}
 			localMethod := method
 
-			r.Path("/v{version:[0-9.]+}" + localRoute).Methods(localMethod).HandlerFunc(wrap)
+			r.Path("/v{version:[0-9]+.[0-9]+}" + localRoute).Methods(localMethod).HandlerFunc(wrap)
```
dongluochen: Do we have API version format defined somewhere? With this change, "v3", "v1.3.1" are not valid.

api/primary.go
nishanttotla: Fixed this, thanks.

api/primary.go
nishanttotla: @dongluochen I couldn't find one. We don't seem to be using this version anywhere in Swarm so far, I believe it's only meant for the API to mimic that of Engine. What do you suggest?

MHBauer:
How is `v1.2323.535` malformed? Unlikely, yes, but not wrong.

api/primary.go
dongluochen: I suggest no change :)
Or we define it (or find it) in Engine API and follow that.

nishanttotla: Alright, let me ask someone in Engine and figure it out.

nishanttotla:
@MHBauer see discussion with @dongluochen above. I'll try to find out what the acceptable versions are. Currently the Swarm API doesn't really use the version AFAIK, even though users can provide it. Since we have this to mimic Engine API, I think it would make sense to just use what Engine uses.

vieux:
engine format is `x.xx`

vieux:
IMO we should merge it

nishanttotla:
To confirm @vieux's previous comment, here's where Engine defines API version: https://github.com/docker/docker/blob/master/api/common.go

It would perhaps also be nice if we have a `minVersion` like Engine, which will allow us to more closely control API changes, but we can handle that in a later PR.

cc @dongluochen @MHBauer

dongluochen:
LGTM.

PR #1697, opened 2016-01-28
Fix double start with Mesos

jimenez:
There is no difference between create and start in Mesos so Docker run was technically doing 2 starts.
This PR is to prevent a second start to happen.
Signed-off-by: Isabel Jimenez <contact@isabeljimenez.com>

### DIFF
--- a/api/primary.go
+++ b/api/primary.go
@@ -69,7 +69,7 @@ var routes = map[string]map[string]handler{
 		"/containers/{name:.*}/unpause":       proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/rename":        postRenameContainer,
 		"/containers/{name:.*}/restart":       proxyContainerAndForceRefresh,
-		"/containers/{name:.*}/start":         proxyContainerAndForceRefresh,
+		"/containers/{name:.*}/start":         postContainersStart,
 		"/containers/{name:.*}/stop":          proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/update":        proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/wait":          proxyContainerAndForceRefresh,
--- a/cluster/mesos/cluster.go
+++ b/cluster/mesos/cluster.go
@@ -177,6 +177,15 @@ func (c *Cluster) UnregisterEventHandler(h cluster.EventHandler) {
 	c.eventHandlers.UnregisterEventHandler(h)
 }
 
+// StartContainer starts a container
+func (c *Cluster) StartContainer(container *cluster.Container) error {
+	// if the container was started less than a second ago in detach mode, do not start it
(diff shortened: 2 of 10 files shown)


api/primary.go
```diff
@@ -68,7 +68,7 @@ var routes = map[string]map[string]handler{
 		"/containers/{name:.*}/unpause":       proxyContainerAndForceRefresh,
 		"/containers/{name:.*}/rename":        postRenameContainer,
 		"/containers/{name:.*}/restart":       proxyContainerAndForceRefresh,
-		"/containers/{name:.*}/start":         proxyContainerAndForceRefresh,
+		"/containers/{name:.*}/start":         postContainersStart,
```
abronan: Does this have a side effect on the Swarm side? (if we omit the force refresh)

/cc @vieux

test/integration/mesos/api/logs.bats
```diff
@@ -29,6 +29,7 @@ function teardown() {
 	# verify
 	run docker_swarm logs test_container
 	[ "$status" -eq 0 ]
+[ "${#lines[@]}" -eq 3 ]
```
abronan: nit: small indentation issue :ghost:

cluster/mesos/cluster.go
```diff
@@ -177,6 +177,15 @@ func (c *Cluster) UnregisterEventHandler(h cluster.EventHandler) {
 	c.eventHandlers.UnregisterEventHandler(h)
 }
 
+// StartContainer starts a container
+func (c *Cluster) StartContainer(container *cluster.Container) error {
+	// if the container was started less than a second ago, do not start it
+	if time.Now().Unix()-container.Created > 1 {
```
tnachen: Can we provide more comments why we even check this and why this is possible?

tnachen: And also if somehow there is a delay more than 1 second of the second request we will still see double start right?

vieux: yes, but if the container is already started, start does nothing so it's not an issue.

The issue is for cases like `docker run busybox ls`

api/primary.go
vieux: nope seems good to me

test/integration/mesos/api/logs.bats
jimenez: fixed

vieux:
LGTM, test failure unrelated ping @abronan

abronan:
LGTM

nishanttotla:
@vieux @abronan @dongluochen one side-effect of this change is that the HostConfig isn't being passed on when starting a container, causing issues like #1781. I think it's a bug and needs to be fixed.

vieux:
@nishanttotla ok let's fix it, it's quite trivial, but

HostConfig is deprecated see the note in  https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#start-a-container the hostconfig should be sent at create instead.

PR #1711, opened 2016-01-29
Allow more retries for node validation

dongluochen:
Jenkins reported several failures from 'nodes_validated' recently while I cannot reproduce locally. It's possible that Jenkins runs at heavy loaded machine and takes more time to finish validation. 


abronan:
Indeed, this seems like an intermittent timing issue.

LGTM

nishanttotla:
LGTM

Issue #1709, opened 2016-01-29
Regression "label affinity in parallel" is always failing


chanwit:
I did regression locally a lot on my boxes recently as I cannot wait `janky` to finish the test and found that this test "label affinity in parallel" is always failing on my boxes.

Could any one confirm this?

```
not ok 4 label affinity in parallel
# (in test file ./affinities.bats, line 141)
#   `[ "${#lines[@]}" -eq  2 ]' failed
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing discovery service" name=nodes uri="127.0.0.1:5230,127.0.0.1:5231"
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing strategy" name=spread
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing filter" name=health
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing filter" name=port
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing filter" name=dependency
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing filter" name=affinity
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing filter" name=constraint
# time="2016-01-29T20:31:13Z" level=debug msg="Initializing cluster" name=swarm
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=HEAD route="/containers/{name:.*}/archive"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/networks/{networkid:.*}"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/_ping"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/images/json"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/images/{name:.*}/history"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/containers/ps"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/containers/{name:.*}/changes"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/containers/{name:.*}/stats"
# time="2016-01-29T20:31:13Z" level=debug msg="Registering HTTP route" method=GET route="/exec/{execid:.*}/json"
[...]

PR #1714, opened 2016-01-30
	fix soft / resource computation

### LINKED ISSUE #1698: Node resources are not considered in soft filter
Hi,

I have a concern about how soft filter works, let me explain it with an example:
There is only one node in swarm cluster: `node1` which has 4GB memory and a label `env=prod`, now user creates a container which requests 3GB memory and with a soft filter `constraint:env==~dev`, so the container can be created successfully in `node1` since the filter is soft and `node1` has enough memory for the container. So far so good.

But if there are two nodes in the cluster: `node1` (same with the above `node1`) and `node2` which has 2GB memory and a label `env=dev`, and user still creates a container which requests 3GB memory and with a soft filter `constraint:env==~dev`, but this time, the container will fail to be created due to no available nodes (`node2` does not have enough resources for the container). But I think it seems this does not make sense, the same container can be created when no nodes have the required label, but can NOT be created when there is a node has the required label.

IMHO, the container should be created in both of these two cases (in `node1`).

vieux:
The issue was clearly explained in #1698, thanks @zhq527725

Fix #1698

### DIFF
--- a/scheduler/filter/filter.go
+++ b/scheduler/filter/filter.go
@@ -54,17 +54,7 @@ func New(names []string) ([]Filter, error) {
 }
 
 // ApplyFilters applies a set of filters in batch.
-func ApplyFilters(filters []Filter, config *cluster.ContainerConfig, nodes []*node.Node) ([]*node.Node, error) {
-	candidates, err := applyFilters(filters, config, nodes, true)
-
-	if err != nil {
-		candidates, err = applyFilters(filters, config, nodes, false)
-	}
-
-	return candidates, err
-}
-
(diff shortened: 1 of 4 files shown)


vieux:
ping @abronan @dongluochen

abronan:
@vieux The test failed on travis :cry:

vieux:
@abronan whoops my bad

abronan:
LGTM (failure related to #1709, so should be good to merge)

vieux:
ping @dongluochen

vieux:
ping @nishanttotla

dongluochen:
@vieux LGTM for 1.1.0. I think there are some duplicate computation with current approach. I'd suggest we revisit it in the later releases.

vieux:
@dongluochen agreed

PR #1718, opened 2016-01-31
How to get Swarm and Machine provisioning

### LINKED ISSUE #1580: Inconsistencies in commands between manual setup and HA setup guides
Hi all, I recently walked through the process of establishing a multi-manager Docker Swarm cluster and noticed some inconsistencies in the Docker commands between the multi-manager setup guide (1) and the manual Swarm setup guide (2). For example, the manual setup guide uses commands like `docker run -d swarm manage...` whereas the HA setup guide just says `swarm manage...`. I'm not clear why the `docker run` part of these commands was omitted in the HA setup guide.

I'm happy to submit a PR (this will be my first time contributing) to resolve these inconsistencies, assuming these inconsistencies aren't intentional. Am I missing a reason why the commands are different (aside from the flags, of course)?

(1): https://docs.docker.com/swarm/multi-manager-setup/
(2): https://docs.docker.com/swarm/install-manual/

moxiegirl:
Adding a how to get Swarm
Fix and close #1580 
Supplements https://github.com/docker/machine/pull/2957

### DIFF
--- a/docs/provision-with-machine.md
+++ b/docs/provision-with-machine.md
@@ -1,0 +1,190 @@
+<!--[metadata]>
++++
+title = "Provision with Machine"
+description = "Provision with Machine"
+keywords = ["docker, Swarm, clustering, provision, Machine"]
+[menu.main]
+parent="workw_swarm"
+weight=5
++++
+<![end-metadata]-->
+
+
+# Provision a Swarm cluster with Docker Machine
+
+You can use Docker Machine to provision a Docker Swarm cluster. Machine is the
+Docker provisioning tool. Machine provisions the hosts, installs Docker Engine
+on them, and then configures the Docker CLI client. With Machine's Swarm
+options, you can also quickly configure a Swarm cluster as part of this
+provisioning.
+
+This page explains the commands you need to provision a basic Swarm cluster on a
+local Mac or Windows computer using Machine. Once you understand the process,
+you can use it to setup a Swarm cluster on a cloud provider, or inside your
+company's data center.
+
+If this is the first time you are creating a Swarm cluster, you should first
+learn about Swarm and its requirements by [installing a Swarm for
+evaluation](install-w-machine.md) or [installing a Swarm for
+production](install-on-aws.md). If this is the first time you have used Machine,
+you should take some time to [understand Machine before
+continuing](https://docs.docker.com/machine).
+
+
+## What you need
+
+If you are using Mac OS X or Windows and have installed with Docker Toolbox, you
+should already have Machine installed. If you need to install, see the
+instructions for [Mac OS X](https://docs.docker.com/engine/installation/mac/) or
+[Windows](https://docs.docker.com/engine/installation/mac/).
+
+Machine supports installing on AWS, Digital Ocean, Google Cloud Platform, IBM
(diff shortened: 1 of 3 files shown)


docs/multi-manager-setup.md
```diff
@@ -11,15 +11,15 @@ weight=3
 
 # High availability in Docker Swarm
```
moxiegirl: @abronan Can you check and make sure I add the `docker run -d swarm` appropriately?

abronan: @moxiegirl I think it's a bit tricky for `<tls-config-flags>` in this case because the TLS certs needs to be mounted as a volume to make sure the container sees the certs. I'm afraid that this would cause confusion. Maybe this can be a part added to the TLS docs that explains how to mount a volume with the certs when using the Swarm image (if not done yet). In this case no need to mention it here we can redirect to the TLS docs.

Also we should enforce using `docker run --net=host swarm <...>` because of multiple problems that could occur on the networking side as a good practice.

@vieux @dongluochen Can you confirm? Did I forget something on running Swarm with the Image?

vieux: `--net=host` improves performences but I'm not sure we want to enforce it.

`<tls-config-flags>` I agree with alex, you need to mount them in the container, and it might confuse people if we explain this here.

moxiegirl: OK. As we are late in the game. I'm gonna remove the HA work.  We can test and add it back in later.

moxiegirl:
@vieux @abronan Ok, I pulled the HA work until we get time to test it.

If everything else looks good can I get an LGTM pls?

MHBauer:
80 char word wrap in some places and not in others. Would be nice to be consistent.

docs/provision-with-machine.md
```diff
@@ -0,0 +1,174 @@
...
+If you are connecting to typical host environment with Machine, you use the `env` subcommand, like this:
+
+```
+eval "$(docker-machine env local)"
+```
+
+Docker Machine provides a special `--swarm` flag with its `env` command to connect to Swarm nodes.
+
+```
+docker-machine env --swarm HOST_NODE_NAME
```
MHBauer: This is a command followed by output, right? Can we separate this first line or otherwise indicate that?

MHBauer:
What's the easiest way to check the rendering of the docs? I'm not sure what the equivalent of `make docs`.

docs/provision-with-machine.md
moxiegirl: @MHBauer This is syntax.  So,  I don't provide output for that.  The real issue is the previous command which does show a node name.

moxiegirl:
`make docs` is the way to check. From the `docs` run make docs and go to the page.  You'll need to pull this pr.

docs/provision-with-machine.md
moxiegirl: I'm not sure what you mean, the output is shown --- that's what lines 134-139 are.

moxiegirl:
Broken doc build is due to the fact I'm linking to two documents in the PR that Rolfe is merging.

abronan:
Overall LGTM for this one

vieux:
LGTM ping @mgoelzer

PR #1724, opened 2016-02-01
Adding default name (unknown) for pending engine

### LINKED ISSUE #1689: Use (unknown) name for pending nodes
```
$ docker -H :2375 info
Containers: 2
Images: 6
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 3
 : xx.xx.xx.xx:2375
  └ Status: Pending
  └ Containers: 0
  └ Reserved CPUs: 0 / 0
  └ Reserved Memory: 0 B / 0 B
  └ Labels:
  └ Error: i/o timeout. Are you trying to connect to a TLS-enabled daemon without TLS?
  └ UpdatedAt: 2016-01-27T02:45:51Z
 : 127.0.0.1:4444
  └ Status: Pending
  └ Containers: 0
  └ Reserved CPUs: 0 / 0
  └ Reserved Memory: 0 B / 0 B
  └ Labels:
  └ Error: Cannot connect to the docker engine endpoint
  └ UpdatedAt: 2016-01-27T02:45:41Z
node-3: xx.xx.xx.xx:2375
  └ Status: Healthy
  └ Containers: 2
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 1.016 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=4.2.0-23-generic, operatingsystem=Ubuntu 14.04.3 LTS, storagedriver=aufs
  └ Error: (none)
  └ UpdatedAt: 2016-01-27T02:45:41Z
CPUs: 2
Total Memory: 2.033 GiB
Name: macbook.local
```

When a node is in pending, it has no name (since we cannot connect to it, it's expected) but I find the UI not pretty 

Could we have `(unknown): 127.0.0.1:4444` instead of `: 127.0.0.1:4444`

/cc @dongluochen

nishanttotla:
Fixes #1689. This change only adds a default name to Engines whose validation is pending. We should open another issue to further discuss about what information needs to be shown (or hidden) for pending engines.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -837,7 +837,11 @@ func (c *Cluster) Info() [][2]string {
 	sort.Sort(cluster.EngineSorter(engines))
 
 	for _, engine := range engines {
-		info = append(info, [2]string{" " + engine.Name, engine.Addr})
+		engineName := "(unknown)"
+		if engine.Name != "" {
+			engineName = engine.Name
+		}
+		info = append(info, [2]string{" " + engineName, engine.Addr})
 		info = append(info, [2]string{"  └ Status", engine.Status()})
 		info = append(info, [2]string{"  └ Containers", fmt.Sprintf("%d", len(engine.Containers()))})
 		info = append(info, [2]string{"  └ Reserved CPUs", fmt.Sprintf("%d / %d", engine.UsedCpus(), engine.TotalCpus())})


cluster/engine.go
```diff
@@ -116,6 +116,7 @@ type Engine struct {
 func NewEngine(addr string, overcommitRatio float64, opts *EngineOpts) *Engine {
 	e := &Engine{
 		Addr:            addr,
+		Name:            "(unknown)",
```
vieux: I would like to have this added only in info.
It may have some side effects here.

nishanttotla: It's not clear to me how there can be side effects, but perhaps it makes sense to do this in Info anyway, since other display changes for pending engines (if any) will be in info.

vieux: maybe `docker run -e constraint:node==(unknown) ...`

vieux:
LGTM

nishanttotla:
Ping @dongluochen

dongluochen:
LGTM. thanks @nishanttotla.

Issue #1686, opened 2016-01-26
events api not streaming events


nicolaka:
Version : swarm:1.0.1

Setup: 3 Swarm Managers with 2 Swarm nodes

Issue: Sometimes issuing `docker events` against a Swarm manager will not stream any events that the docker engines are reporting. Restarting the Swarm managers fixed the issue. Could not reproduce.

Example:

Here i'm creating containers as I listen for swarm events. But no events are showing.

$ sudo docker -H tcp://127.0.0.1:2376  events

$

After restarting swarm manager , it works as expected.

MHBauer:
Was this a freshly set up swarm, or were you listening for events for a while? I know we have integration tests so listen for events, but I'm not sure how we'd test a problem that requires a swarm to sit around for a while.

nicolaka:
Nop, and not sure exactly what caused it but it did happen to me before with a single manager. I'll try to reproduce.

dongluochen:
@nicolaka Did the command exit by itself? Or did you send a Ctrl^C? What's swarm version?

nicolaka:
@dongluochen nop, it just stays silent. Swarm version is 1.0.1.

vieux:
@nicolaka I think I found a workaround https://github.com/docker/swarm/pull/1728

Issue #1203, opened 2015-09-08
docker events command doesn't work with swarm


garagatyi:
I tried to run swarm in accordance to docs at https://docs.docker.com/swarm/install-w-machine/.
Docker events doesn't return anything. Here is how I check it:

```
~$ docker -H tcp://192.168.99.101:3376 info
Containers: 12
Images: 5
Storage Driver: 
Role: primary
Strategy: spread
Filters: affinity, health, constraint, port, dependency
Nodes: 2
 swarm-agent-00: 192.168.99.102:2376
  └ Containers: 6
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 1.022 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=4.0.9-boot2docker, operatingsystem=Boot2Docker 1.8.1 (TCL 6.3); master : 7f12e95 - Thu Aug 13 03:24:56 UTC 2015, provider=virtualbox, storagedriver=aufs
 swarm-master: 192.168.99.101:2376
  └ Containers: 6
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 1.022 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=4.0.9-boot2docker, operatingsystem=Boot2Docker 1.8.1 (TCL 6.3); master : 7f12e95 - Thu Aug 13 03:24:56 UTC 2015, provider=virtualbox, storagedriver=aufs
Execution Driver: 
Kernel Version: 
Operating System: 
CPUs: 2
Total Memory: 2.043 GiB
Name: 729aa5032753
ID: 
Http Proxy: 
Https Proxy: 
No Proxy: 
$ time docker -H tcp://192.168.99.101:2376 events --since '2015-09-06T12:15' --until '2015-09-07T12:15' --filter 'event=start'
2015-09-08T13:19:54.000000000+03:00 729aa5032753c8181a0d3d12fa0d0a3f4017893da450655fa7daa732d70840d8: (from swarm:latest) start

2015-09-08T13:19:54.000000000+03:00 794093de3c6f2e266edce60e1a2f5ebf99dc8727e3755caed2c812760a145b7b: (from swarm:latest) start

2015-09-08T14:07:52.000000000+03:00 740df9cde5d730c0ea3db21fe1864da5ea1b5c4e8e3c31eb5fb8c8a120a8bdf5: (from ubuntu) start

[...]

PR #1728, opened 2016-02-02
retry monitor events on EOF

vieux:
Looks like StartMonitorEvents from dockerclient sometimes returns `EOF` when you try to monitor the events.

this should be fixed when using engine-api/client.

In the meantime, let's retry.

Fix https://github.com/docker/swarm/issues/1686
Fix https://github.com/docker/swarm/issues/1203

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -153,6 +153,23 @@ func (e *Engine) Connect(config *tls.Config) error {
 	return e.ConnectWithClient(c)
 }
 
+// StartMonitorEvents monitors events from the engine
+func (e *Engine) StartMonitorEvents() {
+	log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Debug("Start monitoring events")
+	ec := make(chan error)
+	e.client.StartMonitorEvents(e.handler, ec)
+
+	go func() {
+		if err := <-ec; err != nil && err.Error() != "EOF" {
+			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Errorf("Error monitoring events: %s", err)
+		} else if err != nil {
+			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Debug("EOF monitoring events, restarting")
+			e.StartMonitorEvents()
+		}
+		close(ec)
+	}()
+}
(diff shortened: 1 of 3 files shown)


vieux:
ping @abronan @dongluochen :green_heart:

abronan:
LGTM

cluster/engine.go
```diff
@@ -153,6 +153,22 @@ func (e *Engine) Connect(config *tls.Config) error {
...
 
+// StartMonitorEvents monitors events from the engine
+func (e *Engine) StartMonitorEvents() {
+	log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Debug("Start monitoring events")
+	ec := make(chan error)
+	e.client.StartMonitorEvents(e.handler, ec)
+
+	go func() {
+		if err := <-ec; err == nil || err.Error() != "EOF" {
+			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Errorf("Error monitoring events: %s", err)
```
dongluochen: Why `err == nil` is considered as error?

vieux: you are right, it should never return nil

cluster/engine.go
```diff
@@ -153,6 +153,22 @@ func (e *Engine) Connect(config *tls.Config) error {
...
+	log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Debug("Start monitoring events")
+	ec := make(chan error)
+	e.client.StartMonitorEvents(e.handler, ec)
+
+	go func() {
+		if err := <-ec; err != nil && err.Error() != "EOF" {
+			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Errorf("Error monitoring events: %s", err)
+		} else if err != nil {
+			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Debug("EOF monitoring events, restarting")
+			e.StartMonitorEvents()
```
dongluochen: I'm a little worry about recursive call without closing the channel.

vieux: I added a close below

dongluochen:
LGTM.

PR #1732, opened 2016-02-02
Fix three typo in docs/scheduler/filter.md

Uhsac:
Hi I find 3 typos while reading scheduler docs:
dafult -> default
by by -> by
want run -> want to run

I don't know if I have to open an issue for a typo, because I don't find contribution guidelines for this.
If I have to, just tell me, I can do it.

### DIFF
--- a/docs/scheduler/filter.md
+++ b/docs/scheduler/filter.md
@@ -53,7 +53,7 @@ When creating a container or building an image, you use a `constraint` or
 
 Node constraints can refer to Docker's default tags or to custom labels. Default
 tags are sourced from `docker info`. Often, they relate to properties of the Docker
-host. Currently, the dafult tags include:
+host. Currently, the default tags include:
 
 * `node` to refer to the node by ID or name
 * `storagedriver`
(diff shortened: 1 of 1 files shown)


nishanttotla:
LGTM. Thanks @Uhsac! You don't have to open an issue to correct typos.

docs/scheduler/filter.md
```diff
@@ -101,7 +101,7 @@ new containers.
 Continuing the previous example, assuming your cluster with `node-1` and
 `node-2`, you can run a MySQL server container on the cluster.  When you run the
 container, you can use a `constraint` to ensure the database gets good I/O
-performance. You do this by by filter for nodes with flash drives:
+performance. You do this by filter for nodes with flash drives:
```
abronan: Very nit because I'm not an english native, but that sentence sounds weird to me, maybe:

```
by filter for nodes -> by filtering for nodes
```

nishanttotla: @abronan I think in the former case `filter` is being used as a noun. In which case `by using filter for nodes` would be more appropriate, or `by filtering for nodes` as you suggested.

vieux:
LGTM

ping @moxiegirl

docs/scheduler/filter.md
vieux: according to @moxiegirl :pray: please use `by filtering for nodes` as @abronan suggested

Uhsac:
I fixed the PR according to @abronan said

MHBauer:
LGTM

abronan:
Cool, thanks @Uhsac!

LGTM

PR #1735, opened 2016-02-02
update dockerclient

vieux:


vieux:
ping @dongluochen

vieux:
ping @abronan

vieux:
@dongluochen today `docker inspect <container>` on swarm is using a proxy so you already see `Networks` in docker inspect but I think we should still merge this PR so we have the networks in memory.

abronan:
LGTM

dongluochen:
LGTM.

Issue #1738, opened 2016-02-02
swarm doesn't support docker network ls type filter


dongluochen:
Docker [17782](https://github.com/docker/docker/pull/17782) added --filter flag to network ls to hide predefined networks. Swarm doesn't support it yet. 

```
dchen@vm4:~$ ./docker -H vm2:2372 network ls --filter type=custom
NETWORK ID          NAME                  DRIVER
7c4e189ea414        vm3/bridge            bridge              
40a16c3f8697        vm3/docker_gwbridge   bridge              
9aa678328195        vm3/none              null                
e8742e22077d        vm3/host              host                
12510c2d7c85        help                  overlay             
1b82f9e0ff53        newnet                overlay             
744edb5a2c4e        brNetForInspect       overlay             
d655fe013cf5        devnet                overlay           

dchen@vm4:~$ ./docker -H vm2:2372 network ls --filter type=builtin
NETWORK ID          NAME                  DRIVER
e8742e22077d        vm3/host              host                
12510c2d7c85        help                  overlay             
1b82f9e0ff53        newnet                overlay             
744edb5a2c4e        brNetForInspect       overlay             
d655fe013cf5        devnet                overlay             
7c4e189ea414        vm3/bridge            bridge              
40a16c3f8697        vm3/docker_gwbridge   bridge              
9aa678328195        vm3/none              null          
```

PR #1739, opened 2016-02-02
support docker network ls --filter type=XXX

vieux:
Fix https://github.com/docker/swarm/issues/1738

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -14,6 +14,10 @@ type Network struct {
 	Engine *Engine
 }
 
+func (network *Network) isPreDefined() bool {
+	return (network.Name == "none" || network.Name == "host" || network.Name == "bridge")
+}
+
 // Networks represents an array of networks
 type Networks []*Network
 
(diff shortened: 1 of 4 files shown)


vieux:
ping @dongluochen @abronan

abronan:
LGTM

@vieux I guess we're ok with `engine-stable` test failing for now until we update after the release right?

vieux:
@abronan I'm going to update the PR to skip the test for engine < 1.10

cluster/network.go
```diff
@@ -14,6 +14,10 @@ type Network struct {
 	Engine *Engine
 }
 
+func (network *Network) isPreDefined() bool {
+	return (network.Name == "none" || network.Name == "host" || network.Name == "bridge")
```
dongluochen: Shouldn't it be `network.Driver`? 
```
dchen@vm4:~$ ./docker -H vm2:2372 network ls
NETWORK ID          NAME                  DRIVER
d655fe013cf5        devnet                overlay             
12510c2d7c85        help                  overlay             
40a16c3f8697        vm3/docker_gwbridge   bridge              
9722189bb82b        vm3/none              null                
abf0616c3dd0        vm3/bridge            bridge              
```

vieux:
@dongluochen  should be good to merge

cluster/network.go
vieux: @dongluochen no, you can create a network with the bridge driver, as you created it, it's not predefined

dongluochen:
LGTM.

PR #1741, opened 2016-02-03
Disallow outdated API versions

### LINKED ISSUE #1733: Swarm API versions need refining
I see ~~two issues~~ one issue with Swarm API versions
1. Currently, the Swarm API allows arbitrary version numbers. This doesn't make sense since Engine sets a lower limit (https://github.com/docker/docker/blob/012a3b6e74d753e1b0560d8f9b085b0fa0a80c0f/api/common.go#L19). I think Swarm should stay in sync with that. It makes sense to prevent issues such as #1727 caused by too many differences across API versions.
2. ~~A static API Version is returned for the `/version` endpoint in `api/handlers.go`. Now that we have the API version in the context, we should return that instead, while setting a static default, also like Engine does it.~~

@vieux @dongluochen @abronan thoughts?

nishanttotla:
Fixes #1733. Disallows outdated API versions using a `MINAPIVERSION` field that tracks the oldest engine version supported (in this case 1.18 corresponding to engine version 1.6.x). Also bumping up APIVERSION to 1.22 for the 1.1.0 release.

### DIFF
--- a/api/primary.go
+++ b/api/primary.go
@@ -2,11 +2,13 @@ package api
 
 import (
 	"crypto/tls"
+	"fmt"
 	"net/http"
 
 	"net/http/pprof"
 
 	log "github.com/Sirupsen/logrus"
+	versionpkg "github.com/docker/docker/pkg/version"
 	"github.com/docker/swarm/cluster"
 	"github.com/gorilla/mux"
 )
@@ -137,6 +139,15 @@ func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHan
 	return r
 }
 
+func versionCheckMiddleware(c *context, w http.ResponseWriter, r *http.Request, handlerFunc handler) {
+	if c.apiVersion != "" && versionpkg.Version(c.apiVersion).LessThan(MINAPIVERSION) {
+		httpError(w, fmt.Sprintf("API Version %s is too old. Supported versions > %s.", c.apiVersion, MINAPIVERSION), http.StatusBadRequest)
+		return
+	}
+	// API version passes checks, now call the handler
+	handlerFunc(c, w, r)
+}
+
 func setupPrimaryRouter(r *mux.Router, context *context, enableCors bool) {
 	for method, mappings := range routes {
 		for route, fct := range mappings {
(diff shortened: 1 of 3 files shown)


nishanttotla:
ping @vieux @dongluochen

api/primary.go
```diff
@@ -137,6 +139,15 @@ func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHan
 	return r
 }
 
+func versionCheckMiddleware(c *context, w http.ResponseWriter, r *http.Request, handlerFunc handler) {
+	if c.apiVersion != "" && versionpkg.Version(c.apiVersion).LessThan(MINAPIVERSION) {
+		httpError(w, fmt.Sprintf("API Version %s is too old. Supported versions > %s.", c.apiVersion, MINAPIVERSION), http.StatusBadRequest)
```
vieux: `>=` ?

nishanttotla: Fixed, thanks!

vieux:
LGTM we already say in the doc that swarm requires engine 1.6.0 or higher, let's enforce it in the API too.

vieux:
ping @dongluochen

dongluochen:
I think we should update samalba/dockerclient APIVersion first since it's still at 1.15. Later we can enforce it. 
```
dchen@vm2:~/go/src/github.com/samalba/dockerclient$ grep -irnw . -e "apiversion" --include=\*.go
./dockerclient.go:20:	APIVersion = "v1.15"
```

vieux:
@dongluochen you are right, I moved it to the 1.2.0 milestone

nishanttotla:
@dongluochen thanks for pointing that out. I'm looking into `samalba/dockerclient`, and it seems like it's as simple as bumping up `APIVersion`. Whether or not this PR is merged before the release, shouldn't we have bump version in `dockerclient` too?

dongluochen:
@nishanttotla Yes we should bump version in `samalba/dockerclient` first, in case external users are using it. After some time we can set appropriate minimum API version in swarm.

nishanttotla:
Adding dependency #1879

nishanttotla:
Closing this PR, it's outdated.

PR #1756, opened 2016-02-04
remove goveralls

vieux:
once again goveralls is broken, and it breaks our travis.
we don't really use it, let's remove it


vieux:
ping @dongluochen @abronan

abronan:
LGTM

dongluochen:
LGTM

Issue #1749, opened 2016-02-03
Do not really on mux vars for container, network, volume names


doronp:
e.g.

https://github.com/docker/swarm/blob/master/api/handlers.go#L452

This may prevent plugins from modifying the respected resources names before being processed.
Use the request instead.

nishanttotla:
@doronp Can you clarify the issue a little further? What exactly are you trying to modify?

doronp:
@nishanttotla 
Any plugin which changes anything is the URL will be able to modify the request but not the mux vars as it runs in another process.

doronp:
See PR #1770

PR #1770, opened 2016-02-08
Do not use mux for request data (issue #1749)

doronp:
Signed-off-by: Doron Podoleanu doronp@il.ibm.com


doronp:
See Issue #1749

PR #1774, opened 2016-02-09
Extract leadership package and update Godeps

abronan:
This is mostly cosmetic but the `leadership` package was always considered as an extension to `docker/libkv` which is in its own repository now. It makes more sense to extract it and consider it as a library.

@docker/swarm-maintainers WDYT?

(would be nice to check if it didn't messed up with Godeps somehow, I had some issues adding the new dependency)


abronan:
@vieux Fixed the go version in `Godeps.json`, thanks for finding this :)

vieux:
LGTM thanks @abronan

dongluochen:
LGTM.

Issue #1752, opened 2016-02-03
Swarm introduces linear performance degradation as #networks or #containers increases


ezrasilvera:
It seems that Swarm introduces linear performance degradation as the number of networks and/or the  number of containers increases. I was focusing mainly on  starting a container and creating a network, but this issue affect many other commands like network delete, container delete, etc..

I made some profiling analysis and I believe that the source of this performance problem is the usage of RefreshNetworks() and RefreshContainers().

I think there are few issues with the current design/implementation:
Duplication - when we initiate "create container"  for example- Swarm explicitly calls refreshNetwork, however when the create container event will be received Swarm will initiate another refresh on the specific engine
Redundancy - When creating a container or a network Swarm calls refreshNetwork which goes over all engines and all networks on each engine. We we cerate a resources (container or network) we can actually do a focused discovery (e.g., during createNetwork send a listNetworks only for the id of the newly created network) 

As Docker now supports overlay networks the situation becomes much worse as each engine may see many networks!

I also made some tests with eliminating some of the  redundant refresh calls and indeed most of the Swarm overhead was removed 

A real fix would probably require some more investigation as currently the relevant functions are also used in situations in which you do want to discover everything (like swarm restart)
![swarm-nw](https://cloud.githubusercontent.com/assets/13567561/12795573/53ebb1da-cac3-11e5-9dc1-1a47348f31af.png)

ezrasilvera:
@abronan I'm not sure this is a network specific issue. I think it's a wider issue related to the overall Swarm "discovery" mechanism. The ability to create user-defined network just brought the issue to the surface as the number of networks increases. 
However we also see similar behavior related to # of containers

vieux:
@ezrasilvera thanks for the comprehensive report

We are definitely concerned about this and we want to improve performances.
I agree with you `refreshNetworks()` is the biggest issue, networks are the only place where you can do an API call on one engine and it's going to affect others engines, without swarm knowing about it.

That's why we are currently doing a refresh on every engine and that's really not ideal.

We are about to release 1.1.0 (tomorrow) but we are going to work on this very soon and release a `1.1.1` as soon as we reach significant improvement.

Do you mind sharing with us the branch you used for the `reduced refreshNW` tests ?

ezrasilvera:
@vieux sure. will do. 
[...]

PR #1777, opened 2016-02-09
remove refresh networks from cluster to improve perfs

vieux:
Related to https://github.com/docker/swarm/issues/1752

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -451,7 +451,10 @@ func (e *Engine) DeleteNetwork(network *Network) {
 // AddNetwork adds a network to the internal engine state.
 func (e *Engine) AddNetwork(network *Network) {
 	e.Lock()
-	e.networks[network.ID] = network
+	e.networks[network.ID] = &Network{
+		NetworkResource: network.NetworkResource,
+		Engine:          e,
+	}
 	e.Unlock()
 }
 
@@ -924,22 +927,41 @@ func (e *Engine) String() string {
 
 func (e *Engine) handler(ev *dockerclient.Event, _ chan error, args ...interface{}) {
 	// Something changed - refresh our internal state.
-	switch ev.Status {
-	case "pull", "untag", "delete", "commit":
-		// These events refer to images so there's no need to update
-		// containers.
+
+	switch ev.Type {
+	case "network":
+		e.RefreshNetworks()
+	case "volume":
+		e.RefreshVolumes()
+	case "image":
 		e.RefreshImages()
-	case "die", "kill", "oom", "pause", "start", "stop", "unpause", "rename":
-		// If the container state changes, we have to do an inspect in
-		// order to update container.Info and get the new NetworkSettings.
-		e.refreshContainer(ev.ID, true)
-		e.RefreshVolumes()
-		e.RefreshNetworks()
-	default:
-		// Otherwise, do a "soft" refresh of the container.
-		e.refreshContainer(ev.ID, false)
-		e.RefreshVolumes()
-		e.RefreshNetworks()
+	case "container":
+		switch ev.Action {
+		case "die", "kill", "oom", "pause", "start", "stop", "unpause", "rename":
+			e.refreshContainer(ev.ID, true)
+		default:
+			e.refreshContainer(ev.ID, false)
+		}
+	case "":
+		// docker < 1.10
+		switch ev.Status {
+		case "pull", "untag", "delete", "commit":
+			// These events refer to images so there's no need to update
+			// containers.
+			e.RefreshImages()
+		case "die", "kill", "oom", "pause", "start", "stop", "unpause", "rename":
+			// If the container state changes, we have to do an inspect in
+			// order to update container.Info and get the new NetworkSettings.
+			e.refreshContainer(ev.ID, true)
+			e.RefreshVolumes()
+			e.RefreshNetworks()
+		default:
+			// Otherwise, do a "soft" refresh of the container.
+			e.refreshContainer(ev.ID, false)
+			e.RefreshVolumes()
+			e.RefreshNetworks()
+		}
+
 	}
 
 	// If there is no event handler registered, abort right now.
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -427,12 +427,32 @@ func (e *Engine) RemoveImage(image *Image, name string, force bool) ([]*dockercl
 
 }
 
-// RemoveNetwork deletes a network from the engine.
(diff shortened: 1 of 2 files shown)


cluster/swarm/cluster.go
```diff
@@ -216,7 +216,12 @@ func (c *Cluster) RemoveContainer(container *cluster.Container, force, volumes b
 // RemoveNetwork removes a network from the cluster
 func (c *Cluster) RemoveNetwork(network *cluster.Network) error {
 	err := network.Engine.RemoveNetwork(network)
-	c.refreshNetworks()
+	if err == nil {
+
```
nishanttotla: Is this condition meant to be empty?

amitshukla: Should this be "err != nil" + error handling?

ezrasilvera:
@vieux I just started to look into this as well . Will wait to see your changes. I think that another change should be to update the event Handler to handle more focused events and not "refresh"  the network on general events (e.g., don't refresh the NW list for create container event and for connect container just update this specific NWID and CONT_ID )

cluster/swarm/cluster.go
```diff
@@ -216,7 +216,12 @@ func (c *Cluster) RemoveContainer(container *cluster.Container, force, volumes b
...
+	}
+	for _, engine := range c.engines {
+		engine.DeleteNetwork(network)
```
ezrasilvera: Maybe it would be better to call the engine.RemoveNetwork function so we have a single entry point for removing networks and leave the DeleteNetwork as private one?

vieux: if we do that it's like before, it's going to hit every engine via the network and kill our perfs

cluster/swarm/cluster.go
vieux: @nishanttotla sorry, the for should be in the if, thanks :)

cluster/swarm/cluster.go
vieux: but I agree, having 2 functions with almost the same name isn't great.

vieux:
@ezrasilvera I agree about your event comment.

vieux:
@dongluochen @nishanttotla @ezrasilvera can you please review again ?

Thanks.

ezrasilvera:
@vieux maybe I'm missing something here - you removed the refresh on the CreateNetwork as well - how is the network going to be added to the structures of the other engines in case of overlay networks ?
Don't you need a similar loop as in the delete ?

vieux:
@ezrasilvera right for some reasons I thought the event would be enough, but libnetwork doesn't propagate events. I added it, I'm now going to look into the events

cluster/swarm/cluster.go
```diff
@@ -477,7 +469,12 @@ func (c *Cluster) CreateNetwork(request *dockerclient.NetworkCreate) (response *
 	}
 	if nodes != nil {
 		resp, err := c.engines[nodes[0].ID].CreateNetwork(request)
-		c.refreshNetworks()
+		if err == nil {
+			network := c.engines[nodes[0].ID].Networks().Get(resp.ID)
+			for _, engine := range c.engines {
+				engine.AddNetwork(network)
```
dongluochen: We should only do this for overlay/global network.

cluster/swarm/cluster.go
```diff
@@ -216,7 +216,11 @@ func (c *Cluster) RemoveContainer(container *cluster.Container, force, volumes b
 // RemoveNetwork removes a network from the cluster
 func (c *Cluster) RemoveNetwork(network *cluster.Network) error {
 	err := network.Engine.RemoveNetwork(network)
-	c.refreshNetworks()
+	if err == nil {
+		for _, engine := range c.engines {
+			engine.DeleteNetwork(network)
```
dongluochen: this should only for overlay network.

cluster/swarm/cluster.go
vieux: thanks @dongluochen good catch!

cluster/swarm/cluster.go
vieux: thanks @dongluochen good catch!

vieux:
@dongluochen please take another look, I fixed your issues and also fixed `AddNetwork`

dongluochen:
LGTM

vieux:
unrelated failure ping @ezrasilvera

ezrasilvera:
In cluster.CreateNetwork the loop for engine.AddNetwork  should exclude c.engines[nodes[0].ID] 
On this node you explicitly call engine.CreateNetwork  which in  turns create the whole network array on that specific engine (by engine.refreshNetwork) 
Looks minor but you might delay the whole loop due to the lock on that engine..

cluster/engine.go
```diff
@@ -904,22 +927,41 @@ func (e *Engine) String() string {
...
-		// Otherwise, do a "soft" refresh of the container.
-		e.refreshContainer(ev.ID, false)
+	case "volume":
 		e.RefreshVolumes()
-		e.RefreshNetworks()
+	case "image":
+		e.RefreshImages()
+	case "container":
+		switch ev.Action {
+		case "die", "kill", "oom", "pause", "start", "stop", "unpause", "rename":
```
ezrasilvera: when the container is removed I think you do need to update the NW as the NetworkResource holds reference to the connected containers. So we probably should separate those cases

vieux: When you remove a container, docker will either stop or kill it before and it will trigger a `network disconnect` event.
So I don't think a `RefreshNetworks` is needed.

ezrasilvera: I agree it's a minor thing and probably low probability, but you do have a potential race condition there especially if you take into account possible even loosing or delays. 
I'm fine with either ways..

vieux: But if we follow your comment, shouldn't also do in for:

start: (maybe you will miss the network connect)
stop: (maybe you will miss the network disconnect)
kill: (maybe you will miss the network disconnect)

no ?

ezrasilvera: Exactly - eventually we would need to treat each event separately. We can go with your change now and later on come back to fine tune the event handler. In any case we would probably need to do the same thing for refreshVolumes() sometime ...

vieux: We have the refresh loop ad fixed interval exactly for this case: if we miss an event.
I agree let's fine tune later.

Are you ok to merge this one ?

ezrasilvera:
LGTM

abronan:
LGTM

PR #1789, opened 2016-02-10
add Server Version to docker info

vieux:


vieux:
ping @schmunk42

schmunk42:
Nice :+1: I'd also like to see the version of the swarm image, if possible.

Sure, I can filter it with something like: `docker ps -a --filter name=swarm` on the swarm - but having it in `docker info` would also not be wrong.

vieux:
@schmunk42 today you can use `docker version` :)

schmunk42:
Thanks :D never noticed that.

nishanttotla:
LGTM

Issue #1549, opened 2015-12-16
Proposal: Refactor HTTP client used by swarm


dongluochen:
Swarm manager has 3 different ways to send a request to a node, dockerclient, httpclient, and `proxy` mode thru httpclient (see cluster/engine.go and api/handlers.go). The connection between swarm manger and node is important for functionality as well as performance. It's hard to maintain or improve 3 connection methods. For example, if we want to change TCP settings for client it'll be in different places. PR #1517 has difficulty to monitor connection failure from swarm manager to node. We want to merge the clients to one, and implement connection setting/monitoring in this data structure.

This can be associated with improvement on node management #1486.

nishanttotla:
@dongluochen did you make progress on this?

dongluochen:
@nishanttotla I haven't got time to work on it yet.

PR #1791, opened 2016-02-10
use same HTTP client as dockerengine

vieux:
This should reduce the number of open connections

Close https://github.com/docker/swarm/issues/1549

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -124,9 +124,9 @@ func getImages(c *context, w http.ResponseWriter, r *http.Request) {
 	names := r.Form["names"]
 
 	// Create a map of engine address to the list of images it holds.
-	engineImages := make(map[string][]*cluster.Image)
+	engineImages := make(map[*cluster.Engine][]*cluster.Image)
 	for _, image := range c.cluster.Images() {
-		engineImages[image.Engine.Addr] = append(engineImages[image.Engine.Addr], image)
+		engineImages[image.Engine] = append(engineImages[image.Engine], image)
 	}
 
 	// Look for an engine that has all the images we need.
@@ -146,7 +146,7 @@ func getImages(c *context, w http.ResponseWriter, r *http.Request) {
 
 		// If the engine has all images, stop our search here.
 		if matchedImages == len(names) {
-			proxy(c.tlsConfig, engine, w, r)
+			proxy(engine, w, r)
 			return
 		}
 	}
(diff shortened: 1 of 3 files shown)


vieux:
ping @dongluochen

api/handlers.go
```diff
@@ -130,7 +130,7 @@ func getImages(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Look for an engine that has all the images we need.
-	for engine, images := range engineImages {
+	for _, images := range engineImages {
```
MHBauer: If we're not going to use the first argument, can we avoid constructing it up above on L#127? Or maybe put `image.Engine` in there as the key.

api/utils.go
```diff
@@ -102,14 +96,14 @@ func closeIdleConnections(client *http.Client) {
...
 }
 
-func proxyAsync(tlsConfig *tls.Config, addr string, w http.ResponseWriter, r *http.Request, callback func(*http.Response)) error {
-	// Use a new client for each request
-	client, scheme := newClientAndScheme(tlsConfig)
+func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request, callback func(*http.Response)) error {
 	// RequestURI may not be sent to client
 	r.RequestURI = ""
 
+	client, scheme := engine.HTTPClientAndScheme()
```
dongluochen: `client` could be nil in your definition. Should validate it before usage.

api/handlers.go
```diff
@@ -462,7 +462,7 @@ func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	client, scheme := newClientAndScheme(c.tlsConfig)
+	client, scheme := container.Engine.HTTPClientAndScheme()
```
dongluochen: Need to validate `client`.

nishanttotla:
@vieux is this related to #1549?

vieux:
@nishanttotla yes !

MHBauer:
Just to be clear, this makes it use the `samalba/dockerclient` instead of a raw golang http client. Right?

vieux:
@MHBauer `samalba/dockerclient` uses the a  raw golang http client

Before this pr we were creating a new  raw golang http client for every proxy request,
now we simply reuse the same one as `dockerlient`, so we can reuse idle connection.

api/handlers.go
vieux: added

api/utils.go
vieux: added

dongluochen:
LGTM

api/handlers.go
```diff
@@ -462,7 +462,11 @@ func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
...
+	if client == nil {
+		httpError(w, "cannot connect to docker engine", http.StatusInternalServerError)
```
nishanttotla: Small detail - is this the same error message as returned by engine on failed connection?

api/handlers.go
```diff
@@ -767,7 +771,11 @@ func postContainersExec(c *context, w http.ResponseWriter, r *http.Request) {
...
```
nishanttotla: same here

api/handlers.go
vieux: probably not, do you think it should ?

nishanttotla: I suppose it's not an issue. Perhaps just using caps cannot -> Cannot would be nice.

abronan:
LGTM

PR #1793, opened 2016-02-11
fix flacky before tests

vieux:
we shouldn't use `c<number>` as name, since it's a valid start of container id.

In this case, if c1 or c1 start with c3 in their ID, the test will fail


MHBauer:
make sense to me. I can believe that a container ID would start with C1.
LGTM

dongluochen:
LGTM

PR #1796, opened 2016-02-11
Print all constraints when container can't be scheduled

### LINKED ISSUE #1785: Swarm doesn't list all conflicting constraints when it fails to place a container
## What I do

```
docker run -d --name web -p 80:80 nginx
docker run -d --volumes-from web -p 80:80 nginx
```

## What I expect to see

Something telling me that no node can satisfy both constraints "public port 80" and "volumes from web".

## What I see instead

Only the `--volumes-from` constraint is indicated, not the port constraint.
```
docker: Error response from daemon: Unable to find a node fulfilling all dependencies: --volumes-from=web.
```

## Extra notes

FWIW, this is a frequent scenario when mapping host ports with Compose+Swarm (see docker/compose#2866).

If I allocate port 80 on all nodes, then I see the following message:

```
docker: Error response from daemon: unable to find a node with port 80 available.
```

(Which is totally OK because if there are 0 node able to satisfy a single constraint, it's safe to assume that this constraint is the culprit ;-) )

Just for fun, I tried with an affinity constraint, and then I get:

```
docker: Error response from daemon: unable to find a node that satisfies container==6ab2dfe36615ae786ef3fc35d641a260e3ea9663d6e69c5b70ce0ca6cb373c02.
```

I don't know how the constraint engine works, but maybe it could report impossible constraints (for which there are zero node), then conflicting constraints (i.e. when the intersection of nodes for 2 given constraints is empty).

nishanttotla:
Fixes #1785.

Special case when no healthy node is found, in which case nothing else is printed. The logic is in `scheduler/filter/filter.go`

### DIFF
--- a/scheduler/filter/affinity.go
+++ b/scheduler/filter/affinity.go
@@ -77,8 +77,8 @@ func (f *AffinityFilter) Filter(config *cluster.ContainerConfig, nodes []*node.N
 	return nodes, nil
 }
 
-// Get a list of the affinities found in the container config.
-func (f *AffinityFilter) GetAllFilters(config *cluster.ContainerConfig) ([]string, error) {
+// GetFilters returns a list of the affinities found in the container config.
+func (f *AffinityFilter) GetFilters(config *cluster.ContainerConfig) ([]string, error) {
 	allAffinities := []string{}
 	affinities, err := parseExprs(config.Affinities())
 	if err != nil {
--- a/scheduler/filter/affinity.go
+++ b/scheduler/filter/affinity.go
@@ -76,3 +76,16 @@ func (f *AffinityFilter) Filter(config *cluster.ContainerConfig, nodes []*node.N
 
 	return nodes, nil
 }
+
+// Get a list of the affinities found in the container config.
+func (f *AffinityFilter) GetAllFilters(config *cluster.ContainerConfig) ([]string, error) {
+	allAffinities := []string{}
+	affinities, err := parseExprs(config.Affinities())
+	if err != nil {
+		return nil, err
+	}
+	for _, affinity := range affinities {
+		allAffinities = append(allAffinities, fmt.Sprintf("%s%s%s (soft=%t)", affinity.key, OPERATORS[affinity.operator], affinity.value, affinity.isSoft))
+	}
+	return allAffinities, nil
+}
--- a/scheduler/filter/filter.go
+++ b/scheduler/filter/filter.go
@@ -2,6 +2,7 @@ package filter
 
 import (
 	"errors"
+	"fmt"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/swarm/cluster"
@@ -16,7 +17,7 @@ type Filter interface {
 	Filter(*cluster.ContainerConfig, []*node.Node, bool) ([]*node.Node, error)
 
 	// Return a list of constraints/filters provided
-	GetAllFilters(*cluster.ContainerConfig) ([]string, error)
+	GetFilters(*cluster.ContainerConfig) ([]string, error)
 }
 
 var (
@@ -66,10 +67,29 @@ func ApplyFilters(filters []Filter, config *cluster.ContainerConfig, nodes []*no
 	for _, filter := range filters {
 		candidates, err = filter.Filter(config, candidates, soft)
 		if err != nil {
-			return nil, err
+			// special case for when no healthy nodes are found
+			if filter.Name() == "health" {
+				return nil, err
+			}
+			return nil, fmt.Errorf("Unable to find a node that satisfies the following conditions %s", listAllFilters(filters, config, filter.Name()))
 		}
 	}
 	return candidates, nil
+}
+
+// listAllFilters creates a string containing all applied filters
+func listAllFilters(filters []Filter, config *cluster.ContainerConfig, lastFilter string) string {
+	allFilters := ""
+	for _, filter := range filters {
+		list, err := filter.GetFilters(config)
+		if err == nil && len(list) > 0 {
+			allFilters = fmt.Sprintf("%s\n%v", allFilters, list)
+		}
+		if filter.Name() == lastFilter {
+			return allFilters
+		}
+	}
+	return allFilters
 }
 
 // List returns the names of all the available filters
--- a/scheduler/filter/filter.go
+++ b/scheduler/filter/filter.go
@@ -14,6 +14,9 @@ type Filter interface {
 
 	// Return a subset of nodes that were accepted by the filtering policy.
 	Filter(*cluster.ContainerConfig, []*node.Node, bool) ([]*node.Node, error)
+
+	// Return a list of constraints/filters provided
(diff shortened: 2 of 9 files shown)


nishanttotla:
ping @dongluochen @jpetazzo @dnephin

dnephin:
Nice, this looks like it solves the problem. Thanks!

test/integration/build_with_filters.bats
```diff
@@ -17,7 +17,9 @@ function teardown() {
 
 	run docker_swarm build --build-arg="constraint:node==node-9" $TESTDATA/build
 	[ "$status" -eq 1 ]
-	[[ "$output" == *"Error response from daemon: unable to find a node that satisfies node==node-9"* ]]
+	echo "THE OUTPUT IS $output"
```
vieux: leftover debug

test/integration/port-filters.bats
```diff
@@ -40,7 +41,9 @@ function teardown() {
 	# When trying to start the 3rd one, it should be error finding port 80.
 	run docker_swarm run --expose=80 -p 80:80 busybox echo 3
 	[ "$status" -ne 0 ]
-	[[ "${lines[0]}" == *"unable to find a node with port 80 available"* ]]
+	echo "THE OUTPUT IS $output"
```
vieux: same here

scheduler/filter/filter.go
```diff
@@ -63,12 +67,28 @@ func ApplyFilters(filters []Filter, config *cluster.ContainerConfig, nodes []*no
 	for _, filter := range filters {
 		candidates, err = filter.Filter(config, candidates, soft)
 		if err != nil {
-			return nil, err
+			// special case for when no healthy nodes are found
+			if filter.Name() == "health" {
+				return nil, err
+			}
+			return nil, fmt.Errorf("Unable to find a node that satisfies the following constraints %s", listAllFilters(filters, config))
```
vieux: can we say conditions ? if we use the same terms as the code, they aren't all constraints

test/integration/build_with_filters.bats
nishanttotla: oh no, sorry about that. I'll fix it.

amitshukla:
@nishanttotla @vieux With this change, we will have the following problem:
1. We list all the conditions that the user specified to schedule a container
2. We don't list the specific condition check that failed

Is this worse than the current error message as the user now has no knowledge of the specific check that failed? Perhaps printing:

> Condition 1
> Condition 2
> Condition 3 [Check failed]
> Condition 4

Will be more informational

nishanttotla:
@amitshukla that is correct, but given that our checks run in a batch, even right now the error message that gets printed doesn't necessarily tell you if that specific check would have failed on its own.

What you're suggesting
> Condition 1
> Condition 2
> Condition 3 [Check failed]
> Condition 4

would make sense if it conveyed that Condition 3 fails on its own, To do that, we'd have to run the check on all nodes as opposed to those filtered from checking Conditions 1 and 2, potentially making the overall operation more expensive.

scheduler/filter/affinity.go
```diff
@@ -76,3 +76,16 @@ func (f *AffinityFilter) Filter(config *cluster.ContainerConfig, nodes []*node.N
 
 	return nodes, nil
 }
+
+// Get a list of the affinities found in the container config.
+func (f *AffinityFilter) GetAllFilters(config *cluster.ContainerConfig) ([]string, error) {
```
dongluochen:
```
$ fgt golint ./...
scheduler/filter/affinity.go:80:1: comment on exported method AffinityFilter.GetAllFilters should be of the form "GetAllFilters ..."
scheduler/filter/constraint.go:55:1: comment on exported method ConstraintFilter.GetAllFilters should be of the form "GetAllFilters ..."
scheduler/filter/dependency.go:59:1: comment on exported method DependencyFilter.GetAllFilters should be of the form "GetAllFilters ..."
scheduler/filter/health.go:40:1: comment on exported method HealthFilter.GetAllFilters should be of the form "GetAllFilters ..."
scheduler/filter/port.go:122:1: comment on exported method PortFilter.GetAllFilters should be of the form "GetAllFilters ..."
The command "fgt golint ./..." exited with 1.
```

nishanttotla: Thanks, fixing it.

nishanttotla:
Made a slight update, I think now we have a better solution - now we only print a subset of failing constraints. Out of the 5, the first n constraints that fail as a combination are reported, the rest ignored.

For example in a Swarm with two nodes, assume an `nginx` container called `web` is running on node 1, attached to port 80 on the host.

Then, the outputs will look like the following
```
$ docker -H <manager-ip>:<manager-port> run -d --volumes-from web -p 80:80 nginx
Error response from daemon: Unable to find a node that satisfies the following conditions
[port 80 (Bridge mode)]
[--volumes-from=web]
```
This is as expected, but if we provide an additional image affinity constraint, the output remains the same, giving strictly more information to help debug
```
$ docker -H <manager-ip>:<manager-port> run -d --volumes-from web -p 80:80 -e affinity:image==nginx nginx
Error response from daemon: Unable to find a node that satisfies the following conditions
[port 80 (Bridge mode)]
[--volumes-from=web]
```

Thanks @amitshukla for the discussion on this!

scheduler/filter/filter.go
```diff
@@ -14,6 +15,9 @@ type Filter interface {
 
 	// Return a subset of nodes that were accepted by the filtering policy.
 	Filter(*cluster.ContainerConfig, []*node.Node, bool) ([]*node.Node, error)
+
+	// Return a list of constraints/filters provided
+	GetAllFilters(*cluster.ContainerConfig) ([]string, error)
```
dongluochen: I suggest changing `GetAllFilters` to `GetFilters`.

nishanttotla: Alright, I'll make the change. Any particular reason for preferring one over the other?

dongluochen: I feel `All` is not necessary. Another reason is you are calling `GetAllFilters` in `ApplyFilters` and `listAllFilters`. For me it reduces readability.

nishanttotla: Fair point, it's changed now!

dongluochen:
The test seems to fail repeatedly. Please take a look.

nishanttotla:
@dongluochen it seems like a `zk discovery: node removal` test, not sure that it's related.

dongluochen:
LGTM.

nishanttotla:
Ping @docker/core-swarm-maintainers

amitshukla:
@nishanttotla any update?

nishanttotla:
@amitshukla this PR is actually ready on my end, needs review.

nishanttotla:
@amitshukla @vieux @dongluochen I just rebased this PR. I think we can include it in 1.2.

vieux:
LGTM

dongluochen:
LGTM

PR #1797, opened 2016-02-11
Removing hard-coded ips from swarm w/ machine doc

nishanttotla:
Using `docker-machine ip` to get rid of hard-coded ip addresses.

### DIFF
--- a/docs/install-w-machine.md
+++ b/docs/install-w-machine.md
@@ -97,7 +97,7 @@ Here you use the discovery backend hosted on Docker Hub to create a unique disco
 
 Here, you connect to each of the hosts and create a Swarm manager or node.
 
-1. Get the IP addresses of the three VMs. For example:
+1. List the VMs to check that they're set up and running. For example:
 
         $ docker-machine ls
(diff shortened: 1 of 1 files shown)


nishanttotla:
ping @moxiegirl @dongluochen

MHBauer:
Do we still want the step ` 1. Get the IP addresses of the three VMs. For example:` around? Maybe add a sentence like L#147 within that step. Specifically the `You can show a docker-machine IPaddr by name` part.

nishanttotla:
@MHBauer you're right. Perhaps we could change it to something like "Check that the machines are running" or something. Since this tutorial is a very basic one, it would be good to leave some redundancy IMO.

MHBauer:
I do agree about using `docker-machine ip` though. If we have it, we should use it, and I think it is good to show that you can both list all of the machines, as well as get the ip of a specific machine by name. It should prevent confusion about mixing up ip addresses in use.

dongluochen:
LGTM. I think using `docker-machine ip` is better than hard coding IP address.

nishanttotla:
Updated with better descriptions.

vieux:
LGTM ping @moxiegirl

moxiegirl:
LGTM

PR #1798, opened 2016-02-11
Fix the swarm manager container creation (using docker-machine ssh)

jacobmarshall:
Modified the step (on the "Evaluate Swarm in a sandbox" page) which tells the user how to set up the Swarm container on the manager. Running the command from the docs as it stands causes the user's machine's `/var/lib/boot2docker` directory to be mounted into the Swarm manager container, instead of the manager VM's `/var/lib/boot2docker` directory.

I'm new to this, so if I've made a mistake, all good - my apologies.

### DIFF
--- a/docs/install-w-machine.md
+++ b/docs/install-w-machine.md
@@ -106,13 +106,13 @@ Here, you connect to each of the hosts and create a Swarm manager or node.
         manager   *        virtualbox   Running   tcp://192.168.99.100:2376           v1.9.1
 
 
-2. Your client should still be pointing to Docker Engine on `manager`. Use the following syntax to run a Swarm container as the primary Swarm manager on `manager`.
+2. Now we connect directly to the `manager` VM, using the `docker-machine ssh` command, to run a Swarm container as the primary Swarm manager.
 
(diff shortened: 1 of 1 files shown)


nishanttotla:
@jacobmarshall did you have issues running through the tutorial? Since `docker-machine` has limited access to the host (Mac or Windows) file systems (`/Users` directory), all other paths come from the VM's file system, as stated here: https://docs.docker.com/engine/userguide/containers/dockervolumes/

So I'm not sure that this change is needed.

cc @nathanleclaire can you confirm?

jacobmarshall:
If that is indeed the case, which I do not doubt, I can only assume my issue was to do with a change I made to the commands when I ran them. I did however run into an error at one point, which I will grab off the computer I ran the set up on later, which stated it was unable to load the certs (after doing a `docker logs` on the swarm manager container).

nishanttotla:
I just reconfirmed, and the current way of mounting works for me. I have a Gist of all commands if you want to run it quickly: https://gist.github.com/nishanttotla/e4b623ce47e944086ebc

jacobmarshall:
Cheers @nishanttotla, I appreciate your time.

PR #1801, opened 2016-02-11
Fix race condition in rescheduling test

dongluochen:
Container reschedule is done in 2 steps, container `create` and container `start`. If container `inspect` is issued randomly, the container status could be `created` or `running`. Current rescheduling tests may fail because they expect only `running` state. This change is to remove such uncertainty. 


abronan:
LGTM

nishanttotla:
LGMT

PR #1806, opened 2016-02-12
[WIP] proxy requests for build

vieux:
It fixes `FROM <private_image>` in build and make the code much cleaner

Replace #1556


vieux:
@docker/swarm-maintainers can you guys take a look to this one ? I don't get why it's falling on 1.9

vieux:
since it wasn't introduced in 1.1.0, let's move this one to 1.2.0

api/handlers.go
```diff
@@ -1066,6 +1068,59 @@ func proxyContainerAndForceRefresh(c *context, w http.ResponseWriter, r *http.Re
...
+	rp := regexp.MustCompile("buildargs=(.*)7D&")
+	r.URL.RawQuery = rp.ReplaceAllString(r.URL.RawQuery, "buildargs="+url.QueryEscape(string(raw))+"&")
+
+	engine, err := c.cluster.GetEngine(clusterConfig)
+	if err != nil {
+		httpError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	if engine == nil {
```
dongluochen: This shouldn't happen, and I think it bad practice if it does happen.

api/handlers.go
```diff
@@ -1066,6 +1068,59 @@ func proxyContainerAndForceRefresh(c *context, w http.ResponseWriter, r *http.Re
...
+		return
+	}
+
+	cb := func(resp *http.Response) {
+		// force fresh images
+		engine.RefreshImages()
+	}
+
+	err = proxyAsync(engine, w, r, cb)
+	err = proxy(engine, w, r)
```
dongluochen: I think these 2 lines need some clarification or comment.

Issue #1734, opened 2016-02-02
New label to limit the number of containers per node


xFragger:
just as discussed in #534 i would need to have a label (or something like this) to limit the number of containers per node. 
What about something like com.docker.containersmax oder com.docker.swarm.maxcontainers?
It should be absolutely CPU-/Memory-/Somethingelse-independent. 
When no node is found, the "error" should be in the same manner like the error, when no node with that port or constraint is found, something beginning with "unable to find a node ...".

I would help, but my go experience is like zero, maybe i've got the time end of next week, to help a bit with this

dongluochen:
@xFragger I'm not very sure about CPU/Memory/Somethingelse-independent. Container clustering is about resource utilization and your case seems to fit in that. I had a similar case and I used memory limit to do it. I can only run 4 instances of the following command to use up 4 GB memory on the node.  
```
dchen@vm4:~$ ./docker -H vm2:2372 run  -m 1G -d busybox sh
de884da33e794b3b507a9bac62a8650196e9cc83e4f048127fbcd06b3313e54c
dchen@vm4:~$ ./docker -H vm2:2372 info
Containers: 18
 Running: 0
 Paused: 0
 Stopped: 18
Images: 19
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 1
 vm3: xxx.xxx.xxx.xxx:4444
  └ Status: Healthy
  └ Containers: 18
  └ Reserved CPUs: 1 / 2
  └ Reserved Memory: 1 GiB / 4.053 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.19.0-25-generic, operatingsystem=Ubuntu 14.04.3 LTS, storagedriver=aufs
  └ Error: (none)
  └ UpdatedAt: 2016-02-03T21:44:12Z
Plugins: 
 Volume: 
 Network: 
Kernel Version: 3.19.0-25-generic
Operating System: linux
Architecture: amd64
CPUs: 2
Total Memory: 4.053 GiB
Name: vm2
Experimental: true
```

xFragger:
But with this, i limit the memory per container to 1GB, i don't want to do this. 

I'm sure, that that is exactly what i need: a max of n containers per node, controlled by a label at the node

xFragger:
Another more general solution could be to allow <, >, >= and <= as possible constraint expressions and bring in some "variables" to compare to

I would label all my instances with a maxContainerCount label und check at launch time by
[...]

PR #1809, opened 2016-02-12
Filter for containerslots-label

xFragger:
As proposed in #1734

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1076,6 +1076,7 @@ func proxyContainerAndForceRefresh(c *context, w http.ResponseWriter, r *http.Re
 	if err != nil {
 		if container == nil {
 			httpError(w, err.Error(), http.StatusNotFound)
+			return
 		}
 		httpError(w, err.Error(), http.StatusInternalServerError)
 		return


xFragger:
i'm confused about that failing build... 
```
The command "fgt golint ./..." exited with 1.
$ fgt find test/ -type f \( -name "*.sh" -or -name "*.bash" -or -name "*.bats" \) -exec grep -Hn -e "^ " {} \;
```
This exits with 0 at my local mac, so any suggestions how to fix this?
And any comments about the filter?

abronan:
@xFragger:

```
scheduler/filter/slots.go:11:2: exported var ErrNoNodeWithFreeSlotsAvailable should have comment or be unexported
```

`ErrNoNodeWithFreeSlotsAvailable` should have a comment beginning by:

```
// ErrNoNodeWithFreeSlotsAvailable <...>
```

xFragger:
So what is the next step here? Any suggestions? Anything i should change? Is there any hope, this will be adopted?

dongluochen:
@xFragger Thanks for your PR! Can you add integration test under test/integration? You can take a look at other examples like port-filters.bats. We would also like you to update documents but let's do that in a separate PR after this.

nishanttotla:
@xFragger Could you also please squash your commits into a single one?

Also, if this gets merged after #1796, you might have to write a little more code to add code to print all constraints :)

xFragger:
Sorry for my git-stupidness, but i'm not able to squash the commits at the moment... i think it's caused by the 2 merges from master... can someone help me out with that? 
About the docs, what speaks against using this PR for that?
Which docs(files) should i update?
Any idea, why travis fails?

test/integration/containerslots-filter.bats
```diff
@@ -0,0 +1,56 @@
...
+load helpers
+
+function teardown() {
+	swarm_manage_cleanup
+	stop_docker
+}
+
+@test "containerslots filter" {
+    export DOCKER_DAEMON_ADDITIONAL_PARAMETERS="--label containerslots=2"
+	start_docker_with_busybox 2
```
vieux: you can use `start_docker_with_busybox 2 --label containerslots=2` no need for `DOCKER_DAEMON_ADDITIONAL_PARAMETERS`

test/integration/containerslots-filter.bats
```diff
@@ -0,0 +1,56 @@
...
+    [ "$status" -eq 0 ]
+    run docker_swarm run -d -t busybox sh
+    [ "$status" -eq 0 ]
+    run docker_swarm run -d -t busybox sh
+    [ "$status" -eq 0 ]
+
+    # And the number of running containers should be 5.
+	run docker_swarm ps
+	[ "${#lines[@]}" -eq 6 ]
+}
```
vieux: please add a newline character

test/integration/containerslots-filter.bats
```diff
@@ -0,0 +1,56 @@
...
+    run docker_swarm run -d -t busybox sh
+    [ "$status" -eq 0 ]
+    run docker_swarm run -d -t busybox sh
+    [ "$status" -eq 0 ]
+    run docker_swarm run -d -t busybox sh
+    [ "$status" -eq 0 ]
+
+    # And the number of running containers should be 5.
+	run docker_swarm ps
+	[ "${#lines[@]}" -eq 6 ]
```
vieux: indentation seems wrong

test/integration/containerslots-filter.bats
```diff
@@ -0,0 +1,56 @@
...
+
+	# Use busybox to save image pulling time for integration test.
+	# Running the first 4 containers, it should be fine.
+	run docker_swarm run -d -t busybox sh
+	[ "$status" -eq 0 ]
+	run docker_swarm run -d -t busybox sh
+	[ "$status" -eq 0 ]
+	run docker_swarm run -d -t busybox sh
+    [ "$status" -eq 0 ]
+    run docker_swarm run -d -t busybox sh
```
vieux: indentation seems wrong
 Add a line note

nishanttotla:
@xFragger if you can't squash commits with `soft` reset, you might have to make a backup copy of the code, do a hard reset, then redo the changes. The current list of commits needs to cleaned up to keep the repo history clean.

Secondly, you will need to make appropriate changes since #1796 was merged. Let us know if you have questions :)

xFragger:
so i will make the changes for #1796, when you guys give the okay, i will hard reset and redo everything

nishanttotla:
@xFragger sure, whatever you're comfortable with. In general, to avoid too many commits, you can use `git commit --amend` to add changes to an existing commit, then force push using `git push -f` to reflect those changes remotely.

xFragger:
thought a while about it.... isn't my case similar to the "healthy-case" and an exception as well?
if it can't find a node without free slots, its the same as it can't found a node that is healthy.

As an alternative i could just return something like "free slot"

scheduler/filter/slots.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Filter is exported
+func (f *SlotsFilter) Filter(_ *cluster.ContainerConfig, nodes []*node.Node, _ bool) ([]*node.Node, error) {
+	result := []*node.Node{}
+
+	for _, node := range nodes {
+
+		if slotsString, ok := node.Labels["containerslots"]; ok {
+			slots, err := strconv.Atoi(slotsString) //if err => cannot cast to int, so no candidate for us
```
dongluochen: > //if err => cannot cast to int, so no candidate for us
              
I think it wrong to implicitly not using a node. It'd be better to fail it explicitly.

xFragger: I'm open to any ideas.... but i don't know, what i should do alternatively...
Return an error, to prevent starting a container at all?

dongluochen: I'm thinking about 2 different approaches.
- This label is invalid so it doesn't take effect. The node would be included as a candidate. You may add a error logging for the manager of this invalid label. It might generate too much log if we do it here though. 
- Do validation on this label and put the node to `unhealthy` state. Nodes in `unhealthy` state wouldn't take new containers. But `docker -H manager:port info` would show the node as `unhealthy`. So it's an explicit failure. 

What do you think?

xFragger: I would be okay with both approaches.... i don't think its too much logging, but the problem is: who reads the syslog?! i guess nearly nobody... 

The other approach sounds better too me, someone startet a docker daemon with a faulty label value, so it will never be healthy to start any containers on it. Will the user see, "why" it is unhealthy? There is some kind of error field in the info view, right?

dongluochen: For the second approach it's a little more complicated than I thought. Yes swarm info can show errors for a node. But `unhealthy` is designed for reachability failure, while `pending` is designed for node validation. See #1486. It looks like `pending` is a better state for this failure. But it was a one-way flow. Now if docker label gets updated on-the-fly the node will move back to `pending`. It may require some work.

xFragger: As i know, you can't update the label of a running daemon, can you?
So it won't mode back, right?

dongluochen: Docker 1.10 allows label update, see (https://blog.docker.com/2016/02/docker-1-10/).

> Daemon configuration file: It’s now possible to configure daemon options in a file and reload some of them without restarting the daemon so, for example, you can set new daemon labels and enable debug logging without restarting anything.

xFragger: Damn.... 
2 other ideas, that came...

1. what about not using a label, but a command line argument for the swarm agent that has to be int?
2. what about the first suggestion, to fail but log something, so if the user wonders why his containers won't start at node xy he looks into the log and sees what the reason is?

dongluochen: > what about not using a label, but a command line argument for the swarm agent that has to be int?

If you want to add this to Docker Daemon, I think label is the only option. You can control each docker engine's slot number. The other option is to add it to Swarm manager. The problem is that it'd apply to all docker engines equally, i.e., every node can host *k* containers. It's not good if some nodes have more resources or they run different jobs. 

> what about the first suggestion, to fail but log something, so if the user wonders why his containers won't start at node xy he looks into the log and sees what the reason is?

Again, logs are not obvious as user may not look at it. So I consider this as failing implicitly. My suggestion is to ignore an *invalid* label and let container run. It's consistent with how we deal with other labels.

xFragger: So "just" ignore the label, as it was not there, i would be okay with that.

The first idea was for the docker daemon, it was for the swarm agent, running on the nodes, can't they have a command line argument at start-time?

dongluochen: > The first idea was for the docker daemon, it was for the swarm agent, running on the nodes, can't they have a command line argument at start-time?

I do not follow. Swarm agent is docker daemon. Label is what you can put as user specification.

xFragger: You start one swarm master that controls your swarm and on every docker daemon you start one swarm-agent, don't you?
```docker run -d swarm join --advertise=<node_ip>:2375 consul://<consul_ip>:8500```
and my idea was to introduce another variable there (if possible)
```docker run -d swarm join --slots=3 --advertise=<node_ip>:2375 consul://<consul_ip>:8500 ```

Comment (dongluochen) [inline scheduler/filter/slots.go]: It's not a good idea to repurpose `join` to change a agent's configuration. Swarm may deprecate `join`.

Comment (xFragger) [inline scheduler/filter/slots.go]: So, to be sure, i'm going to"just" ignore the label, as it was not there, when its not castable to integer and then we could possibly merge this thing?

Comment (dongluochen) [inline scheduler/filter/slots.go]: I think that's a acceptable solution.

**Comment by xFragger** [comment]
One last problem about #1796. If i got it right, i just return something like "free slot". It more something like the health filter in my opinion, which is a special case. 

I think if all slots are full on all nodes, the message should be something like "Unable to find a node that satisfies the following conditions" linebreak "[free slots]", right?

**Comment by xFragger** [comment]
waiting for your feedback, will hard-reset+redo to clean up commit-history, when everything is okay for you

File: test/integration/containerslots-filter.bats [hunk 6]
```diff
@@ -0,0 +1,77 @@
...
+	[ "$status" -eq 0 ]
+	run docker_swarm run -d -t busybox sh
+	[ "$status" -eq 0 ]
+	run docker_swarm run -d -t busybox sh
+	[ "$status" -eq 0 ]
+
+	# And the number of running containers should be 5.
+	run docker_swarm ps
+	[ "${#lines[@]}" -eq 6 ]
+}
```
dongluochen: Format problem.

dongluochen:
> I think if all slots are full on all nodes, the message should be something like "Unable to find a node that satisfies the following conditions" linebreak "[free slots]", right?

Take a look at https://github.com/docker/swarm/pull/1796. Each filter can reduce the candidate set. We are printing the filters till the candidate set is empty.

test/integration/containerslots-filter.bats
xFragger: missing linebreak? or something else?

xFragger:
sorry, could you just help for that last little thing? what excatly should my filter return unter which circumstances?

test/integration/containerslots-filter.bats
dongluochen: On line 77 I see a red return at the end.

dongluochen:
You already implements the `GetFilters` function. If this is the function leading to empty candidate set, `free slots` would show up. 

```
func (f *SlotsFilter) GetFilters(config *cluster.ContainerConfig) ([]string, error) {
	return []string{"free slots"}, nil
}
```

Here is the example for port filter and how the test works.  
https://github.com/docker/swarm/blob/master/scheduler/filter/port.go#L123
https://github.com/docker/swarm/blob/master/test/integration/port-filters.bats#L45

xFragger:
Yeah i just did, but the question was, if this is okay for you and if we are ready, if i merge all changes to one commit?

dongluochen:
Yes please squash the changes.

PR #1824, opened 2016-02-17
Passing HostConfig for /start endpoint

### LINKED ISSUE #1781: Start container command not forwarding the HostConfig to the container
Client tools (Docker plugin for Jenkins in our case) that rely on passing (additional) HostConfig in the start container command do not work since 1.1.0. For example the port bindings are not passed and the container is inaccessible.

I believe the commit for 'Add support for container rescheduling on node failure.' (https://github.com/docker/swarm/commit/13f60212f57dae8ba480c28bf26261bd7331ea4) broke this flow.

From the Docker Remote API docs:
> 
Start a container:
For backwards compatibility, this endpoint accepts a HostConfig as JSON-encoded request body.
Reference: https://docs.docker.com/engine/reference/api/docker_remote_api_v1.22/#start-a-container

nishanttotla:
Fixes #1781. `HostConfig` is being deprecated for the `/start` endpoint, but we must support it until Engine does.


nishanttotla:
Ping @vieux @abronan @dongluochen some tests are failing and they're related, I'm not sure why. It would be really helpful if you could help review the code changes, since this issue is on the 1.1.1 Milestone.

api/handlers.go
```diff
@@ -747,6 +754,8 @@ func postContainersStart(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	container.Config.HostConfig = hostConfig
```
vieux: you are erasing the hostconfig given at create, i thik we need to merge them instead.

vieux: Or, as we did before, don't even save the config on start, just pass it via are to dockerclient, I think it makes sense

nishanttotla:
Closing PR since I made a git mess. Reopening a new one instead, sorry about that!

api/handlers.go
nishanttotla: Thanks @vieux, this makes sense. Sorry I have to close this PR due to some git issues, let's move the conversation over to #1829.

PR #1829, opened 2016-02-17
Passing HostConfig for /start endpoint

### LINKED ISSUE #1781: Start container command not forwarding the HostConfig to the container

nishanttotla:
Fixes #1781. `HostConfig` is being deprecated for the `/start` endpoint, but we must support it until Engine does.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -740,6 +740,26 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
 
 // POST /containers/{name:.*}/start
 func postContainersStart(c *context, w http.ResponseWriter, r *http.Request) {
+	hostConfig := &dockerclient.HostConfig{
+		MemorySwappiness: -1,
+	}
+
+	buf, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		httpError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	r.Body.Close()
+
+	if len(buf) == 0 {
+		hostConfig = nil
+	} else {
+		if err := json.Unmarshal(buf, hostConfig); err != nil {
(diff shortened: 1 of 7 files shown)


api/handlers.go
```diff
@@ -747,7 +754,7 @@ func postContainersStart(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := c.cluster.StartContainer(container); err != nil {
+	if err := c.cluster.StartContainer(container, &hostConfig); err != nil {
```
nishanttotla: @vieux how about something like this?

nishanttotla:
Tests are failing because of a single issue - `Error response from daemon: EOF` with error code 125 when running containers. In addition, I see `http: multiple response.WriteHeader calls` in the Swarm console. Pretty sure it's a small issue, so any help debugging is appreciated!

nishanttotla:
Ping @vieux @dongluochen

vieux:
LGTM

nishanttotla:
Test failure seems unrelated, also tests pass locally.

dongluochen:
LGTM.

PR #1835, opened 2016-02-18
Add support some run options

jimmyxian:
Depend: https://github.com/samalba/dockerclient/pull/228

Support following options:
+ --oom-score-adj
+ --tmpfs
+ --device-read-iops
+ --device-write-iops
+ --device-read-bps
+ --device-write-bps
+ --blkio-weight-device


abronan:
Sweet! Thanks @jimmyxian, seems like the test (`docker run with resources`) fails in the CI though. We should disable the tests for docker `1.9` as some of those flags are only supported since docker `1.10.x`.

It also fails for `1.10` though so I guess one of the parameter has a wrong argument.

jimmyxian:
@abronan Yeah, you are right. I will check that. :)

vieux:
@jimmyxian samalba/dockerclient#228 was merged :)

jimmyxian:
@vieux Thanks. :)
I will modify the integration test.

jimmyxian:
ping @abronan @vieux 
Done! 
BTW: There are something wrong with `--blkio-weight-device` integration test. I have removed it. :)

test/integration/api/run.bats
```diff
@@ -99,6 +99,35 @@ function teardown() {
 	[[ "${output}" == *"someDnsOption"* ]]
 	# stop-signal
 	[[ "${output}" == *"\"StopSignal\": \"SIGKILL\""* ]]
+
+	# following options are introduced in docker 1.10, skip older version
+	run docker --version
+	if [[ "${output}" != "Docker version 1.1"* ]]; then
```
dongluochen: This skips future versions like '1.20' or '2.x'. Maybe you can skip docker 1.9 which is the only version it doesn't support.

abronan:
Thanks @jimmyxian! LGTM when comment above is addressed.

test/integration/api/run.bats
jimmyxian: OK, Thanks for your advice. Dong! :)

jimmyxian:
@abronan Aha, will do it now.

dongluochen:
LGTM. Thanks @jimmyxian!

abronan:
:tada:

PR #1841, opened 2016-02-18
tests: increase zookeeper discovery retry count in CI

abronan:
Zookeeper tests seems a bit flaky since we added more travis instances or since the changes in node management. Because we have a variable timer now to connect to the Agents.

This increases the retry count to match the ones of other tests to give a bit more window to discover the nodes. Let's try with `20` for now.

### DIFF
--- a/test/integration/discovery/zk.bats
+++ b/test/integration/discovery/zk.bats
@@ -36,11 +36,11 @@ function teardown() {
 	# Start 2 engines and make them join the cluster.
 	start_docker 2
 	swarm_join "$DISCOVERY"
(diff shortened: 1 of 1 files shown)


nishanttotla:
LGTM

vieux:
LGTM

abronan:
Let's wait for Jenkins to trigger first before merging

abronan:
`not ok 146 refresh loop detects failure` failed in conjunction to `not ok 52 zk discovery: node removal`. Do not merge. This needs more investigation.

Issue #1846, opened 2016-02-19
Docker swarm manager doesn't restore exposed ports when rescheduling the container


antimack:
Hi,

I have a 3 node swarm cluster with swarm token(1 manager and 2 agent nodes, daemon on all of these nodes started with params DOCKER_OPTS="-H tcp://0.0.0.0:2375 -H unix:///var/run/docker.sock --label=environment=test")
what has been done:
1. Swarm manager running on master node with cmd "docker run -d --name swarm_master -p 3375:2375 swarm -experimental manage token://$SW_TOKEN"
2. Swarm agents running on master node with "docker run -d --name $NODE_NAME swarm -experimental join --addr=172.18.0.12:2375 token://$SW_TOKEN"
3. I've exported DOCKER_HOST on master node and started simple Nginx container with "docker run -d --name web1 --restart=always --env="reschedule:on-node-failure" -p 80:80 nginx"

```
vagrant@master:~$ docker info
Containers: 1
 Running: 1
 Paused: 0
 Stopped: 0
Images: 2
Server Version: swarm/1.1.2
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 2
 node2: 172.18.0.12:2375
  └ Status: Healthy
  └ Containers: 0
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 1.019 GiB
  └ Labels: environment=test, executiondriver=native-0.2, kernelversion=3.13.0-77-generic, operatingsystem=Ubuntu 14.04.4 LTS, storagedriver=aufs
  └ Error: (none)
  └ UpdatedAt: 2016-02-19T12:43:58Z
 node3: 172.18.0.13:2375
  └ Status: Healthy
  └ Containers: 1
  └ Reserved CPUs: 0 / 1
  └ Reserved Memory: 0 B / 1.019 GiB
  └ Labels: environment=test, executiondriver=native-0.2, kernelversion=3.13.0-77-generic, operatingsystem=Ubuntu 14.04.4 LTS, storagedriver=aufs
  └ Error: (none)
  └ UpdatedAt: 2016-02-19T12:43:45Z
Plugins:
 Volume:
 Network:
Kernel Version: 3.13.0-77-generic
Operating System: linux
Architecture: amd64
CPUs: 2
Total Memory: 2.038 GiB
Name: 0e03c10fa9b4
Experimental: true
```

```
vagrant@master:~$ docker ps
[...]

PR #1850, opened 2016-02-19
don't erase hostconfig on refresh

vieux:
Fix: https://github.com/docker/swarm/issues/1846

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -620,6 +620,9 @@ func (e *Engine) updateContainer(c dockerclient.Container, containers map[string
 		}
 		// Convert the ContainerConfig from inspect into our own
 		// cluster.ContainerConfig.
+		if info.HostConfig != nil {
+			info.Config.HostConfig = *info.HostConfig
+		}
 		container.Config = BuildContainerConfig(*info.Config)
 
 		// FIXME remove "duplicate" lines and move this to cluster/config.go
--- a/test/integration/rescheduling.bats
+++ b/test/integration/rescheduling.bats
@@ -171,3 +171,31 @@ function containerRunning() {
 	# c2 should have been rescheduled from node-1 to node-0
 	retry 5 1 containerRunning "c2" "node-0"
 }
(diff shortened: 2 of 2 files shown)


vieux:
ping @nishanttotla

cluster/engine.go
```diff
@@ -620,6 +620,9 @@ func (e *Engine) updateContainer(c dockerclient.Container, containers map[string
 		}
 		// Convert the ContainerConfig from inspect into our own
 		// cluster.ContainerConfig.
+		if info.HostConfig != nil {
+			info.Config.HostConfig = *info.HostConfig
```
nishanttotla: why is the explicit dereference needed here?

cluster/engine.go
```diff
@@ -620,6 +620,9 @@ func (e *Engine) updateContainer(c dockerclient.Container, containers map[string
...
+		}
 		container.Config = BuildContainerConfig(*info.Config)
 
 		// FIXME remove "duplicate" lines and move this to cluster/config.go
```
nishanttotla: In the lines below this, are the HostConfig updates needed anymore?

cluster/engine.go
vieux: because `HostConfig` embed in `Config` is an object, not a pointer

cluster/engine.go
vieux: yes

cluster/engine.go
nishanttotla: Right, my question was about the RHS. I didn't realize the `HostConfig` inside info is a pointer, so my bad.

nishanttotla:
LGTM.

nishanttotla:
Also ping @abronan seems like `21 leader election` test fails often.

dongluochen:
LGTM

nishanttotla:
Ping @vieux I think we should merge this.

abronan:
LGTM

PR #1855, opened 2016-02-22
support docker run --net <node>/<network>

vieux:

### DIFF
--- a/cluster/config.go
+++ b/cluster/config.go
@@ -196,6 +196,18 @@ func (c *ContainerConfig) RemoveAffinity(affinity string) error {
 	return nil
 }
 
+// AddConstraint to config
+func (c *ContainerConfig) AddConstraint(constraint string) error {
+	constraints := c.extractExprs("constraints")
+	constraints = append(constraints, constraint)
+	labels, err := json.Marshal(constraints)
+	if err != nil {
+		return err
+	}
+	c.Labels[SwarmLabelNamespace+".constraints"] = string(labels)
+	return nil
+}
+
 // HaveNodeConstraint in config
 func (c *ContainerConfig) HaveNodeConstraint() bool {
 	constraints := c.extractExprs("constraints")
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -170,6 +170,13 @@ func (c *Cluster) createContainer(config *cluster.ContainerConfig, name string, 
 		config.SetSwarmID(swarmID)
 	}
 
+	if network := c.Networks().Get(config.HostConfig.NetworkMode); network != nil && network.Scope == "local" {
+		if !config.HaveNodeConstraint() {
+			config.AddConstraint("node==~" + network.Engine.Name)
+		}
+		config.HostConfig.NetworkMode = network.Name
+	}
+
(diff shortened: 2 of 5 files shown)


vieux:
ping @dhiltgen

dhiltgen:
lgtm - I'll give it a spin.

cluster/config.go
```diff
@@ -109,6 +109,11 @@ func BuildContainerConfig(c dockerclient.ContainerConfig) *ContainerConfig {
 	// remove affinities/constraints/reschedule policies from env
 	c.Env = env
 
+	if parts := strings.SplitN(c.HostConfig.NetworkMode, "/", 2); len(parts) == 2 {
```
mavenugo: as of today, `network` name can have `/` character in it .... 
so, i guess `len(parts) == 2` check isnt correct. Also, we have to make sure we append the rest of the parts [1:] back to network name. and hence using parts[0] as the node identifier.

vieux: thanks, I'm going to change to [1:]. regarding the `/` in name, I don't think it can be supported in swarm right now

test/integration/api/network.bats
```diff
@@ -176,3 +175,15 @@ function teardown() {
...
+	start_docker_with_busybox 2
+	swarm_manage
+
+	docker_swarm network create -d bridge node-1/testn
+
+	docker_swarm run -d --net node-1/testn --name test_container busybox sleep 100
+
+	run docker_swarm network inspect testn
+	[[ "${output}" != *"\"Containers\": {}"* ]]
+}
```
mavenugo: Can you also add a non-compose test to handle my previous comment by creating a network with `/` character in it and making sure swarm is able to handle that ?

cluster/config.go
vieux: it's a `splitN` so it's already handled properly

dongluochen: @mavenugo Is there any requirement to support `/` in network name? Can we update networking to not allow `/`? I feel like we might have other problems if we allow arbitrary characters.

test/integration/api/network.bats
vieux: it's already impossible to create a network with a `/` with swarm

abronan:
LGTM

vieux:
@dhiltgen you might want to take another look, I redid the PR a bit cleaner

cluster/swarm/cluster.go
```diff
@@ -170,6 +170,11 @@ func (c *Cluster) createContainer(config *cluster.ContainerConfig, name string,
 		config.SetSwarmID(swarmID)
 	}
 
+	if network := c.Networks().Get(config.HostConfig.NetworkMode); network != nil {
+		config.AddConstraint("node==~" + network.Engine.Name)
```
dongluochen: I'm not sure if this added constraint is helpful or not. If a container is rescheduled 3 times at 3 different nodes, would all these soft constraints added up to `node==~node1; node==~node2; node==~node3`?

vieux: I think it it so the container land on a machine where the network is.

I added a check to not add multiple constraints

dongluochen:
LGTM

Issue #1847, opened 2016-02-19
Swarm 1.1.x volume plugin issues


echupriyanov:
This issue is related to https://github.com/docker/compose/issues/2958

Is it possible, that problem is on Swarm sife?
And also some observations.

My setup:
```
eric@iMac-Eric /V/D/W/s/a/elk-swarm> docker info
Containers: 6
 Running: 6
 Paused: 0
 Stopped: 0
Images: 12
Server Version: swarm/1.1.2
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 2
 swarm-node-05a.cybertonica.aws: swarm-node-05a.cybertonica.aws:2376
  └ Status: Healthy
  └ Containers: 3
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 0 B / 15.69 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=4.2.0-27-generic, operatingsystem=Ubuntu 14.04.4 LTS, storagedriver=aufs, zone=eu-west-1a
  └ Error: (none)
  └ UpdatedAt: 2016-02-19T13:25:24Z
 swarm-node-06a.cybertonica.aws: swarm-node-06a.cybertonica.aws:2376
  └ Status: Healthy
  └ Containers: 3
  └ Reserved CPUs: 0 / 2
  └ Reserved Memory: 0 B / 15.69 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=4.2.0-27-generic, operatingsystem=Ubuntu 14.04.4 LTS, storagedriver=aufs, zone=eu-west-1a
  └ Error: (none)
  └ UpdatedAt: 2016-02-19T13:25:36Z
Plugins: 
 Volume: 
 Network: 
Kernel Version: 4.2.0-27-generic
Operating System: linux
Architecture: amd64
CPUs: 4
Total Memory: 31.37 GiB
Name: f4939d899700
eric@iMac-Eric /V/D/W/s/a/elk-swarm> 
```
Volumes:
```
eric@iMac-Eric /V/D/W/s/a/elk-swarm> docker volume ls
DRIVER              VOLUME NAME
local               swarm-node-05a.cybertonica.aws/04a7ea4417f12f26a70dd6d28731e948eaa7b25605180633b0f5c95bb1f6b877
flocker             es_data_02
flocker             es_data_01
flocker             es_data_01
flocker             es_data_02
[...]

PR #1872, opened 2016-02-24
Fix get volume call for volumes with non-local drivers

echupriyanov:
Related to issue https://github.com/docker/swarm/issues/1847

### DIFF
--- a/cluster/volume.go
+++ b/cluster/volume.go
@@ -32,6 +32,12 @@ func (volumes Volumes) Get(name string) *Volume {
 	if size := len(candidates); size == 1 {
 		return candidates[0]
 	} else if size > 1 {
+		// Match first volume with non-local driver
+		for _, volume := range candidates {
+			if volume.Name == name && volume.Driver != "local" {
+				return volume
+			}
+		}
 		return nil
 	}
 
--- a/cluster/volume_test.go
+++ b/cluster/volume_test.go
@@ -1,0 +1,90 @@
+package cluster
+
+import (
+	"testing"
+
+	"github.com/samalba/dockerclient"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGet(t *testing.T) {
+	engines := []*Engine{
+		{ID: "59C8F4EC-3A71-44C7-BA50-0496B4D8E548",
+			Name: "swarm-node-0"},
+		{ID: "607B1F15-CECA-41B2-9D26-D2626EB14D8D",
+			Name: "swarm-node-1"},
+	}
+	volumes := Volumes{
+		&Volume{
+			Volume: dockerclient.Volume{
+				Name:   "t1",
+				Driver: "local",
+			},
+			Engine: engines[0],
+		},
+		&Volume{
+			Volume: dockerclient.Volume{
+				Name:   "t2",
+				Driver: "local",
+			},
+			Engine: engines[0],
+		},
+		&Volume{
+			Volume: dockerclient.Volume{
+				Name:   "t3",
+				Driver: "rexray",
+			},
+			Engine: engines[0],
+		},
+		&Volume{
+			Volume: dockerclient.Volume{
+				Name:   "t4",
+				Driver: "flocker",
+			},
+			Engine: engines[0],
+		},
+		&Volume{
+			Volume: dockerclient.Volume{
+				Name:   "t1",
+				Driver: "local",
+			},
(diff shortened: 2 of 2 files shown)


nishanttotla:
@echupriyanov do you mind adding a test for your change?

echupriyanov:
Original `cluster/volume.go` had no defined tests, so I’ve added them, including cases with various volume driver types.

wallnerryan:
@nishanttotla Would love to see this merged in before #swarmweek :)

lukemarsden:
@vieux would love to get your eyes on this, please.

cluster/volume_test.go
```diff
@@ -0,0 +1,90 @@
...
+			Engine: engines[1],
+		},
+	}
+	emptyVolumes := Volumes{}
+	assert.Nil(t, emptyVolumes.Get("test"))
+	assert.Nil(t, volumes.Get("t1"))
+	if assert.NotNil(t, volumes.Get(engines[0].ID+"/"+"t1")) {
+		assert.Equal(t, volumes.Get(engines[0].ID+"/"+"t1"), volumes[0])
+	}
+	if assert.NotNil(t, volumes.Get(engines[0].Name+"/"+"t1")) {
```
nishanttotla: @echupriyanov this test looks identical to the one above. I think you meant to use `engines[1]`?

vieux:
LGTM

cluster/volume_test.go
echupriyanov: @nishanttotla it's really 2 different tests - first we prefix volume by engine's ID, in second case we use engine's Name to identify concrete volume.

nishanttotla: Aaah yes you're right, I missed that tiny detail :)

cluster/volume_test.go
```diff
@@ -0,0 +1,90 @@
...
+		assert.Equal(t, volumes.Get(engines[0].Name+"/"+"t1"), volumes[0])
+	}
+	if assert.NotNil(t, volumes.Get("t3")) {
```
nishanttotla: @echupriyanov in your code, you always return the first volume that matches the name and has a non-local driver. Under what condition would you return the volume `t3` on `engines[1]`?

echupriyanov: Good question. Really, non-local volume are not bound to any particular engine, so I'm not sure which engine should be returned. As a base for my solution I used piece of code from lines 44+ in `cluster/volume.go` - it looks like unfinished work for handling non-local volumes. So I've decided to do the same and return first found match.
Regarding the test, `engine[1]` needed to test a cases with identically named local volumes on different engines, so `volumes.Get('t1')` will return `nil` as local volumes should be accessible only with engine's ID or Name prefix.

echupriyanov: An one more note. I've tried patched version with docker-compose 1.6 (v2 file) and and it correctly mounts volumes to different engines (nodes), though `Volumes.Get()` always returns first matched engine.

nishanttotla: cc @thajeztah what do you think?

thaJeztah: Let me ask @cpuguy83, he's most up-to-date with the current approach in docker/docker

cpuguy83: Currently engine probes all drivers for a volume with a given name and returns the first one.

wallnerryan: @cpuguy83 so since this is in line with what docker does today, and got a LGTM from can we merge soon @thaJeztah.  Ive been testing this out on a patched version of swarm as well and havent hit issues.

thaJeztah: @wallnerryan I'm not a maintainer in this repository, but if the behavior is the same as Docker Engine, this "LGTM". But the Swarm maintainers need to make that decision

cluster/volume.go
```diff
@@ -32,6 +32,12 @@ func (volumes Volumes) Get(name string) *Volume {
...
 		return candidates[0]
 	} else if size > 1 {
+		// Match first volume with non-local driver
+		for _, volume := range candidates {
+			if volume.Name == name && volume.Driver != "local" {
+				return volume
+			}
+		}
 		return nil
 	}
```
nishanttotla: @echupriyanov might local volumes also need the check with `/name` as in the loop below?

nishanttotla:
LGTM

Issue #1879, opened 2016-02-25
Proposal: Roadmap to replace samalba/dockerclient with engine-api/client


nishanttotla:
## Problem

Swarm has been using [`samalba/dockerclient`](https://github.com/samalba/dockerclient) to send requests to Engines in the cluster so far. This has been a limiting factor for supporting new features of the Engine API, since `dockerclient` is outdated and does not fully support the current Engine API. 
## Solution

[`engine-api/client`](https://github.com/docker/engine-api/) contains the officially supported Docker API, along with the relevant type structs and client. We need to be using that instead.
## Challenge

All Swarm changes currently go through `dockerclient`, all of which need to be moved over to use `engine-api`, along with new using new type structs defined in `engine-api/types` and updating unit and integration tests. This is a large change that will take time to fully accomplish and test for correctness.
## Proposal

The simplest way to start is to add a new `engine-api/client` object in the `Engine` struct in Swarm, which uses the same underlying HTTP connection that is used by `dockerclient`. This means that the two clients will co-exist in Swarm while using the same underlying connection.

Once this is done, we should start moving over individual API endpoints to use the new client. We must make sure that this does not affect or cause any issues to existing Swarm users, except for the addition of new features that `engine-api` allows us to implement.
## Roadmap

The proposal will be implemented in the following steps:
- [x] Create an `engine-api` client in Swarm, and use it to replace the `Version()` function with `ServerVersion()`. This will involve adding/updating Godeps, creating mock and nop clients, and updating tests. This part is handled in #1880.
- [x] Replace the `Info()` function to use its equivalent in `engine-api`. This is another small change, with several more changes likely needed, because of type mismatches between the `Info` structs in the two client libraries. For simplicity, we should perform type conversions (mostly between `int` and `int64` as needed) - this is a BAD idea in general, but only a temporary solution until some of the next few changes take effect. This part is handled in #1938.
- [x] Change the `ContainerConfig` struct in the `cluster` package to use `ContainerConfig` from `engine-api/types`. This should mostly take care of the messy type conversions required in the previous step. We will also need to change all functions that create/modify `ContainerConfig`. This change will span across a large part of the codebase, and move over complete container management to `engine-api`. This part is handled in #1983, #2127, #2128, #2194.
- [x] Starting in file `handlers.go`, focus on individual API endpoints to move them to the new client. The endpoints can be roughly grouped into the following categories
  - [x] Network modification operations (#1982, #2068)
[...]

PR #1880, opened 2016-02-25
Introducing engine-api client into Swarm

nishanttotla:
This is a proposed initial introduction of the `engine-api` client into Swarm, with the eventual goal of replacing `samalba/dockerclient`. Please follow related discussion in #1879.


vieux:
@nishanttotla good start but can you add a commit that is actually using the engine-api ? :)

nishanttotla:
@vieux so far I only did some local tests to check that simple API endpoints work, and thought that we should add commits that actually use engine-api later on. If you have suggestions for what endpoint(s) to include in this PR, I'm happy to add :)

MHBauer:
I think function calls on `e.client` show mainly what this would target:
```
λ ag 'e\.client' --ignore Godeps/
cluster/engine.go
170:	e.client.StartMonitorEvents(e.handler, ec)
220:	e.client.StopAllMonitorEvents()
376:	info, err := e.client.Info()
386:	v, err := e.client.Version()
433:	array, err := e.client.RemoveImage(name, force)
442:	err := e.client.RemoveNetwork(network.ID)
473:	err := e.client.RemoveVolume(name)
490:	images, err := e.client.ListImages(true)
506:	networks, err := e.client.ListNetworks("")
522:	volumes, err := e.client.ListVolumes()
540:	containers, err := e.client.ListContainers(true, false, "")
566:	containers, err := e.client.ListContainers(true, false, fmt.Sprintf("{%q:[%q]}", "id", ID))
616:		info, err := e.client.InspectContainer(c.Id)
671:			e.client.StopAllMonitorEvents()
798:	err := e.client.RemoveContainer(container.Id, force, volumes)
815:	response, err := e.client.CreateNetwork(request)
825:	volume, err := e.client.CreateVolume(request)
842:	err := e.client.PullImage(image, authConfig)
856:	err := e.client.LoadImage(reader)
870:	_, err := e.client.ImportImage(source, repository, tag, imageReader)
1045:	err := e.client.StartContainer(id, nil)
1059:	err := e.client.RenameContainer(container.Id, newName)
1072:	reader, err := e.client.BuildImage(buildImage)
1080:	err := e.client.TagImage(IDOrName, repo, tag, force)
```
Maybe something simple like `Info()` is where to start?

vieux:
@nishanttotla I would prefer not merge the PR as is, it's only adding dead code since it's used nowhere.

You might look into this method: https://github.com/docker/swarm/blob/master/cluster/engine.go#L375

Can you probably replace `.Info()` and `.Version()` easily.

nishanttotla:
@vieux yeah I don't intend for us to merge as is, without some concrete change like you suggested. I'll add some tiny changes for now.

cluster/engine.go
```diff
@@ -184,6 +185,14 @@ func (e *Engine) StartMonitorEvents() {
 func (e *Engine) ConnectWithClient(client dockerclient.Client) error {
 	e.client = client
 
+	// Use HTTP Client used by dockerclient to create engine-api client
```
dongluochen: I think it better to move this new code to `Connect()` method where client is initialized.

nishanttotla: I wanted to do that, but there are some complications it causes. For instance, if we want to keep a similar workflow where `Connect()` initializes a client and `ConnectWithClient()` updates engine fields, then the signature for `ConnectwithClient()` will have to change to pass both `client` and `apiClient`. This is turn affects the unit tests (and the need to create a mock client for `engine-api`). Maybe I'm missing something, WDYT?

dongluochen: You are right this is the place we add mock client. Let's keep it as it is.

nishanttotla:
I updated `Version()` but updating `Info()` might take some more work to match types, that's eventually going to require changing `ContainerConfig`. I'm working on that.

cluster/engine.go
nishanttotla: Yep. I couldn't get `mock` client working, even though it looks easy. I'll change this if I can.

doronp:
This is a good PR
But interferes with lots of other PRs :-(

nishanttotla:
@doronp thanks for pointing out. Which ones in particular? I'm curious because apart from Godeps, there aren't expected to be too many changes in this one, so even if there's conflict, it should be easy to work out.

vieux:
@nishanttotla why does `Info()` requires a change in `ContainerConfig` ? It only gets the engine info.

nishanttotla:
@vieux you're right. I said so only because I wanted to avoid some `int64` to `int` type conversions. This is done at some places such as
```
e.Cpus = info.NCPU
```
Changing the type of `e.Cpus` in `Engine struct` requires more changes that eventually leads to `ContainerConfig`.

doronp:
@nishanttotla e.g. The Generic plugin and authZ plugin PRs for Swarm (Huge Godeps change) one of:
 #1366 (commit 497e328) or #1646
;-)

nishanttotla:
@doronp you're right, but we shouldn't be afraid of this. For the most part, resolving conflicts should be easy - we just choose the latest godep version for a repo, which I suppose will be the one in this PR (since it's most recent). Besides, it might be a while before this one gets merged. Just for the record, here are the Godeps updated/created in this PR:
```
_workspace/src/github.com/docker/go-connections/
_workspace/src/github.com/opencontainers/runc/
_workspace/src/golang.org/x/net/proxy
_workspace/src/github.com/Microsoft/go-winio/
_workspace/src/github.com/docker/engine-api/client/
_workspace/src/github.com/docker/engine-api/types/
```

doronp:
The first 2 are in the intersection - which versions hashes well take?

nishanttotla:
@doronp I think the last one is also in the intersection. I looked at your PRs (and this one), and here are the commits we should pick for Godeps:
- `_workspace/src/github.com/docker/go-connections/`
 - `34b5052da6b11e27f5f2e357b38b571ddddd3928` (from the current PR, since it's more recent. Note that your two PRs conflict with each other for this)
- `_workspace/src/github.com/opencontainers/runc/`
 - I'm not sure, we should probably just update to the latest version. If your PR gets in after me, you can do it, or else I can. Note that your two PRs conflict with each other on this.
- `_workspace/src/github.com/docker/engine-api/types/
 - `adfa32fb6ccad724d73a16d53ce3defa46df27e5` (from the current PR, since it's more recent)

Please make sure to not overwrite more recent versions, since some of my changes might depend on those.

nishanttotla:
@docker/core-swarm-maintainers this is now ready for review. PTAL :)

A little context

1. I was able to successfully introduce `engine-api` client into Swarm and make it use the same HTTP Client as `samalba/dockerclient`. The only call made by the new client is `ServerVersion()`, which replaces `dockerclient.Version()`. Really small change, but needed a lot of additional changes.

2. I decided to forgo the mock client in `engine-api`, since it's very differently structured, and have a dedicated one inside of Swarm. I also added a nopclient. (related discussion https://github.com/docker/engine-api/issues/117)

Finally, I'm not sure why the lint check is failing in the CI, any help on that is appreciated!

vieux:
@nishanttotla check golint :)

nishanttotla:
Thanks @vieux, tests should pass now. Ping @dongluochen @abronan @MHBauer.

I'd say we should review this PR thoroughly, because future commits that address the move to `engine-api` will heavily depend on this.

cluster/engine.go
```diff
@@ -105,6 +107,7 @@ type Engine struct {
 	networks        map[string]*Network
 	volumes         map[string]*Volume
 	client          dockerclient.Client
+	apiClient       engineapi.APIClient
```
dongluochen: I feel like it may be better to wrap {client, apiClient, failureCount, updateAt} into a struct. The first change is to switch dockerclient.Client to this new client. Next changes are mostly inside this new client struct. It may be better to isolate the changes. Additional benefit is to get failureCount integrated.

nishanttotla: @dongluochen can you elaborate on why `failureCount` and `updatedAt` should be part of the struct too?

Also, to clarify further, you're saying that these should be in a struct which is then part of the Engine struct right? I'm wondering if that should be part of this PR, WDYT?

nishanttotla:
Ping @jimmyxian

cluster/engine.go
```diff
@@ -159,8 +163,10 @@ func (e *Engine) Connect(config *tls.Config) error {
 	if err != nil {
 		return err
 	}
+	// Use HTTP Client used by dockerclient to create engine-api client
+	apiClient, err := engineapi.NewClient("tcp://"+e.Addr, "", c.HTTPClient, nil)
```
nishanttotla: We should probably return if `err` is not `nil`

vieux:
@nishanttotla can you rebase and if it's not too much trouble, split the PR into only 2 commits, one with godeps and one with the rest ?

nishanttotla:
@vieux oh alright. It looks like the latest merge produced conflicts, so rebase shouldn't be hard. I'll have to figure out how to squash commits though.

nishanttotla:
@vieux it's done! Let me know if this looks good :)

cluster/engine.go
```diff
@@ -404,7 +416,7 @@ func (e *Engine) updateSpecs() error {
 		return fmt.Errorf("cannot get resources for this engine, make sure %s is a Docker Engine, not a Swarm manager", e.Addr)
 	}
 
-	v, err := e.client.Version()
+	v, err := e.apiClient.ServerVersion()
 	e.CheckConnectionErr(err)
```
dongluochen: Please update `CheckConnectionErr` to handle errors from engine-api. In engine-api, a connection failure message is the following, which is not covered in `CheckConnectionErr`.  

```
// ErrConnectionFailed is a error raised when the connection between the client and the server failed.
   var ErrConnectionFailed = errors.New("Cannot connect to the Docker daemon. Is the docker daemon running on this host?")        
```

nishanttotla: @dongluochen thanks! Good catch.

nishanttotla: @dongluochen updated, thanks!

nishanttotla:
@vieux @dongluochen do you have more comments? If we can merge this today, I could start with the next steps in the proposal :)

vieux:
No more comment from me :)

dongluochen:
LGTM.

chanwit:
:tada: :tada: :tada:

MHBauer:
Nice.

PR #1892, opened 2016-02-25
Return a 404 error when an image is not available / not accessible on run

MatMaul:
On a run return the same 404 error as the docker engine would when an image is not available / not accessible.
The client can then automatically try to pull the image (including the auth header if defined).

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -516,23 +516,20 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 		httpError(w, err.Error(), http.StatusBadRequest)
 		return
 	}
-	// Pass auth information along if present
-	var authConfig *dockerclient.AuthConfig
-	buf, err := base64.URLEncoding.DecodeString(r.Header.Get("X-Registry-Auth"))
-	if err == nil {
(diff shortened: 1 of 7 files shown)


vieux:
@MatMaul if you do this, the image will be pulled on every node, we don't want that.
We want to pull only on the node the scheduler selected.

MatMaul:
Thanks for the feedback, I didn't though about that. What do you think would be a better way to improve the situation regarding registry auth ? Currently it only supports one authorization for one registry.

I am thinking about two ways :
- modify the docker client to include the same x-registry-auth header with a run as with a pull. This would be the cleanest way (no more fiddling with HttpHeaders inside the config file and repeat the information a second time). I have a patch to do that for the client side, and some WIP to support it in the daemon (do we want that supported in the daemon anyway ?).
- Another possibility without being intrusive into docker itself is to embed the full auths from the config.json into the HttpHeaders. The good auth can then be selected in the manager. We could also write a small script that keep them in sync.

Let me know if you prefer that I open an issue to discuss it.

MatMaul:
https://github.com/MatMaul/docker/tree/registry-auth-on-run

Issue #1619, opened 2016-01-12
Proposal: Swarm use revocable resources from Mesos


ghost:
After [MESOS-1607](https://issues.apache.org/jira/browse/MESOS-1607) and [MESOS-354](https://issues.apache.org/jira/browse/MESOS-354), the revocable resources will offer to Swarm; it'll improve the resources utilisation if Swarm used revocable resources.

I'm also working on [MESOS-4303](https://issues.apache.org/jira/browse/MESOS-4303) to clarify the protocol between Swarm & Mesos; if any suggestion, please let me know.

jimenez:
@klaus1982 +1 on this, thanks for contributing!

qianzhangxa:
I has submitted the design proposal in #1900.

nishanttotla:
Closing due to https://github.com/docker/swarm/pull/2853

PR #1900, opened 2016-02-27
Proposal: Enable Swarm to use Mesos revocable resources.

qianzhangxa:
Currently Swarm can only receive regular resources from Mesos. I think we need to enable Swarm to also receive revocable resources from Mesos, and enable Swarm users to choose type of resources to create their containers, e.g., create high priority container with regular resources and low priority container with revocable resources. In this way, when there are other frameworks running on Mesos along with Swarm, the underutilized resources of other frameworks can be used by Swarm so that the overall resource utilization of the Mesos cluster can be improved.

See the file for a complete description.

### DIFF
--- a/cluster/mesos/use-mesos-revocable-resource-proposal.md
+++ b/cluster/mesos/use-mesos-revocable-resource-proposal.md
@@ -1,0 +1,124 @@
+# Use Mesos Revocable Resource in Swarm
+
+## Introduction
+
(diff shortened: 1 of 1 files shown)


qianzhangxa:
@klaus1982 has created an issue #1619 to trace this, thanks @klaus1982!

gradywang:
PR: #1946

nishanttotla:

PR #1908, opened 2016-02-29
Add missing 'defer'

duglin:
as it stands now the lock at this spot in the code is pretty useless.
All it does it block us but since we release the lock before we check
e.state its a roadblock with no real purpose.
One could argue that a lock isn't needed at all but I think having it
there for consistency with the other funcs would be good.

Signed-off-by: Doug Davis <dug@us.ibm.com>

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -231,7 +231,7 @@ func (e *Engine) isConnected() bool {
 // IsHealthy returns true if the engine is healthy
 func (e *Engine) IsHealthy() bool {
 	e.RLock()
-	e.RUnlock()
+	defer e.RUnlock()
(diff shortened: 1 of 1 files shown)


dongluochen:
LGTM

vieux:
LGTM

dongluochen:
Docker registry is blocking integration tests. 
```
Error response from daemon: Get https://registry-1.docker.io/v2/dockerswarm/swarm-test-env/manifests/latest: Received unexpected HTTP status: 500 Internal Server Error
Command exited with non-zero status 1
```

duglin:
CI all green now

abronan:
Thanks @duglin, LGTM

PR #1910, opened 2016-03-01
Syntax error - changed container to constraint

JohnKDay:
Syntax Error
From:
The `<filter-type>` is either the `affinity` or the `container` keyword. It
identifies the type filter you intend to use.
To:
The `<filter-type>` is either the `affinity` or the `constraint` keyword. It
identifies the type filter you intend to use.


nishanttotla:
LGTM.

Thanks @JohnKDay

PR #1930, opened 2016-03-07
Prevent crash on double calls to Disconnect

dongluochen:
Currently `Disconnect` can be called multiple times, either triggered by `monitorDiscovery` multiple times, or in Mesos case by `removeOffer`. This would crash the manager. Add a `stateDisconnected` to mark an engine removed from the cluster. Resource clean up should be done only once. 

This is part of fix for #1838. 


abronan:
LGTM

vieux:
LGTM thanks @dongluochen

Issue #1771, opened 2016-02-08
install with machine step 2 needs to be improved


moxiegirl:
https://twitter.com/jjasghar/status/696752027251003392

![image](https://cloud.githubusercontent.com/assets/1347209/12894036/59804672-ce49-11e5-9d13-57c45378e435.png)

nishanttotla:
Related #1767

The tutorial in the current form does not work. I'm working on fixing it now.

nishanttotla:
@moxiegirl I think the issue here is also that JJ Asghar is confused that the two docs https://docs.docker.com/swarm/provision-with-machine/ and https://docs.docker.com/swarm/install-w-machine/ are different. Why do we have two of these?

moxiegirl:
@nishanttotla `provision-with-machine.md` explains that it is possible but doesn't show how to do it step-by-step.  `install-w-machine` which we will rename to remove the `machine` assumes a local install in a Toolbox environment but uses the regular `swarm` image to provision.

PR #1932, opened 2016-03-08
Updated Swarm with Machine documentation

jonasrosland:
Removed a lot of information that's not useful for beginners getting started with Docker Swarm.
Rewrote large portions to make use of the `swarm` commands instead of manual configuration.

Signed-off-by: jonasrosland <jonas.rosland@gmail.com>

### DIFF
--- a/docs/install-w-machine.md
+++ b/docs/install-w-machine.md
@@ -14,7 +14,7 @@ weight=-50
 This getting started example shows you how to create a Docker Swarm, the
 native clustering tool for Docker.
 
-You'll use Docker Toolbox to install Docker Machine and some other tools on your computer. Then you'll use Docker Machine to provision a set of Docker Engine hosts. Lastly, you'll use Docker client to connect to the hosts, where you'll create a discovery token, create a cluster of one Swarm manager and nodes, and manage the cluster.
+You'll use Docker Toolbox to install Docker Machine and some other tools on your computer. Then you'll use Docker Machine to provision a set of Docker Engine hosts. Lastly, you'll use Docker client to connect to the hosts, where you'll create a discovery token, create a cluster of one Swarm manager and two nodes, and manage the cluster.
 
 When you finish, you'll have a Docker Swarm up and running in VirtualBox on your
 local Mac or Windows computer. You can use this Swarm as personal development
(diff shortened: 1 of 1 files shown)


vieux:
ping @moxiegirl

moxiegirl:
@jonasrosland Thank you for making this contribution.  Our experience was that users new to Swarm were confused by coming to a "Swarm" page and learning a number of Machine commands. This, in part, why we rewrote this page. We wanted  to de-emphasize the role of `docker-machine`.

The existing content is working as we intended. We'd want to reject this PR and but not the content.  I have two suggestions:

* Make a PR with this content into Docker Machine documentation
* Create a blog post with your content and we will reference it from this page: https://docs.docker.com/swarm/provision-with-machine/

What do you think?

jonasrosland:
Hi @moxiegirl, the reasoning I had was that I learnt these easier to remember commands when getting started with Swarm a while back, and was surprised to see the advanced commands that aren't really needed (IMHO) when looking at the new version of the docs.

Since we already have the `swarm` commands implemented in Docker Machine my thought was that should be promoted instead of demoted in favor of manual commands, to make it easier to find information in the commands themselves (such as `docker-machine create` which displays all the Swarm commands).

Looking at the linked docs it covers pretty much what I've added here, maybe the two docs "Evaluate Swarm in a sandbox" and "Provision with Machine" could be merged somehow? It looks like they are targeting similar setups, so perhaps that would be an option?

jonasrosland:
Hi again @moxiegirl, I think I was a bit confused about the docs, since this file is explicitly named `install-w-machine.md`, and the other is `provision-with-machine.md`. Perhaps a renaming to say `swarm-sandbox.md` or something would remove the confusion (at least for me)? I'll look into the `provision-with-machine.md` docs and see if I can add any value there. I still think it's a bit confusing having both of these though.

nishanttotla:
^ #1771 might be slightly related

moxiegirl:
@nishanttotla @jonasrosland I see the point.  I can certainly rename this file. If you are good with that. I'll close this issue, rename the file, and redirect.  WDYT?

jonasrosland:
@moxiegirl I think renaming the document would be a good idea. The question in #1771 still remains though, why are there two similar documents with essentially the same outcome? The Twitter request links to the blog post I wrote that contains the commands that I've added into the document here.

moxiegirl:
@jonasrosland  `provision-with-machine.md` explains that it is possible but doesn't show how to do it step-by-step.  `install-w-machine` which we will rename to remove the `machine` assumes a local install in a Toolbox environment but uses the regular `swarm` image to provision.

jonasrosland:
Sounds good @moxiegirl. I'll see if I can add something to the other docs then, you can close this now :)

moxiegirl:
Thanks @jonasrosland!

PR #1938, opened 2016-03-09
Updating info function to use engine-api

nishanttotla:
This PR moves over the `Info()` function call over to use `engine-api`, along with corresponding tests. Some mismatch between `int` and `int64` values needed to be adjusted.

This is part of #1879.

### DIFF
--- a/cluster/swarm/cluster_test.go
+++ b/cluster/swarm/cluster_test.go
@@ -26,7 +26,7 @@ func (nopCloser) Close() error {
 }
 
 var (
-	mockInfo = &dockerclient.Info{
+	mockInfo = types.Info{
 		ID:              "test-engine",
 		Name:            "name",
 		NCPU:            10,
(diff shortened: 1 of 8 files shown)


cluster/swarm/cluster_test.go
```diff
@@ -187,15 +186,14 @@ func TestLoadImage(t *testing.T) {
 	// create mock client
 	client := mockclient.NewMockClient()
 	apiClient := engineapimock.NewMockClient()
-	client.On("Info").Return(mockInfo, nil)
+	apiClient.On("Info").Return(mockInfo, nil)
+	apiClient.On("ServerVersion").Return(mockVersion, nil)
```
vieux: why did you have to move `ServerVersion` up in every tests ?

nishanttotla: For cosmetic reasons mostly. Last time I put it separately, away from all the `client.On(...).Return(...)` calls, but it makes sense to have everything together. No change in functionality.

nishanttotla:
ping @dongluochen 

`engine-master` build failed to run, I saw this happen on some other PR also.

dongluochen:
LGTM

vieux:
LGTM

PR #1943, opened 2016-03-10
Deleting redundant code from volume.go

nishanttotla:
Some code in `volume.go` is not needed, and was likely left behind.

### DIFF
--- a/cluster/volume.go
+++ b/cluster/volume.go
@@ -48,9 +48,5 @@ func (volumes Volumes) Get(name string) *Volume {
 		}
 	}
 
-	if len(candidates) == 1 {
-		return candidates[0]
-	}
-
 	return nil
 }


nishanttotla:
Ping @vieux @dongluochen 

Also wondering whether the following is necessary?
```
	// Match /name and return as soon as we find one.
	for _, volume := range volumes {
		if volume.Name == "/"+name {
			return volume
		}
	}
```

MHBauer:
LGTM

The code being removed is duplicated on L#32-33, and candidates is not being assigned into or appended between that reference and this one.

I do not know if the L#44-49 code needs to go away or be L#36 case, or what.

nishanttotla:
Thanks @MHBauer, that's my understanding as well. Will wait for more comments to see what we can do with L#44-49.

dongluochen:
This change LGTM. I think L#44-49 is needed as it's not covered before. The comments need update.

nishanttotla:
@dongluochen Thanks. I'm actually curious about L#44-49 in the context of #1872. Do you know under what conditions L#44-49 might be needed?

PR #1972, opened 2016-03-15
HostConfig for /start endpoint

vieux:
replace https://github.com/docker/swarm/pull/1829/
fix https://github.com/docker/swarm/issues/1781

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -749,7 +749,27 @@ func postContainersStart(c *context, w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if err := c.cluster.StartContainer(container); err != nil {
+	hostConfig := &dockerclient.HostConfig{
+		MemorySwappiness: -1,
+	}
+
+	buf, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		httpError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	r.Body.Close()
+
+	if len(buf) <= 2 {
+		hostConfig = nil
+	} else {
+		if err := json.Unmarshal(buf, hostConfig); err != nil {
+			httpError(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+	}
+
+	if err := c.cluster.StartContainer(container, hostConfig); err != nil {
 		httpError(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
--- a/cluster/cluster.go
+++ b/cluster/cluster.go
@@ -27,7 +27,7 @@ type Cluster interface {
 	Containers() Containers
 
 	// Start a container
-	StartContainer(container *Container) error
+	StartContainer(container *Container, hostConfig *dockerclient.HostConfig) error
 
 	// Return container the matching `IDOrName`
 	// TODO: remove this method from the interface as we can use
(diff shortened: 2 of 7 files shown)


nishanttotla:
I'm a little unsure about when to make this change, since it certainly conflicts with using container config structs from `engine-api`.

vieux:
@nishanttotla it fixes an issue in 1.1 I don't see why not merge it

api/handlers.go
```diff
@@ -749,7 +749,28 @@ func postContainersStart(c *context, w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	buf, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		httpError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	r.Body.Close()
+
+	fmt.Println(len(buf))
```
nishanttotla: @vieux did you mean to remove this print statement?

vieux: yes good catch, thanks

vieux: fixed

vieux:
ping @dongluochen

nishanttotla:
@vieux I thought about it some more, and the reason I was unsure was that `engine-api` doesn't allow for a `HostConfig` on container start. So this change may not be as simple eventually: https://github.com/docker/engine-api/blob/master/client/container_start.go

vieux:
@nishanttotla you'll have to change engine-api then, as long as our minimum supported engine version ca do it, we should do it :)

nishanttotla:
Aaah it's so complicated :)

It seems like the daemon code still supports this: https://github.com/docker/docker/blob/master/daemon/start.go#L16

But the plan is to phase it out in the upcoming release, so I really doubt we can build it into `engine-api`, but I'll ask around.

api/handlers.go
```diff
@@ -749,7 +749,27 @@ func postContainersStart(c *context, w http.ResponseWriter, r *http.Request) {
...
+	hostConfig := &dockerclient.HostConfig{
+		MemorySwappiness: -1,
+	}
+
+	buf, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		httpError(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+	r.Body.Close()
```
dongluochen: My understanding is that this is unnecessary. Request body should be close by the transport automatically.

dongluochen:
I think it's a complicated issue in the middle migrating client. I'm fine with either decision.

nishanttotla:
Even if we upped the minimum supported Engine version, we will have to support this for a little while at least.

One suggestion I have is to update config of the container first if a `HostConfig` is passed on start using `ContainerUpdate` (https://github.com/docker/engine-api/blob/master/client/container_update.go). Then call normal start without `HostConfig`. Is that a reasonable solution?

vieux:
@nishanttotla it might, but let's not block this PR it can be resolved after.

nishanttotla:
LGTM

dongluochen:
LGTM besides the line with `r.Body.Close()`.

nishanttotla:
cc @calavera

api/handlers.go
calavera: Closing doesn't have any side effect, so it should be okay to close it here. You cannot read the buffer again anyways.

dongluochen:
LGTM

Issue #1468, opened 2015-11-30
Docker PS from Swarm and Node Daemon are Different for Some Containers


ekristen:
Not really sure how to best describe this, but docker ps is reporting different information via swarm vs the node's daemon.

While docker ps is not a huge deal, I think this is having some side affects with docker compose and when it goes to deploy a new version of an image, constraints are failing when looking for the current container==CONTAINER_ID

Swarm Version: 1.0.0
Docker Version: 1.8.1 (on the master, on the slave)

`docker ps` from Swarm Master
```
CONTAINER ID        IMAGE                                COMMAND                  CREATED             STATUS                         NAMES
9a2278a12aa0        index.private.io/reports             "node server/index.js"   11 days ago         Restarting (5) 7 days ago      staging-apps02/app_reports_1
ef58f2bd3b0e        index.private.io/reports:dev         "node app.js"            12 days ago         Restarting (1) 9 minutes ago   staging-apps02/app2_reports_1
```

`docker ps` from the actual docker daemon
```
CONTAINER ID        IMAGE                                 COMMAND                  CREATED             STATUS              NAMES
9a2278a12aa0        index.private.io/reports              "node server/index.js"   11 days ago         Up 7 days           app_reports_1
ef58f2bd3b0e        index.private.io/reports:dev          "node app.js"            12 days ago         Up 46 seconds       app2_reports_1
```

`docker inspect staging-apps02/app_reports_1` from swarm master
```
[
{
    "Id": "9a2278a12aa0d28edf0471d6f189a69637415eed73b91de1ca0bd28fd9a4f7f2",
    "Created": "2015-11-18T22:02:59.851932524Z",
    "Path": "node",
    "Args": [
        "server/index.js"
    ],
    "State": {
        "Running": true,
        "Paused": false,
        "Restarting": false,
        "OOMKilled": false,
        "Dead": false,
        "Pid": 3328,
        "ExitCode": 0,
        "Error": "",
        "StartedAt": "2015-11-23T18:40:10.716378013Z",
        "FinishedAt": "2015-11-23T18:40:10.563824412Z"
    },
    "Image": "c86e29bafb5f0259d604a660443aa79e804d788895565e077a208119cf80e2af",
    "NetworkSettings": {
        "Bridge": "",
        "EndpointID": "0fb4b59753e5a1b382f8a2ff6df42e0f6fa331f80a6b66d290056bd1e89dc09b",
        "Gateway": "172.17.42.1",
[...]

PR #1981, opened 2016-03-16
Docker restart event should trigger container inspect

dongluochen:
Docker restart would change container state. Should trigger container inspect to update status.

The issue is tracked by #1468. 


nishanttotla:
LGTM

vieux:
LGTM

PR #1982, opened 2016-03-16
use apiClient for various things

MHBauer:
engine-api has has the prune/noprune option. The matching behavior is to prune. Not sure if we want or need to make that option available with the pass through.

Not sure if the array of pointers vs array of objects needs to be looked at or if it's fine how it is.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -569,14 +569,14 @@ func (e *Engine) RefreshNetworks() error {
 
 // RefreshVolumes refreshes the list of volumes on the engine.
 func (e *Engine) RefreshVolumes() error {
-	volumes, err := e.client.ListVolumes()
+	volumesLsRsp, err := e.apiClient.VolumeList(context.TODO(), filters.NewArgs())
 	e.CheckConnectionErr(err)
 	if err != nil {
 		return err
 	}
 	e.Lock()
 	e.volumes = make(map[string]*Volume)
-	for _, volume := range volumes {
+	for _, volume := range volumesLsRsp.Volumes {
 		e.volumes[volume.Name] = &Volume{Volume: *volume, Engine: e}
 	}
 	e.Unlock()
@@ -887,8 +887,8 @@ func (e *Engine) CreateNetwork(request *dockerclient.NetworkCreate) (*dockerclie
 }
 
 // CreateVolume creates a volume in the engine
-func (e *Engine) CreateVolume(request *dockerclient.VolumeCreateRequest) (*Volume, error) {
-	volume, err := e.client.CreateVolume(request)
+func (e *Engine) CreateVolume(request *types.VolumeCreateRequest) (*Volume, error) {
+	volume, err := e.apiClient.VolumeCreate(context.TODO(), *request)
 
 	e.RefreshVolumes()
 	e.CheckConnectionErr(err)
@@ -896,7 +896,7 @@ func (e *Engine) CreateVolume(request *dockerclient.VolumeCreateRequest) (*Volum
 	if err != nil {
 		return nil, err
 	}
-	return &Volume{Volume: *volume, Engine: e}, nil
+	return &Volume{Volume: volume, Engine: e}, nil
 
 }
 
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -477,12 +477,12 @@ func (e *Engine) updateSpecs() error {
 }
 
 // RemoveImage deletes an image from the engine.
-func (e *Engine) RemoveImage(image *Image, name string, force bool) ([]*dockerclient.ImageDelete, error) {
-	array, err := e.client.RemoveImage(name, force)
+func (e *Engine) RemoveImage(name string, force bool) ([]types.ImageDelete, error) {
+	rmOpts := types.ImageRemoveOptions{name, force, true}
+	dels, err := e.apiClient.ImageRemove(context.TODO(), rmOpts)
 	e.CheckConnectionErr(err)
 	e.RefreshImages()
-	return array, err
-
+	return dels, err
 }
 
 // RemoveNetwork removes a network from the engine.
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -487,7 +487,7 @@ func (e *Engine) RemoveImage(name string, force bool) ([]types.ImageDelete, erro
 
 // RemoveNetwork removes a network from the engine.
 func (e *Engine) RemoveNetwork(network *Network) error {
-	err := e.client.RemoveNetwork(network.ID)
+	err := e.apiClient.NetworkRemove(network.ID)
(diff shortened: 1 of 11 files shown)


nishanttotla:
Adding reference to #1879, this is part of implementing that proposal.

MHBauer:
Yup, I should have marked that. I wanted to give a shot at starting. I might combine some more stuff into this as individual commits. I would like to get some basic eyes-on to start.

nishanttotla:
@MHBauer thanks for helping out with moving to the new client. I'm thinking that it would be a good idea to add some unit tests so that we can monitor these changes over time to ensure stability.

MHBauer:
I'll have to think about how best to do that. I was mainly making sure the integration tests still ran, as I assume those are *using* the actual clients, vs what would probably be a mock here.

MHBauer:
@GordonTheTurtle what are you talking about. They're all signed....

nishanttotla:
@MHBauer that's correct, but unit tests are valuable to catch certain issues integration tests won't. Your changes seem to be isolated enough so it may not be a problem, but in general it would be nice to have unit tests in all packages that use the new client.

cluster/engine.go
```diff
@@ -471,17 +473,17 @@ func (e *Engine) updateSpecs() error {
 }
 
 // RemoveImage deletes an image from the engine.
-func (e *Engine) RemoveImage(image *Image, name string, force bool) ([]*dockerclient.ImageDelete, error) {
-	array, err := e.client.RemoveImage(name, force)
+func (e *Engine) RemoveImage(image *Image, name string, force bool) ([]types.ImageDelete, error) {
```
MHBauer: What is the purpose if the `image *Image` parameter? It wasn't used before, and it's not used after.

nishanttotla: Good point, seems like a redundant argument. You can consider removing it. I'm surprised the compiler didn't complain about it before.

MHBauer:
@nishanttotla test added for `RemoveImage`. I'm not terribly convinced of it's worth considering the work involved. I suppose it does make sure we're calling the one client vs the other.

MHBauer:
I think this is enough for one PR. @docker/core-swarm-maintainers for review, please.

nishanttotla:
@MHBauer sorry for the inconvenience, but could you add wait to rebase until #1996 is merged? Since we're now moving many function calls to `engine-api`, we should keep up to date with changes in method signatures.

MHBauer:
rebased. I think some lines got moved around incorrectly. I am checking that. The end result does build, and the context that the commits provide to each other is all self-contained within this PR.

nishanttotla:
@MHBauer any updates on this? I think we should try to merge it before the final 1.2 release. `engine-master` tests are failing due to #2002, but can you check if the `engine-1.10` failure is related or not?

MHBauer:
I highly doubt it. I don't want to keep punching the build machine til it works. Integration-tests work locally. Can you reproduce the failure? I'll run them again here.

nishanttotla:
@MHBauer if they work locally then it's fine, I just wanted to be sure :)

MHBauer:
I need to rebase this.

nishanttotla:
@MHBauer what's the status of this PR?

I'm slightly surprised that there are no merge conflicts, so can you make sure that everything's alright? :)

We'd like to merge this for 1.2.

vieux:
@MHBauer please rebase :)

vieux:
LGTM

MHBauer:
reproduces locally as well.

MHBauer:
hmn, it's failing on master as well. Now I'm confused.

vieux:
@MHBauer it's unrelated.

@nishanttotla let's merge

cluster/engine_test.go
```diff
@@ -156,10 +158,12 @@ func TestEngineCpusMemory(t *testing.T) {
 	apiClient := engineapimock.NewMockClient()
 	apiClient.On("Info", mock.Anything).Return(mockInfo, nil)
 	apiClient.On("ServerVersion", mock.Anything).Return(mockVersion, nil)
+	apiClient.On("NetworkList", mock.Anything,
+		mock.AnythingOfType("NetworkListOptions"),
+	).Return([]types.NetworkResource{}, nil)
+	apiClient.On("VolumeList", mock.Anything, mock.Anything).Return(types.VolumesListResponse{}, nil)
```
nishanttotla: @MHBauer could you do something like `mock.AnythingOfType("VolumeListOptions")` to be consistent with the above?

cluster/engine_test.go
```diff
@@ -182,10 +186,12 @@ func TestEngineSpecs(t *testing.T) {
...
```
nishanttotla: same here.

cluster/engine_test.go
```diff
@@ -460,3 +472,23 @@ func TestDisconnect(t *testing.T) {
...
 }
+
+func TestRemoveImage(t *testing.T) {
+	engine := NewEngine("test", 0, engOpts)
+
+	imageName := "test-image"
+	dIs := []types.ImageDelete{{Deleted: imageName}}
+
+	apiClient := engineapimock.NewMockClient()
+	apiClient.On("ImageRemove", context.TODO(), mock.Anything).Return(dIs, nil)
```
nishanttotla: This shouldn't be `context.TODO()`, but `mock.Anything` because we might pass a non-empty context in the future. The second argument can be of a specific type though, like suggested above.

nishanttotla:
@MHBauer it looks good overall, just some minor fixes :)

cluster/engine_test.go
MHBauer: No. VolumeList takes a generic filters.Args. NetworksList has a specific struct that it takes. 

I can try ` mock.AnythingOfType("Args")` but that might be less helpful than just being `Anything`. What do you think?

nishanttotla: I'd rather that we be as specific as possible. Especially since all types are changing, and just in case any stray arguments get passed it'd be nice to know (even though for this particular case I don't see that happening). It also helps with debugging in the future.

cluster/engine_test.go
MHBauer: I would rather the test args match what's being passed as close as possible. When the code changes to stop using the null context, we should reevaluate the test at that time as well.

I can see a different improvement to be made here.

cluster/engine_test.go
MHBauer: edited to be `Args`.

dongluochen:
LGTM

MHBauer:
Same two flaky tests, nothing more, so I'll merge it.

nishanttotla:
LGTM. Test failures unrelated, as confirmed by @abronan.

nishanttotla:
@MHBauer :tada:

PR #1983, opened 2016-03-17
Updating ContainerConfig to use engine-api

nishanttotla:
**This is work in progress, PR is meant for review**

This is the first change that will truly affect the entire repo AND the users, and will need discussion as some old API conventions might have to be deprecated.

The important decision here is the change in `ContainerConfig` struct inside `config.go`. Initially, `ContainerConfig` looked like

```
 type ContainerConfig struct {
     dockerclient.ContainerConfig
 }
```
but now it looks like
```
 type ContainerConfig struct {
     container.Config
     container.HostConfig
     network.NetworkingConfig
 }
```
where `container` and `network` are packages in `engine-api/types`. This is keeping in line with the recent Docker API changes where `HostConfig` and `NetworkingConfig` was moved out of `ContainerConfig` (renamed as `Config`) to be independent.

### "This change is too big"
I'm afraid so, but there's no other way to do it, all of this change needs to happen at once. Three function calls have been moved over to use `engine-api`, namely `ContainerCreate`, `ContainerInspect`, and `ContainerList`, along with the relevant changes needed to make these work.

### "How do I review this?"
Fortunately, many changes can be classified as "plumbing", so the root changes are fairly small. Start with `cluster/config.go` and `cluster/container.go` to see how structs have changed. Then check out `cluster/engine.go` where all the real stuff is happening, and scan through the calls being made to ensure that it all makes sense. **Note:** We don't have an embedded `HostConfig` inside `Config` anymore, they're separate, so we don't have to copy data over each time.

Finally, please do pull this branch and test locally, to make sure that it actually works.

This PR is part of a series to implement #1879.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -537,10 +537,23 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 		}
 	)
 
-	if err := json.NewDecoder(r.Body).Decode(&config); err != nil {
+	oldconfig := cluster.OldContainerConfig{
+		ContainerConfig: config,
+		Memory:          0,
+		MemorySwap:      0,
+		CPUShares:       0,
+		CPUSet:          "",
+	}
+
+	if err := json.NewDecoder(r.Body).Decode(&oldconfig); err != nil {
 		httpError(w, err.Error(), http.StatusBadRequest)
 		return
 	}
+
+	// make sure HostConfig fields are consolidated before creating container
+	cluster.ConsolidateResourceFields(&oldconfig)
+	config = oldconfig.ContainerConfig
+
 	// Pass auth information along if present
 	var authConfig *apitypes.AuthConfig
 	buf, err := base64.URLEncoding.DecodeString(r.Header.Get("X-Registry-Auth"))
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -394,7 +394,7 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		if !filters.MatchKVList("label", container.Config.Labels) {
 			continue
 		}
-		if !filters.Match("status", container.Info.State.StateString()) {
+		if !filters.Match("status", cluster.StateString(container.Info.State)) {
 			continue
 		}
 		if !filters.Match("node", container.Engine.Name) {
@@ -423,8 +423,8 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		candidates = candidates[:limit]
 	}
 
-	// Convert cluster.Container back into dockerclient.Container.
-	out := []*dockerclient.Container{}
+	// Convert cluster.Container back into apitypes.Container.
+	out := []*apitypes.Container{}
 	for _, container := range candidates {
 		if before != nil {
 			if container.ID == before.ID {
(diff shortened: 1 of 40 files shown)


MHBauer:
This was the first thing I hit, and I just nope'd out of there into easier territory. I had something similar to your updated `ContainerConfig`, so I think we were on the same path.

nishanttotla:
@MHBauer That sounds like my experience :smiley_cat:. I think this change is the single major one that will take the longest to get right, the rest will either depend on it, or are more isolated. The difficulty here is that it affects a large part of the code, so I'm facing merge conflicts all the time!

nishanttotla:
This PR is currently blocked by https://github.com/docker/engine-api/issues/171

nishanttotla:
After https://github.com/samalba/dockerclient/pull/236 is merged, `State` and `StateString` functions in this PR need to be updated.

nishanttotla:
Ping @docker/core-swarm-maintainers this PR is almost ready, please take a look.

Only some mesos tests are failing, still investigating.

api/handlers.go
```diff
@@ -526,26 +526,29 @@ func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
...
-	var authConfig *dockerclient.AuthConfig
+	var authConfig *apitypes.AuthConfig
 	buf, err := base64.URLEncoding.DecodeString(r.Header.Get("X-Registry-Auth"))
 	if err == nil {
-		authConfig = &dockerclient.AuthConfig{}
+		authConfig = &apitypes.AuthConfig{}
 		json.Unmarshal(buf, authConfig)
 	}
-	containerConfig := cluster.BuildContainerConfig(config)
+	containerConfig := cluster.BuildContainerConfig(config.Config, config.HostConfig, config.NetworkingConfig)
```
vieux: why not just pass `config` here, instead of `config.X, config.Y, config.Z` ?

nishanttotla: I think it made sense earlier when the config struct looked different compared to what it is right now. I should be possible to pass a single config object now. I'm wondering if I should fix that in a follow-up PR though, since the current one will heavily change because of this.

nishanttotla:
Update: The mesos compose tests fail because incorrect json is passed to the container create function.
```
{
  "Tty": false,
  "AttachStderr": false,
  "NetworkDisabled": false,
  "Image": "busybox",
  "Cmd": [
    "sleep",
    "100"
  ],
  "StdinOnce": false,
  "HostConfig": {
    "NetworkMode": "default",
    "Links": [
      
    ],
    "PortBindings": {
      
    },
    "Binds": [
      
    ],
    "LogConfig": {
      "Type": "",
      "Config": {
        
      }
    },
    "VolumesFrom": [
      
    ]
  },
  "Labels": {
    "com.docker.compose.service": "service2",
    "com.docker.compose.project": "compose",
    "com.docker.compose.config-hash": "95969bc3a801ccd851fb65910210c86676b691778716892aa2e57f3e2e593cbd",
    "com.docker.compose.version": "1.6.2",
    "com.docker.compose.oneoff": "False",
    "com.docker.compose.container-number": "1"
  },
  "AttachStdin": false,
  "Env": [
    
  ],
  "CpuShares": 1,
  "AttachStdout": false,
  "OpenStdin": false
}
```
The `CpuShares` should be nested inside `HostConfig` and not outside. I'm figuring out why this is the case.

vieux:
@nishanttotla maybe @dnephin knows why ?

dnephin:
I've created https://github.com/docker/docker-py/issues/1030 and https://github.com/docker/compose/issues/3284

nishanttotla:
Thanks @dnephin. We'll handle the old structs in Swarm for now, until the fix for compose is done.

dongluochen:
LGTM

vieux:
LGTM

MHBauer:
Wow, I didn't realize this was completed. Nice job!

nishanttotla:
@MHBauer thanks! Not much more left now!

PR #1975, opened 2016-03-15
Updating the documentation

moxiegirl:
@vieux Fix of schedule docs.

Signed-off-by: Mary Anthony <mary@docker.com>

PR #1984, opened 2016-03-17
fix duplicate rescheduling.md

vieux:
follow up of https://github.com/docker/swarm/pull/1975

The file is now at `docs/scheduler/rescheduling.md`


dongluochen:
LGTM

PR #1993, opened 2016-03-18
use travis build matrix for different GOOS

MHBauer:
- run individual builds based on GOOS
- only run the tests on the linux build

I got travis set up personally, and decided to take a look at how travis worked.  I thought this was an interesting result, but setting GOOS required a trick. I also figured we could avoid running the tests on the non-linuxs builds. 

I'm not sure effect this has on the badge for the github checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,11 @@ language: go
 go:
   - 1.5.3
 
+env:
+  - BUILD_GOOS=darwin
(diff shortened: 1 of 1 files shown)


MHBauer:
Similar things could be done by using matrix forms for multiple versions of go. Build and test with 1.5, 1.6, & tip, perhaps.

PR #1996, opened 2016-03-18
Plumbing context through all engine-api calls

nishanttotla:
Method signatures were updated to uniformly pass a `net/context.Context` struct for each method (https://github.com/docker/engine-api/pull/140). We should adopt this right now, so that there are fewer issues down the road.


nishanttotla:
Seems like I had a godep issue deleting some files, so build is failing. Will fix this shortly.

MHBauer:
I find it highly amusing that most of the changes are in the fake clients.

I still do not understand godep, so I can't help you there. I'll attempt to give this a build locally.

nishanttotla:
@MHBauer yeah, it's unfortunate that we have to maintain the fake clients, but that's the simplest course of action right now.

I know what the issue is, so your local build may not work either. Some files are being deleted from Godeps which I'm trying to fix.

MHBauer:
Shockingly, it does build and give me a binary. I am missing the files that are shown in the build errors. Like I said, I do not get it.

nishanttotla:
@MHBauer actually can you try building it locally? I found out that all packages that weren't being used got removed by `godep update`, which is really strange, and that is causing the build to fail. For example, `engine-api/client/transport` was removed, even though `engine-api/client` is part of the manifest. I don't understand that behavior.

nishanttotla:
Oh wow, I commented at around the same time. Okay so that's why my local builds passed. Wonder why Travis is complaining then.

nishanttotla:
@MHBauer I think I figured it out, it's this: https://github.com/tools/godep#the-deprecated--r-flag

nishanttotla:
Closing PR since `godep` issues need to be resolved first.

MHBauer:
given vendor support in 1.6, should we both upgrade to 1.6 and migrate from `Godep/_workspace` to `vendor`? Maybe discuss in an issue, perhaps?

nishanttotla:
@MHBauer I think we should in the near future.

PR #1997, opened 2016-03-18
Plumbing context through all engine-api calls

nishanttotla:
Method signatures were updated to uniformly pass a net/context.Context struct for each method (https://github.com/docker/engine-api/pull/140). We should adopt this right now, so that there are fewer issues down the road.

It took some work to get the Godeps right, because something is seriously wrong with handling of nested dependencies, such as in https://github.com/tools/godep/issues/400 (due to this #1996 was failing). I think now Godeps needs for all sub-packages to be explicitly imported. For now I fixed this by manually deleting the offending package (`engine-api`) from the `Godeps/` directory, then running `godep restore`, followed by pulling the latest `engine-api`, then `godep save ./...`.


dongluochen:
CI is failing. Please take a look. 

```
--- FAIL: TestMock (0.00s)
panic: 
mock: Unexpected Method Call
-----------------------------
ServerVersion(*context.emptyCtx)
		0: (*context.emptyCtx)(0xc8200156d0)
The closest call I have is: 
ServerVersion()
```

nishanttotla:
@dongluochen it should be fixed now.
@MHBauer please see this instead :)

MHBauer:
Builds and tests fine for me. I have go1.6 installed for other projects, so I get this nice warning:
```
godep: WARNING: Go version (go1.6) & $GO15VENDOREXPERIMENT= wants to enable the vendor experiment, but disabling because a Godep workspace (Godeps/_workspace) exists
```
It looks like other dependencies besides engine-api were also pulled in or updated. It also seems bizarre that individual packages in a project are being listed explicitly. Probably better to move from godep to vendor sooner rather than later.

nishanttotla:
@MHBauer only `libkv` was updated, I had to do that because `godep` would delete files and cause issues. But you're right, best to move away from Godep ASAP.

nishanttotla:
Test failures seem unrelated. Ping @dongluochen @vieux

vieux:
LGTM

dongluochen:
LGTM

PR #2004, opened 2016-03-22
Skip redundant endpoints in "network inspect"

### LINKED ISSUE #1969: Skip redundant endpoints in "network inspect"
With https://github.com/docker/docker/pull/21160, docker engine will return all the endpoints (local, remote & stale endpoints). The remote and stale endpoints will have be represented as `ep-{endpoint-id}` . Example output from a host that doesn't own this endpoint (which represents the remote container).

```
$ sudo docker network inspect test2
[
    {
        "Name": "test2",
        "Id": "eaa6a4786dcba692bf240f4199848d565b2948c0481d1987051d5ce4eef080b6",
        "Scope": "global",
        "Driver": "overlay",
        "EnableIPv6": false,
        "IPAM": {
            "Driver": "default",
            "Options": null,
            "Config": [
                {
                    "Subnet": "10.0.1.0/24",
                    "Gateway": "10.0.1.1/24"
                }
            ]
        },
        "Internal": false,
        "Containers": {
            "ep-84dcacdbed8ed7fcd2004d43612bb5bba91aceaf12c77ba6943e2c075e9d22dd": {
                "Name": "ep3",
                "EndpointID": "84dcacdbed8ed7fcd2004d43612bb5bba91aceaf12c77ba6943e2c075e9d22dd",
                "MacAddress": "02:42:0a:00:01:02",
                "IPv4Address": "10.0.1.2/24",
                "IPv6Address": ""
            }
        },
        "Options": {}
    }
]
```

This would cause duplicate endpoints in the `docker network inspect` output when executed in a swarm environment.

The fix is to skip such duplicate endpoints. But make sure the stale endpoints (which will not have a duplicate endpoint-id) should be returned.

dongluochen:
Fix #1969. 

### DIFF
--- a/cluster/network.go
+++ b/cluster/network.go
@@ -78,6 +78,34 @@ func (networks Networks) Filter(names []string, ids []string, types []string) Ne
 	return out
 }
 
+// RemoveDuplicateEndpoints returns a copy of input network
+// where duplicate container endpoints in the network are removed.
+// See https://github.com/docker/swarm/issues/1969
+// This function should be disabled when the bug is fixed in Docker network
+func (network *Network) RemoveDuplicateEndpoints() *Network {
+	// build a map from endpointID -> endpointIndex
+	endpointMap := make(map[string]string)
+	// traverse the endpoints to find the correct endpointIndex for each endpointID
+	for endpointIndex, endpointResource := range network.NetworkResource.Containers {
+		endpointID := endpointResource.EndpointID
+		// if this endpointID doesn't exist yet, add it
+		// if this endpointID exists, but endpointIndex is not a duplicate, use
+		// this endpointIndex
+		if _, ok := endpointMap[endpointID]; !ok || !strings.HasPrefix(endpointIndex, "ep-") {
+			endpointMap[endpointID] = endpointIndex
+		}
+	}
+	// Make a copy of the network
+	netCopy := *network
+	// clean up existing endpoints
+	netCopy.Containers = make(map[string]dockerclient.EndpointResource)
+	// add the endpoint index from endpointMap
+	for _, index := range endpointMap {
+		netCopy.Containers[index] = network.Containers[index]
+	}
+	return &netCopy
+}
+
 // Get returns a network using its ID or Name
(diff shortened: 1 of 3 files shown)


vieux:
LGTM

cluster/network.go
```diff
@@ -78,6 +78,34 @@ func (networks Networks) Filter(names []string, ids []string, types []string) Ne
...
+func (network *Network) RemoveDuplicateEndpoints() *Network {
+	// build a map from endpointID -> endpointIndex
+	endpointMap := make(map[string]string)
+	// traverse the endpoints to find the correct endpointIndex for each endpointID
+	for endpointIndex, endpointResource := range network.NetworkResource.Containers {
+		endpointID := endpointResource.EndpointID
+		// if this endpointID doesn't exist yet, add it
+		// if this endpointID exists, but endpointIndex is not a duplicate, use
+		// this endpointIndex
+		if _, ok := endpointMap[endpointID]; !ok || !strings.HasPrefix(endpointIndex, "ep-") {
```
nishanttotla: @dongluochen I'm trying to understand how `!strings.HasPrefix(endpointIndex, "ep-")` is a check for duplicates, as you wrote in the comment above. Can you please clarify?

nishanttotla: Actually, never mind, I read through the issue in more detail, and it makes sense.

nishanttotla:
LGTM

I don't expect the integration tests with `engine-master` to fail because of this change.

cluster/network.go
mavenugo: @dongluochen it would better to check if the `endpointIndex` contains `endpointID` instead of checking for `ep-`. We added the `ep-` prefix mostly for better UX. but for programmatic checks, it is better to check for `endpointID` part of the index.

nishanttotla: If that's the case, then I agree with @mavenugo. We shouldn't rely on the prefix here.

dongluochen: Thanks @mavenugo @nishanttotla . Update the duplicate/stale endpoint identification logic.

PR #2025, opened 2016-03-25
enable rescheduling watchdog only when primary

### LINKED ISSUE #1810: Can't access rescheduled container using its name after failed node recovers
This is about the experimental rescheduling support (#1488).

## What I do

- start a resilient container:
  ```
  docker run -d --name highlander -e reschedule:on-node-failure redis
  ```

- reboot the node of this container:
  ```
  NODE=$(docker inspect --format '{{.Node.Name}}' highlander)
  ssh $NODE sudo reboot
  ```

- wait until the node is done rebooting...

- try to access the container using its name, `highlander`:
  ```
  docker inspect highlander
  ```

## What I expect

Information about the container

## What I see instead

```
docker inspect highlander
[]
Error: No such image or container: highlander
```

## Discussion

Of course, when the node comes back, we end up with two `highlander` containers, and this results into a name clash. Possible workarounds:
- Swarm could just return *all* matching containers
- Swarm could return an error indicating the conflict
- When the node rejoins, Swarm could rename the clashing containers (e.g. `highlander` would become `highlander_conflicts_with_329bcf928295219e9f9cbe...`)

One word of warning about the last option: this could cause problems with the `swarm-manager` containers (i.e. they could end up clashing too, and just doing a special case with the name of the container (e.g. "name starts with swarm") wouldn't be enough since that could be a problem for people using custom Swarm images...)

vieux:
Close #1810

### DIFF
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -148,7 +148,7 @@ func setupReplication(c *cli.Context, cluster cluster.Cluster, server *api.Serve
 
 	go func() {
 		for {
-			run(candidate, server, primary, replica)
+			run(cluster, candidate, server, primary, replica)
 			time.Sleep(defaultRecoverTime)
 		}
 	}()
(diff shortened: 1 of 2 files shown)


vieux:
thanks @jpetazzo

vieux:
ping @dongluochen

jimmyxian:
LGTM

cli/manage.go
```diff
@@ -327,6 +330,5 @@ func manage(c *cli.Context) {
 		server.SetHandler(api.NewPrimary(cl, tlsConfig, &statusHandler{cl, nil, nil}, c.GlobalBool("debug"), c.Bool("cors")))
```
dongluochen: Watchdog is not created in this path.

vieux: good catch

vieux:
PTAL

dongluochen:
New test is failing on engine-1.9 and 1.10.
```
not ok 24 leader election - rescheduling
# (from function `retry' in file ./helpers.bash, line 80,
#  in test file ./replication.bats, line 137)
#   `retry 5 1 containerRunning "c1" "node-1"' failed
```

Old test fails on engine-master. 
```
not ok 29 rescheduling node comes back
# (from function `retry' in file ./helpers.bash, line 80,
#  in test file ./rescheduling.bats, line 189)
#   `retry 5 1 containerRunning "c1" "node-0"' failed
```

dongluochen:
LGTM

jpetazzo:
Tested, and it works. Thanks!

Issue #2027, opened 2016-03-25
Certificate issues in current master branch (HEAD@510ea94)


echupriyanov:
I’ve tried to run swarm manager built from master branch (HEAD@510ea94)
And I see following certificate issues in logs:
```
INFO[0000] Listening for HTTP                            addr=0.0.0.0:4000 proto=tcp
INFO[0000] Leader Election: Cluster leadership lost
INFO[0000] Leader Election: Cluster leadership acquired
DEBU[0000] Watch triggered with 4 nodes                  discovery=etcd
DEBU[0000] Failed to validate pending node: An error occurred trying to connect: Get https://swarm-node-04a.cybertonica.aws:2376/info: x509: certificate is valid for swarm-node-04a.cybertonica.aws, not swarm-node-05a.cybertonica.aws  Addr=swarm-node-04a.cybertonica.aws:2376
DEBU[0000] Failed to validate pending node: An error occurred trying to connect: Get https://swarm-node-03a.cybertonica.aws:2376/info: x509: certificate is valid for swarm-node-03a.cybertonica.aws, not swarm-node-05a.cybertonica.aws  Addr=swarm-node-03a.cybertonica.aws:2376
DEBU[0000] Failed to validate pending node: An error occurred trying to connect: Get https://swarm-node-06a.cybertonica.aws:2376/info: x509: certificate is valid for swarm-node-06a.cybertonica.aws, not swarm-node-05a.cybertonica.aws  Addr=swarm-node-06a.cybertonica.aws:2376
DEBU[0000] Start monitoring events                       id=MGVW:TYQU:XFOD:3PAF:POTR:NEZ4:NTG4:JEMN:FH4U:THWW:FLLY:OXGR name=swarm-node-05a.cybertonica.aws
INFO[0000] Registered Engine swarm-node-05a.cybertonica.aws at swarm-node-05a.cybertonica.aws:2376
```

Same error appears in build from v1.2.0-rc1 tag.
Build from tag v1.1.3 has no such errors.

jpetazzo:
Confirmed -- just ran into this exact same issue on my test clusters. I use Consul for discovery (not that it should matter).

My setup works with 1.1.3-rc2 (7e9c6bd) but not with master (7e9c6bd).

diogomonica:
What I don't understand here is this:

[...]

PR #2029, opened 2016-03-25
Prevent adding ServerName to TLSConfig (Issue #2027)

echupriyanov:
Quick fix for #2027 

Signed-off-by: Eugene Chupriyanov <tchu@tchu.ru>

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -279,7 +279,15 @@ func (c *Cluster) addEngine(addr string) bool {
 func (c *Cluster) validatePendingEngine(engine *cluster.Engine) bool {
 	// Attempt a connection to the engine. Since this is slow, don't get a hold
 	// of the lock yet.
-	if err := engine.Connect(c.TLSConfig); err != nil {
+	//
+	// call to engine.Connect() spoils cluster-wide c.TLSConfig,
+	// so create a separate TLSConfig for each engine
+	var config *tls.Config
+	if c.TLSConfig != nil {
+		config := &tls.Config{}
+		*config = *c.TLSConfig
+	}
+	if err := engine.Connect(config); err != nil {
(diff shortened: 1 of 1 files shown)


vieux:
@echupriyanov we vendor the engine-api code.

Can you please open your PR upstream in https://github.com/docker/engine-api

echupriyanov:
Will do. Frankly, I opened it here mostly to illustrate the cause of the problem. Maybe there is a better place to handle it? Another way is to clear `TLSConfig.ServerName` after each `engine.Connect()` call. That way we don;t depend on [engine-api](https://github.com/docker/engine-api)

diogomonica:
This should be fine. IIRC, hostname will be taken from the server's address if none is given in the `tls.Config`.

The only place where I know playing with `ServerName` might be an issue is when creating a `tls.Client` directly, in which case not passing a `ServerName` might mean we're simply verifying the validity of the certificate chain.

jpetazzo:
(Confirming that cherrypicking this solves #2027 for me)

vieux:
@echupriyanov did you open the PR on engine-api ?

echupriyanov:
Sorry, was busy at primary job. Will do now.

echupriyanov:
Added PR (https://github.com/docker/engine-api/pull/173)

echupriyanov:
Looks like my PR in docker/engine-api is not accepted. Another option I see is to modify `validatePendingEngine()` call to create a copy of `c.TLSConfig` and pass it to `engine.Connect()`.
I will try it tomorrow.

echupriyanov:
@vieux I’m not sure my PR https://github.com/docker/engine-api/pull/173 will be accepted.
So here is another way to resolve certificate/ServerName problem.

echupriyanov:
This PR is obsoleted by https://github.com/docker/engine-api/pull/179

dongluochen:
Thank @echupriyanov for reporting the problem and helping with the fix.

PR #2037, opened 2016-03-27
typo, the port of consul is 8500

CodeJuan:
The command should be same as `$ docker run -d -p 4000:4000 swarm manage -H :4000 --replication --advertise 172.30.0.161:4000 consul://172.30.0.161:8500` in **Step 5. Create Swarm cluster**

### DIFF
--- a/docs/install-manual.md
+++ b/docs/install-manual.md
@@ -279,7 +279,7 @@ replica.
 
 4. Start the Swarm manager. For example:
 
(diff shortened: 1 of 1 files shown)


jimmyxian:
Good catch!
LGTM
Thanks @CodeJuan

nishanttotla:
LGTM.

Thanks @CodeJuan! I wonder how we missed this :open_mouth:

Issue #2039, opened 2016-03-28
API "/images/create" is available, and the document swarm-api is out of date


CodeJuan:
The document [swarm-api](https://github.com/docker/swarm/blob/master/docs/swarm-api.md#missing-endpoints) said the API "/images/create" is not available, but I found that the API "/images/create" is OK, I tried 
```
curl -X POST http://192.168.1.245:4000/images/create?fromImage=busybox
```
And the response is 
```
{"id":"i3","status":"Pulling busybox...","progressDetail":{}}
{"id":"g540","status":"Pulling busybox...","progressDetail":{}}
{"id":"i3","status":"Pulling busybox... : downloaded","progressDetail":{}}
{"id":"g540","status":"Pulling busybox... : downloaded","progressDetail":{}}
```
After that, I checked the images on the nodes by cli `docker images`, and found the image busybox.

So I think the  document [swarm-api](https://github.com/docker/swarm/blob/master/docs/swarm-api.md#missing-endpoints) is out of date.

nishanttotla:
cc @thaJeztah

Issue #1312, opened 2015-10-19
Keep track of the missing API endpoints options in swarm


jimmyxian:
I have compared the latest docker API. Our Swarm is missing the following api OPTIONS:
#### push
- [ ] --disable-content-trust

#### build
- [ ] --disable-content-trust
- [ ] --ulimit=[]
- [x] --build-arg

#### pull 
- [ ] --disable-content-trust
- [ ] --all-tags

#### run 
- [ ] --disable-content-trust
- [x] --stop-signal
- [x] --dns-opt=[] 
- [x] --kernel-memory=
- [x] --group-add=[]
- [x] --memory-swappiness=-1
- [x] --memory-reservation

#### ps
- [ ] --size
- [ ] --filter since=XXX
- [ ] --filter ancestor=\<image\>
- [ ] --filter label=\<somelabel\>
- [x] --filter volume=XXXX

#### load
- [ ] --quiet

#### import
- [ ] --message

#### exec
- [ ] --privileged

jimmyxian:
@vieux I just have done some `run` options, not all. :). Anyway, i will continue to work on this.

jimmyxian:
@abronan Closed?

vieux:
@jimmyxian you wrote "Fix #1312" in the PR #1318 so it closed this one automatically, it's not @abronan he just merged #1318.

I usually use "Related to #xxx" when It doesn't fix entirely the issue.

jimmyxian:
@vieux :laughing: Got it.

nishanttotla:
@jimmyxian would you mind updating this issue? :)

It seems like there are newer options that we don't support with newer releases. We should also compile how our API endpoints differ from those in the upcoming Engine release. Ping @dongluochen .

jimmyxian:
@nishanttotla OK, I will compare the newer options and update this ASAP

jimmyxian:
Replaced by https://github.com/docker/swarm/issues/2333

PR #2040, opened 2016-03-28
the API /images/create is available

CodeJuan:
As I mentioned in https://github.com/docker/swarm/issues/2039


thaJeztah:
@nishanttotla do we have another endpoint that is *not* present in the Swarm API that we can use as a replacement for this example?

dongluochen:
@thaJeztah See https://github.com/docker/swarm/issues/1312

PR #2050, opened 2016-03-30
Fix Consul Leader Election Failure on multi-server and leader soft-restart

abronan:
This fixes an error with Consul which fails to see a new
swarm leader after a consul leader soft restart because
of session renewal and missing status update on leadership

Fixes #1782 

Depends on docker/libkv#120 and docker/leadership#2

Needs more testing with other stores.

/cc @vieux @jpetazzo

Signed-off-by: Alexandre Beslic <alexandre.beslic@gmail.com>

PR #2056, opened 2016-03-30
Moving Image handling to engine-api

nishanttotla:
This change moves over all image handling to `engine-api`. It's self-contained, and relatively small.

It's part of implementing the proposal #1879.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -185,7 +185,7 @@ func getImagesJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	accepteds := filters.Get("node")
 	// this struct helps grouping images
 	// but still keeps their Engine infos as an array.
-	groupImages := make(map[string]dockerclient.Image)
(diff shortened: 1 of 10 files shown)


nishanttotla:
Ping @vieux @dongluochen @abronan

nishanttotla:
Thanks to @dongluochen, we figured out that this PR actually fixes a race condition that's been causing issues in #1983.

abronan:
@nishanttotla leader election tests are fixed with #2050 so this should be good.

vieux:
LGTM

dongluochen:
LGTM

MHBauer:
That's amazing. Switching to engine-api actually fixes a bug? Great!

nishanttotla:
@MHBauer a bug that likely came in because we have two clients, so ¯\_(ツ)_/¯

That said, turns out we were wrong earlier. The bug is not really a race condition, but an improper use of the `engine-api`, so I'll try to fix it separately.

PR #2087, opened 2016-04-06
Remove /containers/ps from the api router

HackToday:
Docker itself have removed this api since 1.9.0, so it should same
for swarm to remove this old api.

Signed-off-by: Kai Qiang Wu(Kennan) <wkqwu@cn.ibm.com>

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -324,7 +324,6 @@ func getVolumes(c *context, w http.ResponseWriter, r *http.Request) {
 	json.NewEncoder(w).Encode(volumes)
 }
 
-// GET /containers/ps
 // GET /containers/json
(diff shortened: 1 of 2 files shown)


vieux:
@HackToday swarm still supports docker 1.6.x so let's not remove it for now.

HackToday:
Thanks @vieux  
One question about, where can find which version is supported in swarm ?

vieux:
@HackToday here: https://github.com/docker/swarm/blob/master/cluster/engine.go#L33

HackToday:
Thanks. I expected we could find it on our official docs, 
IMO just found it in code seems not good idea :0

PR #2107, opened 2016-04-11
fix the bug that connection cannot be closed when the client close request to /containers/{stopped container id}/stats

chengjingtao:
`Bug description`:
when client connect to swarm ,and request to `/containers/{ id of stopped container }/stats` api .the docker server didn't write any data in body to swarm. But when we close the request of client, the connection of swarm and docker server was not be closed.

I have try to fix the bug .

`Reproduce`:
As I say ,you can reproduce the bug : 
1. Join your docker host into swarm
2. Run a container and stop it  on your host (Don't forget to stop it !)
3. Using your browser, and request to  http://swarm-address/containers/{container you just stopped}/stats . and wait for a seconds and stop it .  
4. Do `step 3` for many times and exec `sudo lsof -i:2375` ,  you will see the connection created again and again.  

Signed-off-by:  chengjt chengjt@csdn.net

### DIFF
--- a/api/utils.go
+++ b/api/utils.go
@@ -122,8 +122,8 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request, 
 			CancelRequest(*http.Request)
 		}
 
-		if closeNotifier, ok = w.(http.CloseNotifier); ok == false {
-			log.Error("ResponseWriter was not support CloseNotify")
+		if closeNotifier, ok = w.(http.CloseNotifier); !ok {
+			log.Error("ResponseWriter does not support CloseNotify")
 			return
 		}
 
@@ -131,8 +131,8 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request, 
 		case <-closeNotifier.CloseNotify():
 			{
 				var cancel canceler
-				if cancel, ok = client.Transport.(canceler); ok == false {
-					log.Error("client.Transport was not support CancelRequest")
+				if cancel, ok = client.Transport.(canceler); !ok {
+					log.Error("client.Transport does not support CancelRequest")
 					return
 				}
 
--- a/api/utils.go
+++ b/api/utils.go
@@ -105,12 +105,12 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request, 
 	log.WithFields(log.Fields{"method": r.Method, "url": r.URL}).Debug("Proxy request")
 
 	var (
-		writeCompleted  = make(chan bool)
-		dectedCompleted = make(chan bool)
+		writeCompleted    = make(chan bool)
+		detectedCompleted = make(chan bool)
 	)
 
-    //if docker server did not write any data in body ( just like request /containers/xxx/stats to stopped container) , `io.Copy(NewWriteFlusher(w), resp.Body)` will wait for reading data.
-    //But the client that connect to swarm may close the request . So we should dected it and  cancel the request to docker server.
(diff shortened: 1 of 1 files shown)


MHBauer:
quick notes.
1. gofmt 
2. change all instances of `dected` to `detected`

linecheng:
Ok

api/utils.go
```diff
@@ -103,6 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+			closeNotifier http.CloseNotifier
+			ok            bool
+		)
+
+		type canceler interface {
+			CancelRequest(*http.Request)
+		}
+
+		if closeNotifier, ok = w.(http.CloseNotifier); ok == false {
+			log.Error("ResponseWriter was not support CloseNotify")
```
MHBauer: was -> does

api/utils.go
```diff
@@ -103,6 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+		var (
+			closeNotifier http.CloseNotifier
+			ok            bool
+		)
+
+		type canceler interface {
+			CancelRequest(*http.Request)
+		}
+
+		if closeNotifier, ok = w.(http.CloseNotifier); ok == false {
```
MHBauer: usually I see `!ok` instead of `ok == false`. Style thing, but I think `!ok` is more idiomatic.

api/utils.go
```diff
@@ -103,6 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+			log.Error("ResponseWriter was not support CloseNotify")
+			return
+		}
+
+		select {
+		case <-closeNotifier.CloseNotify():
+			{
+				var cancel canceler
+				if cancel, ok = client.Transport.(canceler); ok == false {
```
MHBauer: same as above. `!ok`.

api/utils.go
```diff
@@ -103,6 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+					log.Error("client.Transport was not support CancelRequest")
```
MHBauer: was -> does

api/utils.go
```diff
@@ -103,6 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+					return
+				}
+
+				cancel.CancelRequest(r)
```
MHBauer: Looking at golang docs I see: `Deprecated: Use Request.Cancel instead.`
Not sure how strict we need to be on enforcing this.
I think the replacement is `close(r.Cancel)`.

MHBauer:
Please squash the different commits after you make changes.

linecheng:
@MHBauer  thanks for your review.
I fixed some grammatical errors .
you are right , `Transport.CancelRequest(r)` is deprecated.
I have try to use "close(r.Cancel)"  //r.Cancel is receive only.
```

ch := make(chan struct{})
r.Cancel=ch
go func(){

     ...
     ...
     close(ch)   // cancel.CancelRequest(r)
     ...
     ...

}()

```

But  I find that my code dose not work  well suddenly
Neither using the old `cancel.CancelRequest(r)` nor `close(r.Cancel)` can cancel the request really, I'm confused about my previous tests.

I am not sure about the reason of that. It seams to be related to `client, scheme, err := engine.HTTPClientAndScheme()`. 
I am checking it and  will squash the different commits  after checking  .

Thanks again.

PR #2125, opened 2016-04-14
Describe CI jobs in a Jenkinsfile

mikedougherty:
This uses a new-ish feature of Jenkins called "pipelines" which allows you to define the desired CI behavior with a DSL in source control. I plan on incorporating (or helping others incorporate) this into other docker/ projects as well, but this is the first 'real' one.

This will also make it easier to add windows CI in a sane way (cc @friism)

@vieux

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,8 @@
+// Only run on Linux atm
+wrappedNode(label: 'docker') {
+  deleteDir()
+  stage "checkout"
+  checkout scm
+
+  documentationChecker("docs")
+}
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,34 @@
+stage "test"
+
+def testTask(imageName, imageTag) {
```
mikedougherty: FYI: this sort of large ugly groovy function we can 'hide away' in another file; either in another repo like I've done in my development for engine (see [here](https://github.com/docker/jenkins-pipeline-scripts/blob/master/src/com/docker/projects/Docker.groovy)) or we should also be able to keep it adjacent in this repo -- though I'll have to do some further investigation on that to be sure.

nishanttotla:
Closing due to lack of activity.

PR #2131, opened 2016-04-15
Fix bug: stop/restart/kill an inexisting container returns 500 rather than 404

jimmycmh:
Fix bug: stop/restart/kill an inexisting container returns 500 rather than 404


dongluochen:
LGTM

PR #2132, opened 2016-04-15
Moving last few image operations to engine-api

nishanttotla:
Addresses #1879

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -1015,10 +1015,26 @@ func (e *Engine) Pull(image string, authConfig *types.AuthConfig) error {
 
 // Load an image on the engine
 func (e *Engine) Load(reader io.Reader) error {
-	err := e.client.LoadImage(reader)
+	loadResponse, err := e.apiClient.ImageLoad(context.TODO(), reader, false)
 	e.CheckConnectionErr(err)
 	if err != nil {
 		return err
+	}
+
+	// wait until the image load is finished
+	dec := json.NewDecoder(loadResponse.Body)
+	m := map[string]interface{}{}
+	for {
+		if err := dec.Decode(&m); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return err
+		}
+	}
+	// if the final stream object contained an error, return it
+	if errMsg, ok := m["error"]; ok {
+		return fmt.Errorf("%v", errMsg)
 	}
 
 	// force fresh images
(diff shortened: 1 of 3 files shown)


dongluochen:
LGTM

cluster/engine.go
```diff
@@ -1015,12 +1015,28 @@ func (e *Engine) Pull(image string, authConfig *types.AuthConfig) error {
 
 // Load an image on the engine
 func (e *Engine) Load(reader io.Reader) error {
-	err := e.client.LoadImage(reader)
+	loadResponse, err := e.apiClient.ImageLoad(context.TODO(), reader, false)
```
abronan: Hmm can we replace those `context.TODO()` with a real context here...

cluster/engine.go
```diff
@@ -1029,7 +1045,13 @@ func (e *Engine) Load(reader io.Reader) error {
...
 // Import image
 func (e *Engine) Import(source string, repository string, tag string, imageReader io.Reader) error {
-	_, err := e.client.ImportImage(source, repository, tag, imageReader)
+	opts := types.ImageImportOptions{
+		SourceName:     source,
+		RepositoryName: repository,
+		Tag:            tag,
+		Source:         imageReader,
+	}
+	_, err := e.apiClient.ImageImport(context.TODO(), opts)
```
abronan: same here

cluster/engine.go
```diff
@@ -1247,7 +1269,13 @@ func (e *Engine) BuildImage(buildImage *types.ImageBuildOptions) (io.ReadCloser,
...
 func (e *Engine) TagImage(IDOrName string, repo string, tag string, force bool) error {
 	// send tag request to docker engine
-	err := e.client.TagImage(IDOrName, repo, tag, force)
+	opts := types.ImageTagOptions{
+		ImageID:        IDOrName,
+		RepositoryName: repo,
+		Tag:            tag,
+		Force:          force,
+	}
+	err := e.apiClient.ImageTag(context.TODO(), opts)
```
abronan: same

vieux: todo is everywhere...

cluster/engine.go
nishanttotla: We've been using `context.TODO()` for all `engine-api` calls so far. What do you recommend?

abronan: We can leave it as is for now until the integration with engine-api is over and we determine where the use of a context is critical, but we should use the contexts properly deriving a default context and adding timeouts when necessary or cancel the context under exceptional circumstances. There are no guarantee that the call to `ImageLoad` will ever return. And the empty context does not capture a timeout value.

abronan:
LGTM

cluster/engine.go
nishanttotla: @abronan I agree. Let's tackle that soon.

PR #2174, opened 2016-04-26
Moving to official consul image

nishanttotla:
There's now an official consul image: https://hub.docker.com/_/consul/

Let's see if our integration tests work with that.

Signed-off-by: Nishant Totla nishanttotla@gmail.com

### DIFF
--- a/test/integration/discovery/consul.bats
+++ b/test/integration/discovery/consul.bats
@@ -12,7 +12,7 @@ DISCOVERY="consul://${STORE_HOST}/test"
 CONTAINER_NAME=swarm_consul
 
 function start_store() {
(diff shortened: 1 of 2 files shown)


nishanttotla:
Nope, doesn't really seem to work, not yet sure why.

abronan:
@nishanttotla thought about doing the migration, yet I'm not sure this is quite "the same" image, might require some tweaks to make it work.

Logs are throwing: `cannot set or renew session for ttl, unable to operate on sessions`

My guess is that the config passed to the image (which shrinks the minimum ttl value to `1s`) is not interpreted and thus operations on session gets rejected because the minimum ttl must be `10s`. Will look into it and commit on top.

thaJeztah:
Yeah, looks like it has a really different entrypoint script, so guess it expects different arguments being passed

nishanttotla:
@abronan makes sense. Hopefully it's not too hard to do (the older image is fine too for now)

vieux:
Nice~

PR #2205, opened 2016-05-03
Containerslots

dongluochen:
Carry over #2137 from @xFragger.

### DIFF
--- a/scheduler/filter/slots.go
+++ b/scheduler/filter/slots.go
@@ -2,9 +2,10 @@ package filter
 
 import (
 	"errors"
+	"strconv"
+
 	"github.com/docker/swarm/cluster"
 	"github.com/docker/swarm/scheduler/node"
-	"strconv"
 )
 
 var (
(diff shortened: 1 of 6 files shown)


vieux:
LGTM

scheduler/filter/slots.go
```diff
@@ -0,0 +1,52 @@
...
+	if len(result) == 0 {
+		return nil, ErrNoNodeWithFreeSlotsAvailable
+	}
+
+	return result, nil
+}
+
+// GetFilters returns just the info that this node failed, because there where no free slots
+func (f *SlotsFilter) GetFilters(config *cluster.ContainerConfig) ([]string, error) {
+	return []string{"free slots"}, nil
```
nishanttotla: I'm not convinced by this. `GetFilters` is supposed to return an understandable description of the constraints, so that they can be printed in case they can't be satisfied, especially in conjunction with other constraints.

dongluochen: @nishanttotla I update it to "available container slots".

nishanttotla: Cool, we can keep that.

nishanttotla:
LGTM

nishanttotla:
🎊 🎊 @xFragger!

PR #2206, opened 2016-05-03
Creating http client in swarm

nishanttotla:
This is experimental, let's see if we can make it work. Had to temporarily change vendor, but will figure out an alternative if this is successful.

Addresses #1879.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -11,6 +11,7 @@ import (
 	"math/rand"
 	"net"
 	"net/http"
+	"net/url"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


nishanttotla:
Ping @dongluochen @vieux how do you feel about this? (test failure is unrelated -- leader election)

nishanttotla:
Ping @dongluochen @vieux @abronan any objections to merging this?

It does modify `vendor` and that's temporary because `samalba/dockerclient` is going away, so we shouldn't really add swarm specific stuff to it and vendor.

dongluochen:
LGTM

vieux:
LGTM

PR #2210, opened 2016-05-05
Do not allow double RLock() of RWMutex

### LINKED ISSUE #2208: swarm hangs in some cases
Hi guys!
we have problems with swarm it just hangs every 10-20 mins.

Output of ```docker version```:

```
docker -H localhost:23750 version
Client:
 Version:      1.8.2
 API version:  1.20
 Package Version: docker-1.8.2-7.el7.centos.x86_64
 Go version:   go1.4.2
 Git commit:   bb472f0/1.8.2
 Built:        
 OS/Arch:      linux/amd64

Server:
 Version:      swarm/1.2.1
 API version:  1.22
 Package Version: 
 Go version:   go1.6.2
 Git commit:   584e10c
 Built:        Sat Apr 23 05:13:23 UTC 2016
 OS/Arch:      linux/amd64
```
Output of ```docker info```:
```
docker -H localhost:23750 info
Containers: 0
Images: 27
Role: primary
Strategy: spread
Filters: health, port, dependency, affinity, constraint
Nodes: 2
 node1-intest.codenvy-stg.com: node1.intest.codenvy-stg.com:2375
  └ Status: Healthy
  └ Containers: 0
  └ Reserved CPUs: 0 / 4
  └ Reserved Memory: 0 B / 28.85 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.10.0-327.4.5.el7.x86_64, operatingsystem=CentOS Linux 7 (Core), storagedriver=devicemapper
  └ Error: (none)
  └ UpdatedAt: 2016-05-04T12:10:38Z
  └ ServerVersion: 1.8.2
 node2-intest: node2.intest.codenvy-stg.com:2375
  └ Status: Healthy
  └ Containers: 0
  └ Reserved CPUs: 0 / 4
  └ Reserved Memory: 0 B / 28.85 GiB
  └ Labels: executiondriver=native-0.2, kernelversion=3.10.0-327.4.5.el7.x86_64, operatingsystem=CentOS Linux 7 (Core), storagedriver=devicemapper
  └ Error: (none)
  └ UpdatedAt: 2016-05-04T12:10:25Z
  └ ServerVersion: 1.8.2
Kernel Version: 3.10.0-327.4.5.el7.x86_64
Operating System: linux
CPUs: 8
Total Memory: 57.7 GiB
Name: intest
```

**Additional environment details (AWS, VirtualBox, physical, etc.):**
We can reproduce this on aws, azure, VBox + vagrant
We have 2 nodes connected using discovery by file
We run swarm like this ```/usr/local/swarm/swarm --log-level debug manage -H 0.0.0.0:23750 file:///usr/local/swarm/node_list 2>> /var/log/swarm.log``` (no as container)

**Steps to reproduce:**
We still trying to reproduce it on plain swarm/docker without our application. For now we can only explain what is happening within integration. But we can describe what we actually do with our integration. So basically we invoke swarm API to launch containers and execute execs inside with logs streaming of execs. our use case looks like:

-  create container
-  start container
-  create exec
-  start exec (we stream exec logs here to the client)
-  stop container
-  remove container

While we continuously running that scenario for some amount of users like ~20 in same time, swam just hangs and became unresponsive it is not possible even execute ```curl localhost:23750/info``` it just hangs. **Only swarm restart helps** and it looks like some leak or lock.

Here is swarm log output in debug mode:
http://pastebin.com/raw/jrw1N93b

**Important**
Once swarm is hang it does not respond on any API call each request just freeze for a while but in swarm logs we can see that swarm accept those calls like
```time="2016-05-04T12:27:20Z" level=debug msg="HTTP request received" method=GET uri="/info"```

It looks like swarm become unresponsive when we perform stopping / starting containers

Please feel free to ask any questions.

### PR DESCRIPTION
**Pull request by dongluochen**
Do not allow double RLock() of RWMutex. It may lead to deadlock if there is Lock() call from another goroutine in between.

Fix #2208.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -760,10 +760,6 @@ func (c *Cluster) Container(IDOrName string) *cluster.Container {
 	if len(IDOrName) == 0 {
 		return nil
 	}
-
-	c.RLock()
-	defer c.RUnlock()
-
 	return c.Containers().Get(IDOrName)
 }
 
@@ -801,9 +797,9 @@ func (c *Cluster) listNodes() []*node.Node {
 	out := make([]*node.Node, 0, len(c.engines))
 	for _, e := range c.engines {
 		node := node.NewNode(e)
-		for _, c := range c.pendingContainers {
-			if c.Engine.ID == e.ID && node.Container(c.Config.SwarmID()) == nil {
-				node.AddContainer(c.ToContainer())
+		for _, pc := range c.pendingContainers {
+			if pc.Engine.ID == e.ID && node.Container(pc.Config.SwarmID()) == nil {
+				node.AddContainer(pc.ToContainer())
 			}
 		}
 		out = append(out, node)
@@ -897,9 +893,6 @@ func (c *Cluster) RANDOMENGINE() (*cluster.Engine, error) {
 
 // RenameContainer rename a container
 func (c *Cluster) RenameContainer(container *cluster.Container, newName string) error {
-	c.RLock()
-	defer c.RUnlock()
-
 	// check new name whether available
 	if !c.checkNameUniqueness(newName) {
 		return fmt.Errorf("Conflict: The name %s is already assigned. You have to delete (or rename) that container to be able to assign %s to a container again.", newName, newName)

### CODE REVIEW

**Comment by nishanttotla** [comment]
LGTM

Ping @vieux @abronan let's merge this ASAP

**Comment by nishanttotla** [comment]
Also, we could have similar problems in other parts of the code, I'm looking at tools to auto-detect double locking like this.

File: cluster/swarm/cluster.go [hunk 1]
```diff
@@ -801,9 +797,9 @@ func (c *Cluster) listNodes() []*node.Node {
 	out := make([]*node.Node, 0, len(c.engines))
 	for _, e := range c.engines {
 		node := node.NewNode(e)
-		for _, c := range c.pendingContainers {
-			if c.Engine.ID == e.ID && node.Container(c.Config.SwarmID()) == nil {
-				node.AddContainer(c.ToContainer())
+		for _, pc := range c.pendingContainers {
```
MHBauer: good rename so we don't shadow the cluster `c`.

cluster/swarm/cluster.go
```diff
@@ -897,9 +893,6 @@ func (c *Cluster) RANDOMENGINE() (*cluster.Engine, error) {
 
 // RenameContainer rename a container
 func (c *Cluster) RenameContainer(container *cluster.Container, newName string) error {
-	c.RLock()
```
vieux: why did you remove this one ?

vieux:
I asked a question, but besides this: LGTM

cluster/swarm/cluster.go
dongluochen: In `c.checkNameUniqueness`, there is `c.RLock()`.

nishanttotla: @vieux in the call to `checkNameUniqueness`, we acquire that lock again
```go
func (c *Cluster) checkNameUniqueness(name string) bool {
	// Abort immediately if the name is empty.
	if len(name) == 0 {
		return true
	}

	c.RLock()
	defer c.RUnlock()
	for _, e := range c.engines {
        ...
```

vieux: ok thanks.

PR #2230, opened 2016-05-11
install-manual: fix sec. swarm manager connect

kolyshkin:
Apparently, without -p 4000:4000 the secondary swarm manager would be inaccessible, basically rendering it useless.

Signed-off-by: Kir Kolyshkin <kir@openvz.org>


dongluochen:
LGTM

Issue #2149, opened 2016-04-20
swarm rescheduling doesnt work with overlay networks


slawekgh:
I have 2 swarm nodes SwarmNode#1  and SwarmNode#2 
I created 8 containers , 4 of them with regular network, 4 of them with overlay network 
next I simulated downtime for SwarmNode#1 

after that swarm rescheduled only these containers which dont have overlay network 

for those which has overlay there is log in swarm manage:

Failed to start rescheduled container XXXXX: 500 Internal Server Error: service endpoint with name ubuntu07 already exists
Failed to start rescheduled container XXXXX: 500 Internal Server Error: service endpoint with name ubuntu06 already exists

doronp:
@slawekgh I would like to have a look. Can you please provide Steps to reproduce, Docker Swarm, versions etc (regular detailed bug report) .

From what you have included my guess would be that in the Engine the network ubuntu07 (it is a network right?) is being recreated when it shouldn't be...

I would assume that you actually created the networks along with the containers and that Swarm is actually redoing the create along with the network.

Any way I will need your inputs to be certain and on the same page and will fix it when in a few if needed.
Cheers,
Doronp

slawekgh:
I have 3 nodes:
gmbNode01 - swarm master + consul 
gmbNode02 - swarm node #1
gmbNode03 - swarm node #2

gmbNode01 - 10.3.0.10
gmbNode02 - 10.3.0.11
gmbNode03 - 10.3.0.12

=====================================
docker info for all 3 nodes:

root@gmbNode01:~# docker info
Containers: 13
 Running: 11
 Paused: 0
 Stopped: 2
Images: 13
Server Version: 1.11.0
Storage Driver: aufs
 Root Dir: /var/lib/docker/aufs
 Backing Filesystem: extfs
 Dirs: 97
 Dirperm1 Supported: true
Logging Driver: json-file
Cgroup Driver: cgroupfs
Plugins:
 Volume: local
 Network: bridge null host
Kernel Version: 4.2.0-23-generic
Operating System: Ubuntu 15.10
OSType: linux
Architecture: x86_64
CPUs: 1
Total Memory: 1.636 GiB
Name: gmbNode01
ID: xxx
Docker Root Dir: /var/lib/docker
Debug mode (client): false
Debug mode (server): false
Registry: https://index.docker.io/v1/
WARNING: No swap limit support

=====
root@gmbNode02:~# docker info
Containers: 1
 Running: 1
 Paused: 0
[...]

PR #2250, opened 2016-05-18
Solve #2149 until the root cause solution is done

doronp:
Signed-off-by: Doron Podoleanu doronp@il.ibm.com


dongluochen:
Thanks for your effort! However I think we should have better understanding the problem, even for a workaround. Let's continue the discussion on #2149.

PR #2271, opened 2016-05-24
integration: increase retry for leader election store failure test

### LINKED ISSUE #2266: Builds fail randomly on test "28 leader election - store failure"
The build fails randomly on test "28 leader election - store failure"
on one of the targets ( engine-1.10  engine-1.10  engine-1.10  engine-1.10 engine-master)

The fail is random: E.g:  only engine-1.10  fails , and then on the next build (exactly the same code !) engine-1.10 will pass and engine-master will fail . On the same test!

abronan:
Can't reproduce the test failure locally so I just suspect slow CI and a too aggressive retry on store restart.

Finding a sensible retry time is a bit tricky not to make the tests take much more time to complete but this doubles the time waiting for the store to be available.

/cc @dongluochen @nishanttotla 

Fix #2266 

### DIFF
--- a/test/integration/replication.bats
+++ b/test/integration/replication.bats
@@ -35,7 +35,8 @@ function start_store_cluster() {
 }
 
 function restart_leader() {
-	# TODO find out who is the leader
+	# TODO at this point we know the leader is node_1 but the function would be more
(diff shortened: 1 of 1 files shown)


nishanttotla:
LGTM

PR #2276, opened 2016-05-25
Remove unneeded Container() interface

resouer:
It's wired that `Container()` interface is duplicated with `Containers().Get()`, this will fix that.

Signed-off-by: Harry Zhang <harryz@hyper.sh>


xiaods:
what.. what purpose on the Container() interface? anyone can input some advise?

PR #2340, opened 2016-06-08
[WIP] [RFC] Experimental implementation of using an external scheduler

mikesimons:
This is a scheduler that uses an external service to determine placement. It presently marshals the scheduling data in to a simplified form (the default form is way too verbose to use verbatim) and encodes this as JSON. This JSON is then POST-ed to the service and decodes the response as a sorted array of engine ids (first being most preferred). These ids are then mapped back in to a node list and returned.

Our use case is incorporating custom decisions from our metrics / health services in placement decisions.

Obviously there is performance overhead in marshalling decisions out to external processes and back but we're only running 40 or so node swarms and the overhead is worth it to us to be able to experiment with better placement strategies.

If there is no interest, please feel free to close the PR, otherwise ideas welcome :smiley:

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -29,7 +29,7 @@ var (
 				flRefreshIntervalMin, flRefreshIntervalMax, flFailureRetry, flRefreshRetry,
 				flHeartBeat,
 				flEnableCors,
-				flCluster, flDiscoveryOpt, flClusterOpt, flRefreshOnNodeFilter, flContainerNameRefreshFilter},
+				flCluster, flDiscoveryOpt, flSchedulerOpt, flClusterOpt, flRefreshOnNodeFilter, flContainerNameRefreshFilter},
 			Action: manage,
 		},
 		{
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -142,6 +142,11 @@ var (
 		Usage: "discovery options",
 		Value: &cli.StringSlice{},
 	}
+	flSchedulerOpt = cli.StringSliceFlag{
+		Name:  "scheduler-opt",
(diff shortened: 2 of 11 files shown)


davidcurrie:
We have a similar requirement for being able to plug in custom scheduling and would love to see an extension point along these lines.

xianlubird:
That's cool

mikesimons:
Looks like there is at least interest from the community here so I'll polish this up a little.
If anyone has any specific requests, now is the time to make them.

My plan is as follows:

1. Fill gaps in key scheduling data (like ports)
2. Add happy path tests
3. Identify edge cases and add failure handling + tests
4. Implement --scheduler-opt
5. Implement alternate connection methods (null, go rpc, socket)

I may also see what it will take to implement something similar in swarmkit since swarm looks like it might be taking a back seat (if not entirely deprecated) in the future.

mikesimons:
I have a bit of refining to do on the simplified schema side of things but otherwise this is fairly complete so I'm looking for feedback.

I realized that most folk are probably running swarm in a container so exec would be sort of pointless (especially since the official swarm image is from scratch). I focused instead on http as the transport. I've left it open to be fairly easy to implement additional transport if required.

This PR now adds the additional "external" scheduler as well as a new cli flag `--scheduler-opt` in the same vein as `--discovery-opt`.

You can start swarm with it as follows (where scheduler is running at `127.0.0.1:1234` and there is a docker node at `127.0.0.1:9123`):
```javascript
./swarm manage --strategy=external --scheduler-opt=url=http://127.0.0.1:1234 nodes://127.0.0.1:9123
```

PR #2342, opened 2016-06-09
Edit docs/discovery.md

NanXiao:
Since only one "<node_ip>" is invisible in Markdown file, use "
node IP address" instead.

Signed-off-by: NanXiao <nan@chinadtrace.org>

### DIFF
--- a/docs/discovery.md
+++ b/docs/discovery.md
@@ -186,7 +186,7 @@ swarm is connected to the public internet. To create your cluster:
 
 2. Create each node and join them to the cluster.
 
-    On each of your nodes, start the swarm agent. The <node_ip> doesn't have to be public (eg. 192.168.0.X) but the the Swarm manager must be able to access it.
+    On each of your nodes, start the swarm agent. The node IP address doesn't have to be public (eg. 192.168.0.X) but the the Swarm manager must be able to access it.
 
         $ swarm join --advertise=<node_ip:2375> token://<cluster_id>
 


jimmyxian:
@NanXiao Good Catch!
But I think we can use \\\<node_ip\\>  instead

NanXiao:
@jimmyxian Yes, we can use slash. I just read "`The node IP address ...`" at the beginning of the `discovery.md`, so I think maybe we should use the same words. Anyway, if you think ` \<node_ip\>` better, it is also OK. :-)

NanXiao:
@jimmyxian 

For https://github.com/docker/swarm/pull/2342/commits/04b8fa8f1d7ccd82fa28cd3dfabd1c8c3a39c8fc, if you prefer `\<node_ip\>`, I can modify it.

For https://github.com/docker/swarm/pull/2342/commits/bccf8c5f6454e9d1c84495f65380410483117a50, could you help to review it? Thanks very much in advance!

xiaods:
how about update?

jimmyxian:
@NanXiao Thanks for the contribution.
Can you make a another PR about `Check whether channel is closed in monitorDiscovery`.
That will be good.:)

NanXiao:
@jimmyxian 
Thanks for your reply!

According to your suggestion, a new PR #2370 is created. Please help to review it, thanks very much!

jimmyxian:
LGTM

PR #2343, opened 2016-06-09
Optimizing matching of image affinity

hdanniel:
The affinity.Match(images...) iterates all over a images slice with the ID and tags of all images for each node in the swarm cluster. With a large number of images  the execution time of this function increases affecting the performance when creating a new container. In our case, with thousands of images for each node, this function takes 4-5 seconds to return.

Our proposal changes the images slice with a map with the image ID or tag as a key and 1 as a value. If images[affinity.value] exists, the node is added to the candidates slice. After this change, currently run in production the time to select a node to create the container with affinity reduces from 10 seconds (2 nodes, 10000 images) to milliseconds.

### DIFF
--- a/scheduler/filter/affinity.go
+++ b/scheduler/filter/affinity.go
@@ -45,18 +45,18 @@ func (f *AffinityFilter) Filter(config *cluster.ContainerConfig, nodes []*node.N
 					candidates = append(candidates, node)
 				}
 			case "image":
-				images := []string{}
+				images := make(map[string]int)
 				for _, image := range node.Images {
-					images = append(images, image.ID)
-					images = append(images, image.RepoTags...)
+					images_map[image.ID] = 1
 					for _, tag := range image.RepoTags {
 						repo, _ := cluster.ParseRepositoryTag(tag)
-						images = append(images, repo)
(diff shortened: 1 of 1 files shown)


xiaods:
@hdanniel cloud you please add a integration testing to track the performance issue.

jinuxstyle:
The proposed changes broken the integration test because it changed the expected behavior of affinity. What the affinity.Match(images...) does is not simply matching image names, instead it's way more powerful because it supports regular expression as well as operators (e.g. !=). So I think it's not appropriate to simply replace it with map. We need a more thorough solution for improving the performance.

jinuxstyle:
The proposed changes broken the integration test because it changed the expected behavior of image affinity. What the affinity.Match(images...) does is not simply matching image names, instead it's way more powerful because it supports regular expression as well as operators (e.g. !=). So I think it's not appropriate to simply replace it with map. We need a more thorough solution for improving the performance.

xiaods:
yeah, @jinuxstyle said right.

hdanniel:
It's correct. The proposed changes only worked for us because we don't need to take into account regular expressions.

PR #2344, opened 2016-06-12
Add curl to the base swarm image

Spritekin:
Allows running swarm commands using curl from within the container. Makes for a very useful way to test the service status. Adds 2MB to the final image.

Very useful when running Health Checks from Marathon using mesos.

Signed-off-by: Guillermo Rodriguez grodriguez@cmcrc.com

PR #2067, opened 2016-04-01
Allow tasks to continue running in Mesos even after swarm manager crashes

Spritekin:
By default, when a framework fails, Mesos will terminate any tasks started by a framework This parameter allows the tasks to continue running for this number of seconds after the swarm manager dies. (default 0)

If Swarm manages to reconnect with the same mesos task ID it had before then tasks will be reassigned to it.

Signed-off-by: Guillermo Rodriguez <grodriguez@cmcrc.com>

PR #2345, opened 2016-06-12
Mesos failover timeout

Spritekin:
By default, when a framework fails, Mesos will terminate any tasks started by that framework.

Signed-off-by: Guillermo Rodriguez <grodriguez@cmcrc.com>


Spritekin:
Note, this is the separation of the changes proposed in the PR #2067

This PR only updates the mesos timeout and frameworkid for task resilience.

The changes to modify the Dockerfile only are now in PR https://github.com/docker/swarm/pull/2344

Spritekin:
I messed up the commits. Closing.

PR #2346, opened 2016-06-12
Tasks to continue running after swarm crash

Spritekin:
By default, when a framework fails, Mesos will terminate any tasks started by a framework

Signed-off-by: Guillermo Rodriguez <grodriguez@cmcrc.com>

### DIFF
--- a/cli/help.go
+++ b/cli/help.go
@@ -47,7 +47,7 @@ Options:
                                     {{printf "\t * swarm.createretry=0\tcontainer create retry count after initial failure"}}
                                     {{printf "\t * mesos.address=\taddress to bind on [$SWARM_MESOS_ADDRESS]"}}
                                     {{printf "\t * mesos.frameworkid=\tspecify a framework id, if not set mesos assigns a random id [$SWARM_MESOS_FRAMEWORKID]"}}
-                                    {{printf "\t * mesos.failovertimeout=0\tkeep tasks launched by this framework running for this number of seconds in case this Swarm manager crashes (mesos will reattach the tasks to a new Swarm manager started with the same framework id) [$SWARM_MESOS_FAILOVER_TIMEOUT]"}}
+                                    {{printf "\t * mesos.failovertimeout=0\tkeep tasks launched by this framework running for this number of seconds in case this Swarm manager crashes [$SWARM_MESOS_FAILOVER_TIMEOUT]"}}
                                     {{printf "\t * mesos.checkpointfailover=false\tcheckpointing allows a restarted slave to reconnect with old executors and recover status updates, at the cost of disk I/O [$SWARM_MESOS_CHECKPOINT_FAILOVER]"}}
                                     {{printf "\t * mesos.port=\tport to bind on [$SWARM_MESOS_PORT]"}}
                                     {{printf "\t * mesos.offertimeout=30s\ttimeout for offers [$SWARM_MESOS_OFFER_TIMEOUT]"}}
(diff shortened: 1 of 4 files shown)


Spritekin:
Note, this is the separation of the changes proposed in the PR #2067

This PR only updates the mesos timeout and frameworkid for task resilience.

The changes to modify the Dockerfile only are now in PR #2344

ezrasilvera:
@Spritekin Looks good. I also think the curl stuff should have a separate PR.  I suggest you close #2067

ezrasilvera:
@Spritekin not sure the help message for the timeout is clear enough, how about: 
keep the tasks lunched by this framework, running for this number of seconds in case this Swarm manager crashes ....

what did you mean by "recover by restarting with the same framwrodk id " ?  I'm not sure we should refer to the HA/recovery flow in the help of this parameter.

Spritekin:
@ezrasilvera 
Closed PR 2067.

Will try to rephrase. The problem is both parameters are very related. The main concept is:

Mesos identifies a framework by its ID. All tasks started by the framework are assigned to that ID. If the framework crashes the tasks keep assigned to that ID for the failover_timeout seconds (default 0 means they are stopped immediately). If you restart a framework with the same framework id before the framework_timeout elapses, then the tasks are reassigned.

Now, about HA, it doesn't work with mesos because of contradicting behaviours. 
a. when you start swarm mesos will assign it a unique ID. So if you start three swarm in HA mode, each will have its own ID. Then, when you start tasks, they will start with the leader ID. But if the leader crashes then mesos can't reassign them to the new elected leader because they have different IDs.

b. So then I tried to start three swarm with the same ID, the first one took leadership, then the other two were waiting. But when mesos sent offers, instead of being sent to the leader they are sent to the last framework connected with that ID (which happens not to be the leader) so the leader never receives offers so it is unable to launch tasks. The solution in this case would be to allow a secondary swarm to reroute offers to the leader.

In any case, my HA is to quickly detect problems in swarm using a Marathon health checks and restart quickly. That's where the curl requirements get in but then that's another story.

Spritekin:
@ezrasilvera 
Ok, I have adjusted the help text. Please tell me if its a bit clearer.

ezrasilvera:
@Spritekin I agree that the to get HA like capabilities (i.e., detect failure and restart a new Swarm master) you need both PRs . This doesn't mean, however, that one can't use each of this separately for whatever reason he wants to :-)  
For example, I can see why people would like to explicitly set the farmeworkID even without th erestart scenario.

ezrasilvera:
that's why I thought that in the help we shouldn't refer to a specific use-case of the parameter but simply to what the parameter does

Spritekin:
@ezrasilvera 
Removed specific use case text from the help strings.

Spritekin:
@ezrasilvera 
Why are the checks failing? Doesn't seem caused by this change.

ezrasilvera:
@Spritekin I think it's not related to your changes. You can see that other builds failed as well. I just push again from time to time to initiate the build (you can use --no-edit and --amend so no real commit is needed)

Spritekin:
@ezrasilvera @jimenez 
What now! I don't see a problem in the documentation and it says its failing!

I like this project but I don't have time for being stopped by nonexistent errors. Please ask someone to fix the build system.

Luck!

Spritekin:
And now I can't try to commit again because it says:
fatal: You are in the middle of a cherry-pick -- cannot amend.

Aaaaaarggggg....

Spritekin:
Closing. Pulled an invalid ticket

PR #2352, opened 2016-06-15
Tasks to continue running after swarm crash

Spritekin:
By default, when a framework fails, Mesos will terminate any tasks started by a framework

Signed-off-by: Guillermo Rodriguez grodriguez@cmcrc.com


Spritekin:
@ezrasilvera @jimenez 
Not sure why this throws errors. I think they are not related to the change. Can you do something please?

nishanttotla:
@Spritekin the errors are unrelated.

nishanttotla:

PR #2354, opened 2016-06-16
Use volume.Scope to determine create scheduling

cpuguy83:
Volumes in engine now have a `Scope` field similar to networks.
This makes `VolumeCreate` take advantage of it so that only 1 node needs
the create action if the `volume.Scope == "global"`.

Please lmk on implementation, not incredibly familiar with this codebase.
This seems like it's working. I notice that networks have some in-memory state that volumes don't seem to have.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -478,6 +478,7 @@ func (c *Cluster) CreateNetwork(name string, request *types.NetworkCreate) (resp
 		return nil, err
 	}
 	if nodes != nil {
+		c.RLock()
 		resp, err := c.engines[nodes[0].ID].CreateNetwork(name, request)
 		if err == nil {
 			if network := c.engines[nodes[0].ID].Networks().Get(resp.ID); network != nil && network.Scope == "global" {
@@ -488,6 +489,7 @@ func (c *Cluster) CreateNetwork(name string, request *types.NetworkCreate) (resp
 				}
 			}
 		}
+		c.RUnlock()
 		return resp, err
 	}
 	return nil, nil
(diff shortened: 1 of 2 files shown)


vieux:
I think we should also add the volume to the internal state when it's global, like we do here: https://github.com/cpuguy83/swarm/blob/88b250a9ce2efd0adc54167318864d778fd37aef/cluster/swarm/cluster.go#L486

cluster/swarm/cluster.go
```diff
@@ -511,21 +511,38 @@ func (c *Cluster) CreateVolume(request *types.VolumeCreateRequest) (*types.Volum
...
-				v, er := engine.CreateVolume(request)
+		// When the scope is `local` we need to create the volume everywhere
+		var i int
+		for _, e := range c.engines {
+			if i == 0 {
+				v, er := e.CreateVolume(request)
 				if v != nil {
 					volume = v
 					err = nil
+					if v.Scope == "global" {
```
dongluochen: If CreateVolume fails for whatever reason, `v` is `nil` and this would not be checked. It'd end up creating Volume on other engines. 

Can scope be derived from request directly? So that it's always clear if it's `global` or `local`.

cpuguy83: No I don't think it could be derived from the request.
Maybe we should just error out if we get an error here?

dongluochen: I think error out is a better approach.

cluster/swarm/cluster.go
```diff
@@ -496,59 +498,60 @@ func (c *Cluster) CreateNetwork(name string, request *types.NetworkCreate) (resp
...
-					volume = v
-					err = nil
-				}
-				if er != nil && volume == nil {
-					err = er
-				}
-			}(e)
-		}
-		c.RUnlock()
+	nodes, err := c.scheduler.SelectNodesForContainer(c.listNodes(), config)
```
cpuguy83: I'm not sure why, but this is failing when a node is specified, like `docker volume create mynode/foo`

Confirmed it works locally when using the node ID.
Some help would be appreciated.

dongluochen: I get this output running locally. 

```
  docker_swarm volume create --name=node-0/test_volume2
  run docker_swarm volume ls
  echo "${output}"
  [ "${#lines[@]}" -eq 5 ]
```

```
   DRIVER              VOLUME NAME
   local               node-0/test_volume
   local               node-0/test_volume2
   local               node-1/71884a16b1716c76860415275876ad4c8c185996857c412200131ea35575f322
   local               node-1/test_volume
   local               node-1/test_volume2
```

cpuguy83: For me if I provide a name like you have done, this call is failing saying that it can't find a node.

dongluochen: I didn't run the commands manually on my machine. I added the line `echo "${output}"` before line 69 to find what's the output form `docker volume ls` and ran the test. The create command was successful.

srust:
Any chance for some love on this? As of today, volumes in swarm are still less than ideal. Duplicate names, multiple nodes being listed for one global volume, etc. And the volume Capabilities from docker engine still not being represented in swarm. 

What's the best path forward here in the swarm context?

PR #2370, opened 2016-06-22
Check whether channel is closed in monitorDiscovery.

NanXiao:
Select on a closed channel will always return immediately, and this
will cause an endless loop in this goroutine.

Signed-off-by: NanXiao nan@chinadtrace.org

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -359,9 +359,13 @@ func (c *Cluster) removeEngine(addr string) bool {
 func (c *Cluster) monitorDiscovery(ch <-chan discovery.Entries, errCh <-chan error) {
 	// Watch changes on the discovery channel.
 	currentEntries := discovery.Entries{}
-	for {
+	for ch != nil || errCh != nil {
 		select {
-		case entries := <-ch:
+		case entries, ok := <-ch:
+			if !ok {
+				ch = nil
+				break
+			}
 			added, removed := currentEntries.Diff(entries)
 			currentEntries = entries
 
@@ -375,7 +379,11 @@ func (c *Cluster) monitorDiscovery(ch <-chan discovery.Entries, errCh <-chan err
 			for _, entry := range added {
 				c.addEngine(entry.String())
 			}
-		case err := <-errCh:
+		case err, ok := <-errCh:
+			if !ok {
+				errCh = nil
+				break
+			}
 			log.Errorf("Discovery error: %v", err)
 		}
 	}


cluster/swarm/cluster.go
```diff
@@ -359,9 +359,13 @@ func (c *Cluster) removeEngine(addr string) bool {
 func (c *Cluster) monitorDiscovery(ch <-chan discovery.Entries, errCh <-chan error) {
 	// Watch changes on the discovery channel.
 	currentEntries := discovery.Entries{}
-	for {
+	for ch != nil || errCh != nil {
 		select {
-		case entries := <-ch:
+		case entries, ok := <-ch:
+			if !ok {
+				ch = nil
```
dongluochen: I like the change of breaking the infinite loop. But I don't think we should set channel to `nil`. It has side effects.

NanXiao: @dongluochen What is the side effects? Thx!

cpuguy83: A nil channel will always block
Calling close on a nil channel will cause a panic.

NanXiao: @cpuguy83 Thanks for your comments! But I don't think these side effects will occur in `monitorDiscovery` function.

dongluochen: @NanXiao Reading from `nil` will block and it'd be hard to debug. We shouldn't set it to `nil`.

NanXiao: @dongluochen According to [Curious Channels](http://dave.cheney.net/2013/04/30/curious-channels): When a nil channel is part of a select statement, it is effectively ignored, so niling a removes it from selection, leaving only b which blocks until it is closed, exiting the loop without spinning.

dongluochen: The problem is this function doesn't own the 2 channels. They could be used (or will be used) in other places. I don't see why setting them to `nil` has any benefit while posing a threat.

NanXiao: @dongluochen 
(1) > The problem is this function doesn't own the 2 channels. They could be used (or will be used) in other places.
Yes, `monitorDiscovery` doesn't own the 2 channels. But I don't think it does matter:  

```
case entries, ok := <-ch:
        if !ok {
		ch = nil
                break
        }
```
If `ok` is `false`, this means the channel has been closed in other places. So it is not meaningful to still wait here to receive messages from this channel. The same logic should apply to `errCh` as well.

(2) > I don't see why setting them to `nil` has any benefit while posing a threat.
If don't set them to `nil`, is there any better method to prevent dead-loop of this `for` statement?

Thanks very much!

dongluochen: > If don't set them to nil, is there any better method to prevent dead-loop of this for statement?

You can just `return` when any channel is closed.

NanXiao: @dongluochen 
> You can just return when any channel is closed.

Maybe it is appropriate for this case, but personally I don't think it is reasonable. If `select` is polling on `5` channels, return the function when `1` channel is closed while others are stilll available doesn't make sense.

Thanks very much!

dongluochen: In that case you can use a local variable to decide if the for loop should continue. The reason here is I don't think `monitorDiscovery` should change the channels. It was not designed to do that.

NanXiao: @dongluochen 
> The reason here is I don't think monitorDiscovery should change the channels. It was not designed to do that.

Very sorry for my stubbornness. I think `monitorDiscovery` doesn't **change** the channels, and it just set local channel variable to  `nil` when it knows the channel is closed.

dongluochen: From the point of this function, the channels are not local.

NanXiao: @dongluochen  
Sorry, I still can't get your point. The framework of `monitorDiscovery` is like this:  

	func (c *Cluster) monitorDiscovery(ch <-chan discovery.Entries, errCh <-chan error) {
		// Watch changes on the discovery channel.
		currentEntries := discovery.Entries{}
		for {
			select {
			case entries := <-ch:
				......				
			case err := <-errCh:
				......
			}
		}
	}
(1) The channels `ch` and `errCh` are parameters of `monitorDiscovery`, and if they are closed at other places, I don't think it is meaningful of still receiving values from these channels;  
(2) For the already closed channels `ch` and `errCh`, they are only visible in `monitorDiscovery`. If setting them values as `nil`, what are the side-effects?

Thanks!

PR #2385, opened 2016-06-28
fix Switch the primary docs

clhlc:
HI, i fix the issue [https://github.com/docker/swarm/issues/2356](url)

### DIFF
--- a/docs/multi-manager-setup.md
+++ b/docs/multi-manager-setup.md
@@ -154,8 +154,8 @@ To switch the `DOCKER_HOST` to use `manager-2` as the primary, you do the follow
     Containers: 0
     Images: 25
     Storage Driver:
-    Role: Replica  <--------- manager-2 is a Replica
-    Primary: 192.168.42.200
+    Role: Primary  <--------- manager-2 is the Primary manager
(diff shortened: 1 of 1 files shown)


nishanttotla:
Fixes #2356

nishanttotla:
Actually the doc is a little confusing, it's not clear if the last section is meant for before the failover happened. I think we should move the whole "Switch the Primary" part to before the failover case.

cc @allencloud

dongluochen:
@nishanttotla @clhlc I think the last section `Switch the primary` should be `Switch DOCKER_HOST pointing to new primary`. We may update text to make it clear.

clhlc:
@nishanttotla  in swarm 1.2.3，primary node info is
`...
Server Version: swarm/1.2.3
Role: primary
Strategy: spread
Filters: health, port, containerslots, dependency, affinity, constraint
...`

abronan:
Thanks @clhlc,

LGTM

(there are definitely some changes to be made but this fixes the immediate issue which was that the message was incorrectly mentioning `replica` instead of `primary`, let's not block this one and make further changes later on if the process is still unclear)

PR #2424, opened 2016-07-20
Fix rescheduling test

dongluochen:
The following line to wait for node-0 recovery doesn't work. Fix it with direct node-0 status validation.

```
retry 30 1 eval "test \"$(docker_swarm info | grep \"Status: Unhealthy\" | wc -l)\" = '0'"
```

This problem surfaces only in Docker-1.12, where `docker start` may be taking longer than before. 

cc @nishanttotla @abronan @allencloud. 


abronan:
@dongluochen Could you please force push to trigger the CI? Thanks.

Otherwise LGTM.

nishanttotla:
LGTM, after triggering CI

PR #2425, opened 2016-07-20
consider all nodes returning from the scheduler

ezrasilvera:
SelectNodesForContainer returns a list of node candidates (priority sorted) , but only the first one is used. In case that engine is faulty we can try the other nodes.


xianlubird:
If the first engine is unhealthy, maybe some log or message should be printed so that we(developer) can know what has happened when scheduler result does not match our expect.
@ezrasilvera

nishanttotla:
Closing due to lack of activity. Please rebase and reopen if you wish to continue working on it.

PR #2437, opened 2016-07-31
Dokku does not have swarm support

josegonzalez:
Fix the docker swarm documentation, which is better than misleading users while we investigate support.

Refs https://github.com/dokku/dokku/issues/1795


josegonzalez:
Closing as I am not going to sign off on this change through the github ui.

PR #2443, opened 2016-08-05
Rename github.com/codegangsta to github.com/urfave

yehohanan7:
Noticed the below comment from the cli module used by swarm, hence updated the path to keep the dependency consistent

"Notice: This is the library formerly known as github.com/codegangsta/cli -- Github will automatically redirect requests to this repository, but we recommend updating your references for clarity."


vieux:
I'm ok to do a rename, but I see +4,185 / −1,404 so it's more than a rename.
I guess you also updated to a newer version ?

yehohanan7:
Hi, Yes, I used the latest revision. If that's unnecessary, I will revert to the older revision.

yehohanan7:
updated to use the same revision of the package now.

PR #2472, opened 2016-09-01
Update engine API version

wsong:
Note that compose/up.bats and api/network.bats are now failing; the former
because docker-compose doesn't support creating Attachable networks yet, and
the latter because the Attachable key doesn't show up in docker network
responses.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -101,63 +101,63 @@
 		},
 		{
 			"ImportPath": "github.com/docker/engine-api/client",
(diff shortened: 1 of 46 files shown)


nishanttotla:
@wsong I'm afraid `engine-api` has now been deprecated, so we can't merge this PR anymore 😞 

To update to latest, we'll have to not only update the vendoring of `docker/docker`, but also change paths all across `swarm`.

If you want to take up this task, please let us know!

Related to #2488.

PR #2525, opened 2016-10-19
Add instruction to upload binary in release

dongluochen:
Some users prefer not to run docker on manager nodes to reduce attack surface. Providing binary would help their deployment. 

Signed-off-by: Dong Chen dongluo.chen@docker.com

### DIFF
--- a/RELEASE-CHECKLIST.md
+++ b/RELEASE-CHECKLIST.md
@@ -13,6 +13,8 @@ git push $GITHUBUSER bump-<version>
 
 Open PR on docker/swarm. Merge the PR before moving to next step. If the release is an RC, then `CHANGELOG.md` should list the latest RC.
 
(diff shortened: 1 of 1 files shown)


vincentwoo:
sorry to make life harder for you, but i noticed that in your release for 1.2.5 you made a tar named

`swarm-1.2.5-linux-x86_64.tgz`

that itself contains

`swarm-1.2.5`

If you look at what releases in docker/docker do, they have a versioned tar:

`docker-1.12.2.tgz`

that contain unversioned binaries like

`dockerd`

I think this is preferable, otherwise any automation scripts will also probably have to rename the swarm binary.

dongluochen:
@vincentwoo Agree. The instruction here is correct. The one I uploaded had binary with version in name wasn't following this instruction. I have replaced it with a new one.

nishanttotla:
LGTM

PR #2528, opened 2016-10-21
Update API handler if election follower channel gets an update

wsong:


wsong:

We noticed that restarting Swarm would frequently result in errors that looked like "No elected primary cluster manager" for several seconds in the logs after Swarm restarted. During this time, requests to the Swarm API would fail.

The source of this error during upgrades appears to be an issue with how Swarm handles leadership election. Swarm uses Docker's leadership library, which acquires a lock here:

https://github.com/docker/leadership/blob/master/candidate.go#L126

Swarm uses this in the primary to set a key (namely, `/v2/keys/docker/swarm/leader`) in, for instance, etcd, whose value is the primary's IP address. This is how we mark leaders in etcd.

Our libkv etcd library holds onto this lock by resetting it every few seconds:

https://github.com/docker/libkv/blob/master/store/etcd/etcd.go#L543

As far as I can tell, the intent here is that as long as the Swarm primary is up, it will continually rewrite this key with a certain TTL (the value we use is 20 seconds). If the primary goes down, the key will expire after 20 seconds and another node will be able to rewrite that key with its own IP address as its value, thus becoming the new Swarm primary.

The problem is that when we restart the Swarm container, the leader key might still be sticking around from the last time the old container wrote it out. The newly-restarted container has to wait up to 20 seconds before it can successfully write out to the leader key; during that time, Swarm reports that it doesn't know who the primary is.

The fix I've implemented here is to essentially check the leader key's value and set up the proper API handler if it's already the correct value.

wsong:
The integration tests appear to pass on my local machine, for what it's worth (Docker engine version 1.12.1).

dongluochen:
@wsong Nice catch of the problem. Can the leader give up Primary at shutdown? When it restarts, it'd be able to grab leader right away. If there are other replica, they would be able to claim Primary quickly.

wsong:
We ran into this problem when we were essentially doing `docker rm -f` on the manager container, which I believe sends a SIGINT signal which would not give the leader any time to do any cleanup.

dongluochen:
@wsong If I understand correctly, the current PR tries to reduce delay for the primary to update/restart. In your case the manager which was primary claims primary after upgrade. It's not what multiple managers are designed to do. Multiple managers should allow other replica to take over primary. 

Running `docker rm -f` is not graceful. I think it better to allow leader failover. How about using this to upgrade swarm?

```
docker stop #managercontainer#
docker run --rm #newmanager#
``` 

With `--rm` the container is gone when it exits.

wsong:
We were actually seeing this issue with just one manager; if it restarted for any reason, it would take a while for its API to become active again because it would keep saying "No elected primary cluster manager" for several seconds after startup:

https://github.com/docker/swarm/blob/master/api/replica.go#L45

It would also probably be worthwhile for the Swarm leader to gracefully give up the lock when it shuts down, but we'd also like to support the case where the Swarm leader gets killed for some reason and comes back up (regardless of whether or not we're doing it).

dongluochen:
@wsong I agree with you that we have 2 cases to address. 

1. Multiple managers failover. The primary should give up leadership at upgrade so followers can take leadership quick. 
2. Primary recovers from crash. When the manager recovers from crash, it understands it was previously primary. It should claim leader right away. 

I think the leader election test failure is related to your change. Please take a look. 

```
not ok 27 leader election - rescheduling
# (from function `retry' in file ./helpers.bash, line 80,
#  in test file ./replication.bats, line 164)
#   `retry 30 1 containerRunning "c1" "node-1"' failed
```

wsong:
The replication tests appear to pass on my local machine; maybe the Jenkins environment is a bit flaky or some timeouts need to be tweaked?

wsong:
I made a few tweaks and replication integration tests appear to be passing now. The changes I made were:

1. Make sure that we report "Role: primary" even if the `candidate` object hasn't been elected yet by checking the value of `follower.Leader()`.
2. Don't update the event handler watchdog until the candidate really gets elected.

wsong:
One thing that perhaps would clear things up: I don't understand why today, when the election follower sees that the elected leader has its own address, we set the Primary to empty string:

wsong:
Whoops, hit send too soon. Anyways, we set the Primary to empty string:

https://github.com/docker/swarm/blob/master/cli/manage.go#L198

And that activates this error message:

https://github.com/docker/swarm/blob/master/api/replica.go#L45

But clearly a primary has, in fact, been elected. Is that error message simply wrong?

xianlubird:
> One thing that perhaps would clear things up: I don't understand why today, when the election follower sees that the elected leader has its own address, we set the Primary to empty string:

Maybe leader has register another router as primary router ?   I also want to know why.

xianlubird:
@wsong 
I notice that swarm will set primary handler when election success
https://github.com/docker/swarm/blob/master/cli/manage.go#L175

nishanttotla:
@wsong can you rebase this PR?

wsong:
Rebased.

wsong:
I've updated this PR with a new commit that also fixes the issue of losing leadership when your TCP connection to the KV store dies. Previously, this PR would ensure that when the follower channel gets an update, if the address of the new leader was the same as our Swarm classic manager's address, we would consider ourselves the leader. However, the follower channel only receives an update when it first connects to the KV store and when an actual change occurs. Meanwhile, if the leadership election candidate goroutine loses its TCP connection to the KV store, it'll say that it's lost the leadership election. This causes the Swarm classic manager to no longer consider itself the leader until it can either write back out to the KV store (which can take up to 20 seconds as it's waiting for its previously-written key's TTL to expire) or until the follower channel sends a new update (which it won't do, because the value of the Swarm leader key hasn't changed).

My new commit just kills the follower and restarts it if we were previously the leader and then we lose the election; this ensures that if we lost the election due to a transient network failure, we reconnect to the KV store immediately, the follower sends out an update with the value of the Swarm leader key, and since the value is our address, we immediately reestablish ourselves as the leader.

cli/manage.go
```diff
@@ -132,81 +140,71 @@ func getDiscoveryOpt(c *cli.Context) map[string]string {
...
 			if isElected {
 				log.Info("Leader Election: Cluster leadership acquired")
+				wasLeader = true
 				watchdog = cluster.NewWatchdog(cl)
 				server.SetHandler(primary)
 			} else {
 				log.Info("Leader Election: Cluster leadership lost")
 				cl.UnregisterEventHandler(watchdog)
 				server.SetHandler(replica)
+				if wasLeader {
```
dongluochen: `run()` was designed to be long running. Why exit if there is no error?

cli/manage.go
```diff
@@ -132,81 +140,71 @@ func getDiscoveryOpt(c *cli.Context) map[string]string {
...
-	leaderCh, errCh := follower.FollowElection()
-	for {
-		select {
 		case leader := <-leaderCh:
 			if leader == "" {
 				continue
 			}
 			if leader == addr {
 				replica.SetPrimary("")
+				server.SetHandler(primary)
```
dongluochen: When a node becomes leader, which event is triggered? 

```
case isElected := <-electedCh:
```
or 
```
case leader := <-leaderCh:
```

cli/manage.go
wsong: Causing `run()` to exit is just an easy way to make sure we reinstantiate the follower object. Note that `run()` already exits in the event of an error.

cli/manage.go
wsong: `electedCh` gets triggered when the manager successfully writes out to the KV store or when it loses the connection to the KV store (it returns true in the former case and false in the latter). `leaderCh` gets triggered any time the KV store's value changes. Thus, when a node becomes leader, both channels get triggered.

cli/manage.go
dongluochen: Each `run` would start `candidate.RunForElection()`. I don't see a cleanup here.

cli/manage.go
dongluochen: As you describe, both get triggered and they do similar but not identical things. It doesn't look clean. I understand it might be like this before this change. But as you are refactoring it, we'd better resolve it.

wsong: Essentially, the Docker leadership election library works as a two-part system: you have one channel for figuring out who the leader is, and another for trying to become the leader. We need both parts here, but the way it's currently wired up is a bit weird: we only consider ourselves the leader if we can successfully write out to the KV store, which means that if node A is the leader and no other nodes are in the election but node A can't write to the KV store for some reason (e.g. network flakiness), node A will report that it doesn't know who the leader is (the "no elected primary cluster manager" error), even though the KV store contains the correct value for the leader.

Maybe we could switch to a model where we use the follower channel to determine whether or not we're the leader, and we don't listen to the candidate's channel at all; that is, we continually try to elect ourselves as the leader, but we ignore whether or not we've successfully written to the KV store and we only look at the current value stored in the KV store, and if it's equal to our address then we consider ourselves the leader.

One failure case here is if you have a node A and a node B who are both in a leadership election. If node A is a leader and it loses its connection to the KV store, it will continue to consider itself the primary during this time, meaning that it will still schedule containers and so on. While node A is disconnected, node B might elect itself the leader, since it can still write out to the KV store. At that point, you will effectively have two primaries, since both node A and B will be responding to API requests.

dongluochen: > At that point, you will effectively have two primaries, since both node A and B will be responding to API requests.

Generally speaking in a consensus system, when the quorum fails, the system becomes read-only, i.e., no active master, which is better than 2 masters. Maybe the primary should give up when it cannot read from KV store.

wsong: Yeah, that approach might be a little extreme. Essentially, what I'm trying to solve with this PR is the phenomenon where a transient network blip causes the leader to think that quorum has been lost because it won't recognize itself as the leader until its previously-written key has expired. Thus, we need to make sure that after we lose a connection to the KV store (i.e. after `electedCh` sends false), we re-read the Swarm leader key's value as soon as we can reestablish a connection to see if we're still the leader.

Maybe there's some code organization cleanup we could do here, but I think the essential model of the two channel system here is what we'll have to stick with under these conditions.

wsong: Any further thoughts here? I think that the solution I've written here (i.e. tear down everything and reestablish a connection as fast as possible when we have a connection error) is the simplest solution using the libraries we have.

cli/manage.go
```diff
@@ -132,81 +140,71 @@ func getDiscoveryOpt(c *cli.Context) map[string]string {
...
 	}()
 
 	server.SetHandler(primary)
 }
 
-func run(cl cluster.Cluster, candidate *leadership.Candidate, server *api.Server, primary *mux.Router, replica *api.Replica) {
-	electedCh, errCh := candidate.RunForElection()
+func run(cl cluster.Cluster, client store.Store, keyPath string, handler *statusHandler, server *api.Server, primary *mux.Router, replica *api.Replica, addr string) error {
+	candidate := handler.candidate
+	electedCh, candidateErrCh := candidate.RunForElection()
```
dongluochen: I think goroutine `Candidate.campaign` is leak when `run` returns.

cli/manage.go
```diff
@@ -132,81 +140,71 @@ func getDiscoveryOpt(c *cli.Context) map[string]string {
...
 	for {
 		select {
 		case isElected := <-electedCh:
 			if isElected {
 				log.Info("Leader Election: Cluster leadership acquired")
+				wasLeader = true
 				watchdog = cluster.NewWatchdog(cl)
 				server.SetHandler(primary)
 			} else {
 				log.Info("Leader Election: Cluster leadership lost")
```
dongluochen: Since you are trying if this node `wasLeader` before, it should be more accurate with 

```
if wasLeader {
    log.Info("Leader Election: Cluster leadership lost")
}
```

cli/manage.go
dongluochen: The election module had many failures before. I think we need to be careful of changes. We should be explicit of what this achieves, not a convenient route.

cli/manage.go
wsong: Good call. Added a deferred `candidate.Stop()` call.

cli/manage.go
wsong: Done.

cli/manage.go
wsong: Definitely. Still, I really want to find a good way to avoid these 20 second long delays when we lose connection to the KV store in a single-node environment.

dongluochen:
Test failure is related. PTAL.

cli/manage.go
wsong: The test failures were caused by this call; we actually only want to call `candidate.Stop()` if we haven't already received an error. If we get an error, the candidate's channels will already be closed.

cli/manage.go
nishanttotla: I think that if existing code has two channels, then we should stick with it in this PR, even if it warrants some clean up in the future.

wsong:
I've added a new commit that completely reinstantiates the candidate object was necessary; we were seeing some issues in testing where the candidate's channels were panicking as the leadership code tried to send things along closed channels.

I also implemented exponential backoff when trying to reattach to the KV store; we saw that in our testing, if you lose connection to the KV store, it'd take at least 10 seconds to reconnect, which was just way too long.

wsong:
I've also added a new commit to block on calls until we get an elected primary cluster manager; this makes it so that if a user happens to hit Swarm classic when a leadership election is taking place, they won't immediately get an error back.

In our testing, this set of changes has pretty much eliminated our "No elected primary cluster manager" errors.

nishanttotla:
@wsong is this still important to get in?

wsong:
This is no longer relevant now that we block in `api/replica.go` until we have an elected primary cluster manager.

PR #2534, opened 2016-10-28
Refresh network cache after a network delete

wsong:

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -267,6 +267,11 @@ func (c *Cluster) RemoveNetwork(network *cluster.Network) error {
 			}
 		}
 	}
+	if err == nil && network.Scope == "swarm" {
+		for _, engine := range c.engines {
+			engine.DeleteNetwork(network)
+		}
+	}
 	if err != nil && network.Scope == "global" {
 		log.Debugf("Failed to remove global scope network %s from any engine...", network.ID)
 	}


wsong:

Integration test failures appear largely unrelated except for a failure in engine-master, which looks like it could be an issue with new engine fields in `network inspect` responses?

wsong:
Hmm, I see that you guys are already updating the cache for all engines during network deletes, but only for `global` networks. Shouldn't we also be doing this for `swarm` networks?

nishanttotla:
@wsong Do you mean `swarm` networks in the context of Swarm mode?

wsong:
That's correct; that is, networks with a scope field set to "swarm".

wsong:
To give a bit more information, here's the specific error we're seeing:

1. Have two manager nodes, node A and node B, in Swarm mode, where node A is the leader.
2. Run `docker network create --driver overlay <node A name>/net1` on node A. Note that we're using the node-specific network creation logic here.
3. Wait for both nodes to pick up this new network (initially, running `docker network ls` against either node shows a blank driver field for this network; after several seconds, this field gets filled in).
4. Run `docker network rm net1; docker network create -d overlay <node B name>/net1 && docker network inspect net1` on node B. Again, we're using the node-specific logic here.
5. It's expected that you'll get back the details of the newly-created net1 network; instead, you see this error:

```
$ docker network rm net1; docker network create -d overlay default-cs1/net1 && docker network inspect net1
net1
e92z5t44t9y7k1yuxl8stskbi
[]
Error: No such network: net1
```

The reason why is that when we try to inspect the new network in node B in this loop here:

https://github.com/docker/swarm/blob/master/cluster/network.go#L133

We have two possible candidates with the network name "net1": the old one from node A and the new one from node B. This results in a 404 down on line 41.

The solution, as best I can see, is to make sure that we delete networks from all nodes during network deletion if the scope is `swarm`.

nishanttotla:
I understand the requirement of this change, but concerned about adding this here because ideally it would be good to keep this requirement away from Swarm itself.

WDYT @dongluochen?

wsong:
I'm not quite sure I follow you; why wouldn't clearing out Swarm's network cache be Swarm's responsibility?

dongluochen:
I think several things could be fixed here. 

On Node A, we shouldn't allow a **global** network to be **node specific**. 

```
ubuntu@ip-172-19-29-201:~$ docker network ls | grep -i net1
7e2a85265458        ip-172-19-29-201/net1   overlay             global
```

> Run docker network rm net1; docker network create -d overlay <node B name>/net1 && docker network inspect net1 on node B. 

`docker network rm net1` doesn't work here. It's related to node specific global issue above. 

```
ubuntu@ip-172-19-71-145:~$ docker network ls | grep -i net1
7e2a85265458        ip-172-19-29-201/net1   overlay             global

ubuntu@ip-172-19-71-145:~$ docker network rm net1
Error response from daemon: network net1 not found

ubuntu@ip-172-19-71-145:~$ docker network create -d overlay ip-172-19-71-145/net1
5e38ec281b894f7e1bddc96b59f4df8ced6130fec1df75749ba4a01e8060fd49

ubuntu@ip-172-19-71-145:~$ docker network ls | grep -i net1
7e2a85265458        ip-172-19-29-201/net1   overlay             global
5e38ec281b89        ip-172-19-71-145/net1   overlay             global

ubuntu@ip-172-19-71-145:~$ docker network inspect net1
[]
Error: No such network: net1
ubuntu@ip-172-19-71-145:~$
```

wsong:
FYI, @dongluochen mentioned offline that he was concerned about overlay networks showing up in `network ls` output with the form `<node name>/<network name>`. This should be impossible in most circumstances, since this only occurs for networks with scope local, which overlay networks cannot have:

https://github.com/docker/swarm/blob/master/api/handlers.go#L278

However, there are some cases (e.g. engine bugs) where overlay networks can acquire scope local:

https://github.com/docker/docker/issues/27639

This is most likely the cause of any issues that Dong saw.

wsong:
We're closing this for now because this should only be an issue with Swarm-mode overlay networks, which Swarm shouldn't really have to worry about.

PR #2539, opened 2016-11-07
Return a 500 for _ping if there is no elected primary

wsong:


api/replica.go
```diff
@@ -50,3 +54,12 @@ func (p *Replica) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		httpError(w, fmt.Sprintf("Unable to reach primary cluster manager (%s): %v", err, p.primary), http.StatusInternalServerError)
 	}
 }
+
+func (p *Replica) ping(w http.ResponseWriter, r *http.Request) {
+	if p.primary == "" {
+		httpError(w, "No elected primary cluster manager", http.StatusInternalServerError)
+		w.Write([]byte{'E', 'R', 'R', 'O', 'R'})
```
wsong: Heh. I wanted to maintain consistency with this:

https://github.com/docker/swarm/blob/master/api/handlers.go#L963

But I'll go ahead and change it to your version.

nishanttotla:
LGTM

Ping @dongluochen

dongluochen:
Essentially you are trying to change `/_ping` from local request to a cluster request. I think it can be done like below. 

```
 - var localRoutes = []string{"/info", "/_ping", "/debug"}
 + var localRoutes = []string{"/info", "/debug"}
```

wsong:
If I understand this correctly, we don't want to just blindly forward _ping requests because the primary might be down, in which case we want _ping to return a 500.

dongluochen:
If you remove `/_ping` from the `localRoutes`, it would come to this block, right? 

```
 	// Otherwise, forward.
 	if p.primary == "" {
 		httpError(w, "No elected primary cluster manager", http.StatusInternalServerError)
 		return
 	}
```

wsong:
Oh yeah, right. Done.

dongluochen:
LGTM

PR #2541, opened 2016-11-07
Give up leadership when manager shuts down

wsong:

### DIFF
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -132,7 +132,7 @@ func getDiscoveryOpt(c *cli.Context) map[string]string {
 	return options
 }
 
-func setupReplication(c *cli.Context, cluster cluster.Cluster, server *api.Server, discovery discovery.Backend, addr string, leaderTTL time.Duration, tlsConfig *tls.Config) {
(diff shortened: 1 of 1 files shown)


nishanttotla:
Several more mesos tests are failing in this one on `engine-master`, even though some of those are failing on all new PRs.

cc @dongluochen

wsong:
The test failures appear to be of the form "Failed to validate pending node: Error reading remote info: json: cannot unmarshal object into Go value of type string" Addr="127.0.0.1:5202"

This is a backwards incompatibility issue in the engine:

https://github.com/docker/docker/issues/28113

The fix would be to use v1.24 or earlier when making engine-api requests. Probably not relevant to this change, per se.

nishanttotla:
@wsong right, thanks for the info. The code seems good on a first look. Have you tested this change thoroughly?

wsong:
I tested this out manually and Swarm seemed to be able to shut down and come back up properly; it was also able to survive being killed and then restarting.

wsong:
Does this PR still look reasonable? As mentioned above, the test failures appear unrelated.

dongluochen:
LGTM

nishanttotla:
LGTM

PR #2548, opened 2016-11-21
if block ends with a return statement, so drop this else and outdent …

YuPengZTE:
Check with golint

Signed-off-by: yupeng <yu.peng36@zte.com.cn>


dongluochen:
LGTM

PR #2492, opened 2016-09-24
#2488 Update vendoring to remove engine-api

michelvocks:
This PR is a huge one. Tell me if I should split it up somehow.

**What I've done**
- Changed imports from `engine-api` to `docker/api/type` or `docker/clients`
- Updated whole vendoring (not sure if this was a good idea)
- `docker/docker/pkg/version` has been replaced upstream with `docker/api/types/versions`
- Docker upstream `docker/pkg/tlsconfig` has been updated and does not work anymore with golang 1.5 (only 1.6 and higher is supported). Function "Clone" is only implemented for 1.6 or higher. Therefore, I upgraded godeps master version to golang 1.7.
- `cluster/event_monitor.go` was changed to be compatible with the changed Events-API interface. 

**What needs to be done?**
- [X] Some tests are currently failing. Working on this.
- [x] Golang needs to be changed to 1.7 in the docs.

This fixes #2488 

ping @nishanttotla @dongluochen @allencloud 

Signed-off-by: Michel Vocks michelvocks@gmail.com

PR #2550, opened 2016-11-23
Make sure to catch connection errors properly when checking engine he…

wsong:
…alth


dongluochen:
LGTM

wsong:
Hmm, it seems https://github.com/docker/swarm/pull/2492 already incorporated this change? I guess there's nothing left to do here.

PR #2551, opened 2016-11-24
Add a node whitelist filter

alexmavr:
This PR adds support for a new scheduling filter that selects candidates based on a provided list of nodes, effectively implementing a "whitelist". This is meant to cover use cases where setting engine labels may not be practical.

The usage is currently implemented as `-e whitelist:node==node-1-name|node-2-name|node-3-name`, and I am definitely open to suggestions for a better UX.

A similar approach that is available today is to use a series of `constraint:name!=<name>` constraints to implement a "blacklist" of nodes. However, that method is not equivalent to a whitelist filter during rescheduling.

### DIFF
--- a/scheduler/filter/filter.go
+++ b/scheduler/filter/filter.go
@@ -34,6 +34,7 @@ func init() {
 		&DependencyFilter{},
 		&AffinityFilter{},
 		&ConstraintFilter{},
+		&WhitelistFilter{},
 	}
 }
 
--- a/scheduler/filter/whitelist.go
+++ b/scheduler/filter/whitelist.go
@@ -1,0 +1,71 @@
+package filter
+
+import (
+	"fmt"
+	"strings"
+
+	log "github.com/Sirupsen/logrus"
+	"github.com/docker/swarm/cluster"
+	"github.com/docker/swarm/scheduler/node"
+)
+
+// WhitelistFilter selects only nodes that are defined in a whitelist by the user
+type WhitelistFilter struct {
+}
+
+// Name returns the name of the filter
+func (f *WhitelistFilter) Name() string {
+	return "whitelist"
+}
+
+// Filter is exported
+func (f *WhitelistFilter) Filter(config *cluster.ContainerConfig, nodes []*node.Node, soft bool) ([]*node.Node, error) {
+	whitelists, err := parseExprs(config.Whitelists())
+	if err != nil {
+		return nil, err
+	}
+
+	for _, whitelist := range whitelists {
+		if !soft && whitelist.isSoft {
+			continue
+		}
+		log.Debugf("matching whitelist: %s%s%s (soft=%t)", whitelist.key, OPERATORS[whitelist.operator], whitelist.value, whitelist.isSoft)
+
+		candidates := []*node.Node{}
+
+		// Handle |-separated node names in the same whitelist
+		whiteNodes := strings.Split(whitelist.value, "|")
+
+		for _, node := range nodes {
+			switch whitelist.key {
+			// Treat all keys as "node name" keys
+			default:
+				for _, whiteNode := range whiteNodes {
+					if node.Name == whiteNode {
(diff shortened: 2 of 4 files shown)


nishanttotla:
@alexmavr can you please rebase this PR, it would be nice to have it on top of #2492.

scheduler/filter/filter.go
```diff
@@ -33,6 +33,7 @@ func init() {
 		&SlotsFilter{},
 		&DependencyFilter{},
 		&AffinityFilter{},
+		&WhitelistFilter{},
 		&ConstraintFilter{},
```
nishanttotla: Are you sure no more changes are needed in `filter.go`? I'm looking more but notice that filters are applied serially, since each filter only serves to further reduce the number of available nodes. The whitelist filter has a fundamentally opposite purpose, so it should be applied last, and the final set should be a union of the result of old filters with whitelisted nodes.

nishanttotla:
The whitelist functionality is definitely new and something that isn't supported in Swarm right now. One concern I have about it is that it puts more responsibility on the user, to make sure that (other) hard constraint requirements aren't accidentally overridden by the whitelist filter. Essentially, it comes with a warning to use at your own risk.

cc @dongluochen

nishanttotla:
@alexmavr in addition, do you mind rebasing one more time? Several tests were fixed recently.

scheduler/filter/filter.go
alexmavr: I don't see why ordering would matter, since they're all individual filters. Whitelist is not meant to circumvent any other filters of the request, but it's meant to add an additional one, in case a user wants to specify a specific subset of nodes where the rest of the filters would be applied.

I'll go ahead and put it last either way though :)

alexmavr:
Rebased and addressed line comment (moved whitelist filter last).

I still don't understand why you say this is a different case for a filter, the use case is to have all types of filters apply, but ignore any candidate nodes that are not part of the whitelisted nodes.

scheduler/filter/filter.go
nishanttotla: Ah! My bad actually, I commented before reading through the rest of the code. I misunderstood what this filter was supposed to do (whitelist to me implied that it's a whitelist of nodes indicating that regardless of the outcome of other filters, a whitelisted node should always be allowed to run the task).

I now understand, and sorry for the confusion.

scheduler/filter/whitelist.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Filter is exported
+func (f *WhitelistFilter) Filter(config *cluster.ContainerConfig, nodes []*node.Node, soft bool) ([]*node.Node, error) {
+	whitelists, err := parseExprs(config.Whitelists())
+	if err != nil {
+		return nil, err
+	}
+
+	for _, whitelist := range whitelists {
```
dongluochen: what's the use case of multiple whitelists for a `docker run` command?

dongluochen: What does it mean when running 
```
docker run --label 'com.docker.swarm.whitelists=["node==node1,node2,node3"]' --label 'com.docker.swarm.whitelists=["node==node2,node3,node4"]'
```

dongluochen:
Once we merge this change, please update Swarm doc to reflect the change.

scheduler/filter/whitelist.go
dongluochen: At a second thought, this way is more consistent with other filters like `constraints`. Let's keep it as it is.

dongluochen:
LGTM

Tests failure looks unrelated. Should investigate separately.

dongluochen:

nishanttotla:
Restarted the CI to run tests one more time.

Code LGTM.

nishanttotla:
LGTM

@alexmavr please submit a PR to update docs.

@mstanleyjones

PR #2565, opened 2016-12-07
Fix network endpoints for rescheduling

### LINKED ISSUE #2562: On-Failure move of an image breaks network alias
Reported on Swarm version 1.2.5, Engine version 1.12.1

Making a network alias with a bridge (and possibly overlay) network isn't preserved when the container gets rescheduled to a different node.

cc @dongluochen @wsong

nishanttotla:
The endpoints were being looked up from the wrong place. The fix turns out to be trivial.

cc @docker/core-swarm-maintainers @wsong @mavenugo 

Fix #2562

### DIFF
--- a/cluster/watchdog.go
+++ b/cluster/watchdog.go
@@ -94,7 +94,7 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 			name = name[1:]
 		}
 
-		if c.NetworkSettings != nil && len(c.NetworkSettings.Networks) > 0 {
+		if c.Info.NetworkSettings != nil && len(c.Info.NetworkSettings.Networks) > 0 {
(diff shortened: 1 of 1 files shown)


nishanttotla:
Test failures are unrelated.

wsong:
Will this be compatible with older engine releases? How about newer engine releases? Do these `Container` objects get populated the same way for all engines?

wsong:
This appears to fix the issue in my manual testing in UCP, for what it's worth.

nishanttotla:
@wsong this should work with older releases as well.

The issue is strictly in the manager reading off the networking config from the wrong place. It seems to have been a bug left over from #2436.

dongluochen:
LGTM

restarting tests.

PR #2568, opened 2016-12-13
Use the swarm mode node ID for the engine ID if it's available

wsong:


wsong:
If a user accidentally duplicates the `/etc/docker/key.json` file on two different machines (e.g. by cloning VMs), Swarm will complain about engines having duplicate IPs. If Swarm mode is enabled, however, we do have access to a unique ID: the Swarm node ID. Thus, this PR just uses that ID if it's available.

dongluochen:
This PR tries to resolve a specific problem that Docker ID is duplicated and Swarm mode is enabled. But I'm afraid it doesn't fit class Swarm's design. 

Swarm takes docker ID as a node's identity which is static after a Docker node is created. Swarm mode ID is not docker ID, and it's dynamic (the ID would be gone when swarm mode switches off). 

Also classic Swarm shouldn't use attributes from Docker swarm mode.

aluzzardi:
I agree with @dongluochen, but at the same time this would really help the transition path between classic and mode.

Not only it would get rid of the duplicate machine issue, but also the node IDs between classic/mode would actually match.

We know this is temporary anyway, I would ask considering for merge.

@dongluochen WDYT? Any alternatives otherwise?

dongluochen:
@aluzzardi The problem is that swarm mode ID is dynamic. If a node switches off swarm mode, this ID would be gone. If Swarm mode restarts, it'll have a new ID. This would confuse classic Swarm.

aluzzardi:
What's the worst that could happen? Those are rare events

@dhiltgen Thoughts on that?

wsong:
I believe that the Swarm classic node would be marked as offline until the Swarm classic container got restarted:

https://github.com/docker/swarm/blob/master/cluster/engine.go#L508

dongluochen:
@wsong When the ID of a node changes, the node would be put to `pending` state. You'll need to remove it from discovery, before adding it back. It was designed to resolve issues like IP change in DHCP environment. The logic is Docker ID is unique, IP might change.

wsong:
Hmm, can we maybe control this behavior with a CLI flag or something like that? In a lot of cases (e.g. when you're spawning new VMs), you're not likely to disable Swarm mode and then re-enable it while the Swarm manager is running, but you do want to keep things from breaking if you duplicate `/etc/docker/key.json` by accident.

nishanttotla:
I'm chiming in a little late on this PR, but another consideration is that it's not recommended to use classic Swarm with Swarm mode anyway, so perhaps this use case is niche enough that we could consider it?

nishanttotla:
Related: https://github.com/docker/docker/issues/30013

wsong:
Did we ever land anywhere on this PR? I'm happy to gate this behavior behind a CLI flag or something like that, but it does improve the usability of Swarm for users in cloned environments (e.g. AMIs), so I'd like to get this in.

dongluochen:
Given that we want to allow classic Swarm to run on top of Docker Swarm mode, I agree to this change.

dongluochen:
LGTM

nishanttotla:
@dongluochen lot of engine-master failures. They could possibly be related to this PR, but something seems wrong.

dongluochen:
Doesn't look related. But something could be wrong. Let's kick it again.

nishanttotla:
Tests a failing on engine-master. @wsong can you take a look?

wsong:
Tests are still failing, but the errors seem to be around starting the Docker engine, which shouldn't be affected by this change.

dongluochen:
failure on Master is unrelated to this PR. It's tracked by https://github.com/docker/swarm/issues/2646.

wsong:
Thanks for merging.

FYI, I did *not* include any logic around gating this behind a CLI flag or anything like that, so now that this PR is merged, anybody running Swarm while Swarm mode is enabled will see this behavior.

PR #2576, opened 2016-12-22
Add an option to refresh the containers cache on certain filters

wsong:

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -31,6 +31,11 @@ import (
 // APIVERSION is the API version supported by swarm manager
 const APIVERSION = "1.22"
 
+var (
+	ShouldRefreshOnNodeFilter  = false
+	ContainerNameRefreshFilter = ""
+)
+
 // GET /info
 func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
 	info := apitypes.Info{
@@ -392,6 +397,28 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	for _, value := range filters.Get("status") {
 		if value == "exited" {
 			all = true
+		}
+	}
+
+	if ShouldRefreshOnNodeFilter {
+		nodes := filters.Get("node")
+		for _, node := range nodes {
+			err := c.cluster.RefreshEngine(node)
+			if err != nil {
+				log.Debugf("could not match node filter for %s: %s", node, err)
+			}
+		}
+	}
+	if ContainerNameRefreshFilter != "" {
+		names := filters.Get("name")
+		for _, name := range names {
+			if name == ContainerNameRefreshFilter {
+				err := c.cluster.RefreshEngines()
+				if err != nil {
+					log.Debugf("names filter detected but unable to refresh all engines: %s", err)
+				}
+				break
+			}
 		}
 	}
 
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -151,4 +151,13 @@ var (
 		Value: "20s",
 		Usage: "Leader lock release time on failure",
 	}
+
+	flRefreshOnNodeFilter = cli.BoolFlag{
+		Name:  "refresh-on-node-filter",
+		Usage: "If true, refresh the cache when a ContainerList call comes in with a node filter",
+	}
+	flContainerNameRefreshFilter = cli.StringFlag{
+		Name:  "container-name-refresh-filter",
+		Usage: "If set, refresh the cache when a ContainerList call comes in with a name filter set to this value",
+	}
 )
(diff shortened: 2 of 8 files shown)


dongluochen:
Build failure. 

```
$ GOOS=linux go build
# github.com/docker/swarm/cli
cli/flags.go:158: flLeaderTTL redeclared in this block
	previous declaration at cli/flags.go:153
cli/flags.go:162: flLeaderTTL redeclared in this block
	previous declaration at cli/flags.go:158
The command "GOOS=linux go build" exited with 2.
```

wsong:
Fixed.

wsong:
Build is now successful. There appear to be some failing tests but they don't seem obviously related to this PR.

cluster/swarm/cluster.go
```diff
@@ -968,6 +968,31 @@ func (c *Cluster) BuildImage(buildContext io.Reader, buildImage *types.ImageBuil
 	return nil
 }
 
+// RefreshEngines refreshes all containers in the cluster (UCP-specific)
```
nishanttotla: @wsong is this useful in any other setting? If so, I don't think we should mention UCP or other closed-source software here.

wsong: Whoops, forgot to remove that comment. Done.

I'm not sure how useful this will be for other applications but it's nice to be able to say "give me a list of containers without using the cache" (or rather, "refresh the cache and then give the containers"). Having a magical name filter is kind of a weird way to do this, but the Docker API doesn't give us a lot of other options; we could use HTTP headers, but the Docker client libraries make it non-trivial to send specific headers for a single request.

cluster/engine.go
```diff
@@ -1422,3 +1422,14 @@ func IsConnectionError(err error) bool {
...
+
+func (e *Engine) RefreshEngine(hostname string) error {
+	if hostname != e.Name {
+		return fmt.Errorf("invalid engine name during refresh: %s vs %s", hostname, e.Name)
+	}
+	return e.RefreshContainers(true)
+}
+
+func (e *Engine) RefreshEngines() error {
+	return e.RefreshEngine(e.Name)
```
nishanttotla: @wsong the interface implies this should be a cluster function that refreshes all engines, but the definition here seems different. Can you clarify?

cluster/swarm/cluster.go
nishanttotla: I see your point, and can't think of a better way to do this given the Docker client restrictions. We should make sure to document this properly though.

cluster/engine.go
wsong: Actually, I realized that this function wasn't used, so I just removed it.

api/handlers.go
```diff
@@ -395,6 +400,32 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
...
+					log.Debugf("could not match node filter for %s: %s", node, err)
+				}
+			}
+		}
+	}
+	if ContainerNameRefreshFilter != "" {
+		names := filters.Get("name")
+		for _, name := range names {
+			if name == ContainerNameRefreshFilter {
+				if len(names) > 0 {
```
dongluochen: There shouldn't be `if len(names) > 0` inside `for _, name := range names`.

api/handlers.go
```diff
@@ -395,6 +400,32 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	if ShouldRefreshOnNodeFilter {
+		nodes := filters.Get("node")
+		if len(nodes) > 0 {
```
dongluochen: this `if` is unnecessary.

cli/flags.go
```diff
@@ -151,4 +151,13 @@ var (
 		Value: "20s",
 		Usage: "Leader lock release time on failure",
 	}
+
+	flRefreshOnNodeFilter = cli.BoolFlag{
+		Name:  "refresh-on-node-filter",
+		Usage: "If true, we will refresh the cache when a ContainerList call comes in with a node filter",
```
dongluochen: do not use `we` in command texts, verb should be good.

cli/flags.go
```diff
@@ -151,4 +151,13 @@ var (
...
+	}
+	flContainerNameRefreshFilter = cli.StringFlag{
+		Name:  "container-name-refresh-filter",
+		Usage: "If set, we will refresh the cache when a ContainerList call comes in with a name filter set to this value",
```
dongluochen: same as above.

wsong:
Addressed comments.

dongluochen:
LGTM

PR #2578, opened 2016-12-22
Cancel request when a proxy request is interrupted

### LINKED ISSUE #2567: Container logs file remains open
Hi there :hand:

When i make the following request:

```bash
curl "http://0.0.0.0:23750/containers/<container-id>/logs?stdout=1&stderr=1&follow=1" --max-time 5
```
0.0.0.0:23750 <- swarm

container logs file on node is opened 1 more time, but when timeout is reached
and connection becomes closed - the file remains open.
I tried different request configurations and got the following results:

- :no_entry_sign: - file remains open
- :ok_hand: - file is closed after some period of time

| curl params                        | follow=    | swarm           | direct daemon request |
| ------------------:                |-----------:| ---------------:|----------------------:|
| --max-time=5                       | true       | :no_entry_sign: |   :ok_hand:           |
| --max-time=5                       | false      | :ok_hand:       |   :ok_hand:           |
|                                    | false      | :ok_hand:       |   :ok_hand:           |
| --no-keepalive --max-time=5        | true       | :no_entry_sign: |   :ok_hand:           |

I used the following command to check open files:

```bash
lsof -p $(cat /run/docker.pid) | awk '{print $9 $10}' | sort | uniq -c | sort -n | grep ".*containers.*log"
```

it produces something like this:
```
1   /var/lib/docker/containers/94e48934b97db014dd9e0b89cca84e4bb332c5101e239644ba5aa19f276ea13e/94e48934b97db014dd9e0b89cca84e4bb332c5101e239644ba5aa19f276ea13e-json.log
885 /var/lib/docker/containers/eacc357e0724eb1f64781e9fb5ce202e282db4e875b10a4479d3c3bf103f8934/eacc357e0724eb1f64781e9fb5ce202e282db4e875b10a4479d3c3bf103f8934-json.log
```
where the first column is a number of openings for a certain file.

##### Versions:

```
# docker version
Client:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Wed Apr 27 00:34:42 2016
 OS/Arch:      linux/amd64

Server:
 Version:      1.11.1
 API version:  1.23
 Go version:   go1.5.4
 Git commit:   5604cbe
 Built:        Wed Apr 27 00:34:42 2016
 OS/Arch:      linux/amd64
```

```
#swarm --version
swarm version 1.2.3 (eaa53c7)
```

```
#cat /etc/redhat-release
CentOS Linux release 7.2.1511 (Core) 
```

dongluochen:
Fix #2567. 

When proxing a request, if the user connection is broken, cancel the request. 

### DIFF
--- a/api/utils.go
+++ b/api/utils.go
@@ -103,23 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request, 
 	r.URL.Host = engine.Addr
 
 	log.WithFields(log.Fields{"method": r.Method, "url": r.URL}).Debug("Proxy request")
-	resp, err := client.Do(r)
-	if err != nil {
+
+	requestDone := make(chan struct{})
+	go func() {
+		defer close(requestDone)
+		var resp *http.Response
+		resp, err = client.Do(r)
+		if err != nil {
+			return
+		}
+		// cleanup
+		defer resp.Body.Close()
+
+		copyHeader(w.Header(), resp.Header)
+		w.WriteHeader(resp.StatusCode)
+		io.Copy(NewWriteFlusher(w), resp.Body)
+
+		if callback != nil {
+			callback(resp)
+		}
+	}()
+
+	type requestCanceler interface {
+		CancelRequest(*http.Request)
+	}
+
+	var closeNotify <-chan bool
+	if closeNotifier, ok := w.(http.CloseNotifier); ok {
+		closeNotify = closeNotifier.CloseNotify()
+	}
+	select {
+	case <-closeNotify:
+		log.WithFields(log.Fields{"method": r.Method, "url": r.URL}).Debug("user connection closed")
+		if rc, ok := client.Transport.(requestCanceler); ok {
+			rc.CancelRequest(r)
+			log.WithFields(log.Fields{"method": r.Method, "url": r.URL}).Debug("request cancelled")
+		}
+		// wait for request finish
+		<-requestDone
+		return err
+	case <-requestDone:
 		return err
 	}
-
-	copyHeader(w.Header(), resp.Header)
-	w.WriteHeader(resp.StatusCode)
-	io.Copy(NewWriteFlusher(w), resp.Body)
-
-	if callback != nil {
-		callback(resp)
-	}
-
-	// cleanup
-	resp.Body.Close()
-
-	return nil
 }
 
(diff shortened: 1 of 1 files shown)


dongluochen:
ping @docker/core-swarm-maintainers.

TylerJewell:
This issue will resolve an open bug in Eclipse Che - https://github.com/eclipse/che/issues/2962#issuecomment-269488994

api/utils.go
```diff
@@ -103,23 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+
+	var closeNotify <-chan bool
+	if closeNotifier, ok := w.(http.CloseNotifier); ok {
+		closeNotify = closeNotifier.CloseNotify()
+	}
+	select {
+	case <-closeNotify:
+		log.Debugf("user connection closed")
+		if rc, ok := client.Transport.(requestCanceler); ok {
+			log.Debugf("cancel request")
```
nishanttotla: nit: this should be `canceling request` IMO

api/utils.go
```diff
@@ -103,23 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+			rc.CancelRequest(r)
+		}
+		// wait for request finish
+		<-requestDone
+		return err
```
nishanttotla: Shouldn't this be a new error, instead of just returning `err`?

api/utils.go
```diff
@@ -103,23 +103,48 @@ func proxyAsync(engine *cluster.Engine, w http.ResponseWriter, r *http.Request,
...
+	case <-requestDone:
+		return err
```
nishanttotla: And this should be `nil`?

dongluochen: This `err` from `resp, err = client.Do(r)` is right. It's the normal case.

api/utils.go
dongluochen: This one is a little tricky. The common use cases are long running streams like `docker events` or `docker logs`. There is no error when requests finish or user interrupt. So I'm returning `err` from `resp, err = client.Do(r)`.

api/utils.go
nishanttotla: You're right, this is correct.

PR #2580, opened 2016-12-27
Send network disconnect requests directly to worker engines

wsong:


wsong:
There are two parts to this change:
1. Always send network disconnect requests to their associated container's
engine, rather than a random engine. This is necessary in Swarm mode because
only the engine with the associated container knows about the container-network
connection, rather than pre-Swarm mode where all engines shared cluster state.
2. Don't translate network names to network IDs before forwarding network
disconnect requests. This is necessary because containers created with network
names in their configs don't map these names to network IDs until they get
started for the first time; thus, network disconnect requests will fail for
these containers if you try to specify a network by ID since the container's
engine won't know about the network ID.

These changes are necessary for Swarm to work when Swarm mode is enabled.

dongluochen:
> Always send network disconnect requests to their associated container's
engine, rather than a random engine. 

This was done on purpose to allow disconnect from a live node so a dead engine will not block disconnect. 
https://github.com/docker/swarm/commit/5347098defaf4671205c1f2317d83b0f18b39a59

wsong:
Updated the PR to try a random engine if we can't reach the network's associated engine.

api/handlers.go
```diff
@@ -1021,6 +1008,23 @@ func proxyNetworkDisconnect(c *context, w http.ResponseWriter, r *http.Request)
...
+			if err != nil {
+				log.Warnf("Could not get a random engine: %s", err)
+			}
+			cb = func(resp *http.Response) {
+				// force fresh networks on this engine
+				randomEngine.RefreshNetworks()
+			}
+			err = proxyAsync(randomEngine, w, r, cb)
+			if err != nil {
+				httpError(w, err.Error(), http.StatusNotFound)
```
dongluochen: Let's keep the original HTTP code.
```
```

wsong: Whoops, fixed.

wsong:
Refactored a bit to use the Docker engine API instead of just proxying HTTP requests.

api/handlers.go
```diff
@@ -994,35 +992,35 @@ func proxyNetworkDisconnect(c *context, w http.ResponseWriter, r *http.Request)
...
+	// then try a random engine if we can't connect to that engine. We
+	// try the associated engine first because on 1.12+ clusters, the
+	// network may not be known on all nodes.
+	err := engine.NetworkDisconnect(container, network, disconnect.Force)
 	if err != nil {
+		if cluster.IsConnectionError(err) && disconnect.Force && network.Scope == "global" {
+			log.Warnf("Could not connect to engine %s: %s, trying to disconnect %s from %s on a random engine", engine.Name, err, disconnect.Container, network.Name)
+			randomEngine, err := c.cluster.RANDOMENGINE()
+			if err != nil {
+				log.Warnf("Could not get a random engine: %s", err)
```
dongluochen: You should return error. `randomEngine` is `nil`.

cluster/engine.go
```diff
@@ -1402,3 +1398,23 @@ func (e *Engine) TagImage(IDOrName string, ref string, force bool) error {
 	// refresh image
 	return e.RefreshImages()
 }
+
+// NetworkDisconnect disconnects a container from a network
+func (e *Engine) NetworkDisconnect(container *Container, network *Network, force bool) error {
+	// send rename request
```
dongluochen: incorrect comment

cluster/engine.go
```diff
@@ -430,11 +430,7 @@ func (e *Engine) CheckConnectionErr(err error) {
 	// dockerclient. We need string matching for these cases. Remove the first character to deal with
 	// case sensitive issue.
 	// docker/api returns ErrConnectionFailed error, so we check for that as long as dockerclient exists
-	if err == dockerclient.ErrConnectionRefused ||
-		engineapi.IsErrConnectionFailed(err) ||
-		strings.Contains(err.Error(), "onnection refused") ||
-		strings.Contains(err.Error(), "annot connect to the docker engine endpoint") ||
-		strings.Contains(err.Error(), "annot connect to the Docker daemon") {
+	if IsConnectionError(err) {
```
dongluochen: The comments should be moved to `IsConnectionError()`.

wsong:
Addressed comments.

dongluochen:
LGTM

PR #2584, opened 2016-12-29
Remove dependency on IPv4 addresses

### LINKED ISSUE #2503: Manager fails to find addresses for an IPv6-only swarm
Hi,

I'm trying to setup an IPv6-only swarm based on consul backend. I am using Docker 1.12.1 on Linux Ubuntu.
The backend works fine, consul found the 3 nodes and the master: 

<pre>
$  consul members
 Node           Address                        Status  Type    Build  Protocol
 swarm-master   [2001:db8:7301:51::f090]:8301  alive   server  0.5.0  2
 swarm-node-01  [2001:db8:7301:51::f091]:8301  alive   client  0.5.0  2
 swarm-node-02  [2001:db8:7301:51::f092]:8301  alive   client  0.5.0  2
 swarm-node-03  [2001:db8:7301:51::f093]:8301  alive   client  0.5.0  2
</pre>

On each node, dockerd is run with `--ipv6 --fixed-cidr` options, so that swarm containers can at least connect to the consul backend.

Then I started the swarm nodes by telling them to advertise the FQDN of the node 
`--advertise=node1.fqdn:2375`
The FQDN can be resolved as IPv6 address from the master node without problem, but the swarm manage container gives as debug output:
`Failed to validate pending node: no suitable address found" Addr="node1.fqdn:2375"`

My second option was to advertise the IPv6 address of the node instead of the FQDN
`--advertise=[2001:db8:7301:51::f091]:2375`
But then the swarm manage output was:
`Failed to validate pending node: Cannot connect to the Docker daemon. Is the docker daemon running on this host?" Addr="[2001:db8:7301:51::f091]:2375"`
I checked the connection from the master node (not the swarm manage container!) to the remote docker, and it worked fine.

My (probably wrong) guesses are:
- swarm manage can not handle IPv6-only resolving FQDN
- when force with an IPv6 address for a node, swarm manage can not connect to it using IPv6

Thanks in advance for your feedback

dongluochen:
Fix #2503. 

`Connect()` in cluster/engine.go only resolves to IPv4 address. We should remove dependency on IP address family. 

```
addr, err := net.ResolveIPAddr("ip4", host)
```

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"net/http"
 	"os"
 	"runtime"
@@ -499,7 +500,7 @@ func getContainersJSON(c *context, w http.ResponseWriter, r *http.Request) {
 		tmp.Ports = make([]apitypes.Port, len(container.Ports))
 		for i, port := range container.Ports {
 			tmp.Ports[i] = port
-			if port.IP == "0.0.0.0" {
+			if ip := net.ParseIP(port.IP); ip != nil && ip.IsUnspecified() {
 				tmp.Ports[i].IP = container.Engine.IP
 			}
 		}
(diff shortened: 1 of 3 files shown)


api/handlers.go
```diff
@@ -539,7 +540,16 @@ func getContainerJSON(c *context, w http.ResponseWriter, r *http.Request) {
 	data = bytes.Replace(data, []byte("\"Name\":\"/"), []byte(fmt.Sprintf("\"Node\":%s,\"Name\":\"/", n)), -1)
 
 	// insert node IP
-	data = bytes.Replace(data, []byte("\"HostIp\":\"0.0.0.0\""), []byte(fmt.Sprintf("\"HostIp\":%q", container.Engine.IP)), -1)
+	if engineIP := net.ParseIP(container.Engine.IP); engineIP != nil {
+		var orig string
+		if engineIP.To4() != nil {
+			orig = fmt.Sprintf("\"HostIp\":\"%s\"", net.IPv4zero.String())
```
aaronlehmann: This could be a good place to use backticks:

```go
orig = fmt.Sprintf(`"HostIp":"%s"`, net.IPv4zero.String())
```

dongluochen: thanks!

PR #2585, opened 2016-12-30
Ignore volume-from attribute for rescheduling containers on node fail

xianlubird:
In some case, container has `volume-from` attribute. When reschedule this container on node fail, it can not to be created on other nodes. So we should ignore `volume-from` attribute when node failed reschedule.

@dongluochen @nishanttotla WDYT ? Maybe it can be merged in 1.2.6 millstone.

### DIFF
--- a/cluster/watchdog.go
+++ b/cluster/watchdog.go
@@ -134,6 +134,13 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 			endpointsConfig[k] = v
 		}
 		c.Config.NetworkingConfig.EndpointsConfig = endpointsConfig
+		
+		//For volume-from case, ignore it
+		if c.Config.HostConfig.VolumesFrom != nil {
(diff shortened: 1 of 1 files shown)


nishanttotla:
This change makes sense, but I wouldn't set it to `nil` without any warnings. There should at least be a log message to help with debugging, in case this change affects users.

WDYT @dongluochen?

xianlubird:
I agree with you @nishanttotla . I have changed code for adding warn log info .
Thanks

xianlubird:
@dongluochen @nishanttotla @allencloud WDYT ?

dongluochen:
IMHO I don't think we should change container configuration in rescheduling. Classic Swarm allows user to control the containers. It's possible users put the same container on every node. When a container has external dependencies, it's a `stateful` container. Classic Swarm runs assuming external dependencies are available. Making this change doesn't feel right. If the original request asks for `volume-from`, how could rescheduling doesn't require that? If `volume-from` is optional, user should update its image to deal with it.

PR #2587, opened 2017-01-05
Support more filters in network list operations

wsong:
This is necessary when doing `docker stack rm` in Docker 1.13+ engines, because
the CLI client deletes networks by doing a network list call with a label
filter and then deleting everything that gets returned. If you do this against
Swarm, then you wind up deleting all of your networks.


dongluochen:
LGTM

PR #2592, opened 2017-01-09
update go-zookeeper to rev 1d7be4e

convergentio:
Update the go-zookeeper library godep to pick up a fix for a spin in its lock implementation:
https://github.com/samuel/go-zookeeper/pull/145

This change affects docker/swarm when using ZooKeeper discovery only.


dongluochen:
thanks @convergentio! 

LGTM

PR #2593, opened 2017-01-10
Add support for the name filter when listing volumes

alexmavr:
This PR enables the `name` filter in the `/volumes` API endpoint. The implementation works in accordance to that of the daemon itself:
- Volumes are matched if any substring of the volume name matches the filter
- When multiple `name` filters are provided, each result must match at least one of them. 

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -318,7 +318,29 @@ func getVolume(c *context, w http.ResponseWriter, r *http.Request) {
 func getVolumes(c *context, w http.ResponseWriter, r *http.Request) {
 	volumesListResponse := volumetypes.VolumesListOKBody{}
 
+	// Parse filters
+	filters, err := dockerfilters.FromParam(r.URL.Query().Get("filters"))
+	if err != nil {
+		httpError(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	names := filters.Get("name")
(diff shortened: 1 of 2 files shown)


alexmavr:
@dongluochen @nishanttotla is it possible to add this to the next release milestone?

nishanttotla:
@alexmavr correct me if I'm wrong, but isn't Docker behavior to match a prefix string, as opposed to any substring?

alexmavr:
@nishanttotla it seems to be a substring, rather than a prefix:

```
$ docker volume ls --filter name=vol
DRIVER              VOLUME NAME
local               avolume
local               myvol
```

dongluochen:
@alexmavr Can you add test for this change?

alexmavr:
@dongluochen most definitely! Could you please recommend a location for such a test, as I couldn't find any existing tests that exercise similar code segments?

dongluochen:
@alexmavr You can add test to `test/integration/api/volume.bats`. You can find an example on image filtering at `./test/integration/api/images.bats:27:	run docker_swarm images --filter node=node-0`

alexmavr:
Thanks! I amended my commit to add filter coverage in the existing `docker volume ls` test.

dongluochen:
LGTM

PR #2600, opened 2017-01-18
Use the network name instead of the ID if it was passed in during disconnects

wsong:
If you try to disconnect a container from a network before that container was
started, it'll fail unless you use the network's name. This is because the
container isn't marked as connected to the network via ID until it's actually
started; until then, the daemon only keeps track of the network name as it
relates to the container.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1036,6 +1036,13 @@ func networkDisconnect(c *context, w http.ResponseWriter, r *http.Request) {
 		httpError(w, fmt.Sprintf("No such network: %s", networkid), http.StatusNotFound)
 		return
 	}
+	// If the incoming request used the network's name instead of the ID,
+	// make the request to the daemon with the network's name as well.
+	if strings.Contains(networkid, network.Name) {
+		networkid = network.Name
+	} else {
+		networkid = network.ID
+	}
 
 	// make a copy of r.Body
 	buf, _ := ioutil.ReadAll(r.Body)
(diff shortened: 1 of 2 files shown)


wsong:
This only occurs when Swarm mode is enabled, for what it's worth:

```
$ docker network create -d overlay --attachable testnet
srpcvl6cbnxc0sghi0zf9ugm9
$ docker create --net testnet busybox sleep 24h
b54c8721e323da34bad1c51734b63669848cb89f0da76b28364e016e05b28740
$ docker network disconnect srpcvl6cbnxc0sghi0zf9ugm9 b54c8721e323da34bad1c51734b63669848cb89f0da76b28364e016e05b28740
Error response from daemon: container b54c8721e323da34bad1c51734b63669848cb89f0da76b28364e016e05b28740 is not connected to the network srpcvl6cbnxc0sghi0zf9ugm9
$ docker network disconnect testnet b54c8721e323da34bad1c51734b63669848cb89f0da76b28364e016e05b28740
```

Note how disconnecting using the network name works fine, but using the network ID fails.

dongluochen:
This fix wouldn't work. https://github.com/docker/docker/blob/master/client/interface.go#L92 defines the interface as 
```
NetworkDisconnect(ctx context.Context, networkID, container string, force bool) error
```

Actually tests fail this change.

wsong:
That argument is actually misnamed; you can in fact pass in a network name:

https://github.com/docker/docker/blob/a3dff7c3dbb949662be0037de0e51c00cfa68364/daemon/container_operations.go#L984

The bug is that I wasn't stripping off the node names from the network names. Done.

wsong:
Network disconnect tests appear to be passing now.

dongluochen:
LGTM

dongluochen:
Ping @nishanttotla @allencloud

PR #2613, opened 2017-01-24
Retry container rescheduling

convergentio:
Previously, the manager attempted to reschedule containers with a policy
of on-node-failure only once.  If this single rescheduling attempt failed,
the container would never be restarted unless the user took action.

Now, the manager will retry failed rescheduling attempts every 20 seconds,
up to a maximum of 5 minutes.

At the moment the rescheduling retry policy is hard-coded; it might be
nice to expose this policy to the user via a command line option.

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -28,6 +28,7 @@ var (
 				flTLS, flTLSCaCert, flTLSCert, flTLSKey, flTLSVerify,
 				flRefreshIntervalMin, flRefreshIntervalMax, flFailureRetry, flRefreshRetry,
 				flHeartBeat,
+				flRescheduleRetry, flRescheduleRetryInterval,
 				flEnableCors,
 				flCluster, flDiscoveryOpt, flClusterOpt, flRefreshOnNodeFilter, flContainerNameRefreshFilter},
 			Action: manage,
--- a/cli/flags.go
+++ b/cli/flags.go
@@ -152,6 +152,17 @@ var (
 		Usage: "Leader lock release time on failure",
 	}
 
+	flRescheduleRetry = cli.IntFlag{
+		Name:  "reschedule-retry",
+		Value: 1,
+		Usage: "The limit on the number of attempts to reschedule a container",
+	}
+	flRescheduleRetryInterval = cli.StringFlag{
+		Name:  "reschedule-retry-interval",
+		Value: "0s",
+		Usage: "The delay between container rescheduling attempts.",
+	}
+
 	flRefreshOnNodeFilter = cli.BoolFlag{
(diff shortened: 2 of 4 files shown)


dongluochen:
@convergentio Let's discuss the design. We don't have much evidence that adding retry would help. We should discuss under what condition repeated retries would help. 

If we want to do that, I think it has to be configurable. It could be configurable at cluster level (cluster parameter defined at manager start).

convergentio:
Hi @dongluochen,

A few scenarios where the addition of retries would help with cluster robustness:

1. The attempt to start the rescheduled container fails due to a transient error in the docker engine (e.g. https://github.com/docker/docker/issues/20398).

2. The container being rescheduled had a lease on a global Docker Volume (e.g. a network accessible Volume provided by a third-party VolumeDriver plugin) that prevents the rescheduled container from mounting the volume until the lease expires.

dongluochen:
> The attempt to start the rescheduled container fails due to a transient error in the docker engine (e.g. docker/docker#20398).
The container being rescheduled had a lease on a global Docker Volume (e.g. a network accessible Volume provided by a third-party VolumeDriver plugin) that prevents the rescheduled container from mounting the volume until the lease expires.

@convergentio Thanks for the use cases! While adding retries may help the cause. There is no guarantee that your container would be running with retry. So external (human) effort is still required. Generally speaking, an event-driven model may be a good fit. When you get a container operation failure or an event, you can validate the current status, and take action accordingly. 

That said, we try to reduce user effort on frequent paths. Yes they are configured in `cluster-opt`. For example, `swarm.createretry` is to specify retries after initial failure.  

```
* swarm.createretry=0			container create retry count after initial failure
```

convergentio:
> Generally speaking, an event-driven model may be a good fit. When you get a container operation failure or an event, you can validate the current status, and take action accordingly.

Consider a swarm cluster configured with --replication.  Are you suggesting that an external application (monitor) watches the /event stream published by a swarm manager?  This would work as long as the swarm manager being watched is not running on the host that went down.  At that point there would be a race between the monitor process connecting to the /event endpoint of another swarm manager in the cluster and containers being rescheduled.  If the monitor loses the race it will not notice that a container has failed to be rescheduled.

> That said, we try to reduce user effort on frequent paths. Yes they are configured in cluster-opt. For example, swarm.createretry is to specify retries after initial failure.

Right; do you have thoughts on how to get swarm Cluster implementation to stop retrying if the manager has lost leadership in a cluster configured with --replication?  At the moment the cluster implementation is oblivious to leadership, whereas the watchdog can be started and stopped depending on leadership changes easily.

PR #2617, opened 2017-01-26
Gracefully handle container removal after start

alexmavr:
When a client uses the v1.25 API version to create a container with the `--rm` flag, the removal of the container is expected to occur server-side, rather than client-side. If a short-lived container is used, then the handler of the `ContainerStart` operation may hit a `Container Not Found` error while attempting to refresh containers.

A very easy repro for this is the following:
```
$ docker run --rm busybox echo hello
docker: Error response from daemon: Error: No such container: 72d5403456de362e8a15fdd6b25f2a17e39559b75119776985075cd
```

In this PR, I suggest a silent cache cleanup as a workaround for this issue.
Some potential enhancements could be:
- Use the `engineapi/types/container/host_config` field which includes the AutoRemove field and detect its use
- Also check if the version is 1.25 or higher.

However, I think that this proposal also accounts for a race condition that is present in older versions where one swarm client could issue a `ContainerStart` while a sideband daemon client would perform a `ContainerRemove` on that container immediately after it's started. Since the time interval is pretty thin, that's very unlikely to happen without some network-originating service interruptions

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -1344,13 +1344,13 @@ func (e *Engine) cleanupContainers() {
 }
 
 // StartContainer starts a container
-func (e *Engine) StartContainer(id string, hostConfig *dockerclient.HostConfig) error {
+func (e *Engine) StartContainer(container *Container, hostConfig *dockerclient.HostConfig) error {
 	var err error
 	if hostConfig != nil {
-		err = e.client.StartContainer(id, hostConfig)
+		err = e.client.StartContainer(container.ID, hostConfig)
 	} else {
 		// TODO(nishanttotla): Should ContainerStartOptions be provided?
-		err = e.apiClient.ContainerStart(context.Background(), id, types.ContainerStartOptions{})
+		err = e.apiClient.ContainerStart(context.Background(), container.ID, types.ContainerStartOptions{})
 	}
 	e.CheckConnectionErr(err)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


nishanttotla:
@alexmavr test failures are concerning, but not sure yet if they are related.

alexmavr:
@nishanttotla they look unrelated to me, is there any way to ask jenkins to retest them?

nishanttotla:
Yup, I'm restarting the CI now.

nishanttotla:
Seeing the CI all green after ages!

cc @dongluochen

dongluochen:
>Use the engineapi/types/container/host_config field which includes the AutoRemove field and detect its use

I prefer checking this condition. So we are backward compatible. `docker run` without `--rm` is supposed to find the container. 

>However, I think that this proposal also accounts for a race condition that is present in older versions where one swarm client could issue a ContainerStart while a sideband daemon client would perform a ContainerRemove on that container immediately after it's started. 

Such race condition is bad practice. We should report error to user.

alexmavr:
@dongluochen I added a commit which also checks whether the `HostConfig.AutoRemove` field was set to true when the container was created. I also added a simple integration test and manually confirmed that the workaround results in the expected end-to-end behavior on engine 1.13 using a 1.13 client, and that the expected debug-level log message is displayed at the logs.

Please let me know if you have any other considerations on this PR

dongluochen:
Thanks @alexmavr. Can you take a look at this test failure? 

```
not ok 145 docker start with AutoRemove
# (from function `docker' in file api/../helpers.bash, line 42,
#  from function `docker_swarm' in file api/../helpers.bash, line 53,
#  in test file api/start.bats, line 34)
```

alexmavr:
@dongluochen It looks like the `--rm` flag was moved from `docker start` to `docker create `in 1.13, which is what was causing this test to fail in engines < 1.13. I removed that test and added a `docker run --rm` test instead, which should cover all supported versions

dongluochen:
LGTM. Thanks @alexmavr!

PR #2619, opened 2017-01-27
When rescheduling containers, don't try to create them with multiple networks

wsong:
The Docker daemon does not support creating containers with multiple network
configurations, so we need to reconnect them one at a time. This was already
being done, but we were not clearing out the Config's NetworkSettings object
so the daemon was still returning an error.

I also cleared out the IP address for EndpointsSettings objects for networks
that did not have subnets, because setting an explicit IP address for
networks without subnets also results in an error from the daemon.

### DIFF
--- a/cluster/watchdog.go
+++ b/cluster/watchdog.go
@@ -107,7 +107,7 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 			clusterNetworks := w.cluster.Networks().Uniq()
 			for networkName, endpoint := range c.Info.NetworkSettings.Networks {
 				net := clusterNetworks.Get(endpoint.NetworkID)
-				if net != nil && net.Scope == "global" {
+				if net != nil && (net.Scope == "global" || net.Scope == "swarm") {
 					// record the network, they should be reconstructed on the new container
 					globalNetworks[networkName] = endpoint
 					ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
@@ -121,6 +121,19 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 			}
 		}
 
+		// Clear out the network configs that we're going to reattach
+		// later.
+		endpointsConfig := map[string]*network.EndpointSettings{}
+		for k, v := range c.Config.NetworkingConfig.EndpointsConfig {
+			net := w.cluster.Networks().Uniq().Get(v.NetworkID)
+			if net != nil && (net.Scope == "global" || net.Scope == "swarm") {
+				// These networks are already in globalNetworks
+				// and thus will be reattached later.
+				continue
+			}
+			endpointsConfig[k] = v
+		}
+		c.Config.NetworkingConfig.EndpointsConfig = endpointsConfig
 		newContainer, err := w.cluster.CreateContainer(c.Config, c.Info.Name, nil)
 		if err != nil {
 			log.Errorf("Failed to reschedule container %s: %v", c.ID, err)
@@ -133,6 +146,23 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 		// see https://github.com/docker/docker/issues/17750
 		// Add the global networks one by one
 		for networkName, endpoint := range globalNetworks {
+			hasSubnet := false
+			network := w.cluster.Networks().Uniq().Get(networkName)
+			if network != nil {
+				for _, config := range network.IPAM.Config {
+					if config.Subnet != "" {
+						hasSubnet = true
+						break
+					}
+				}
+			}
+			// If this network did not have a defined subnet, we
+			// cannot connect to it with an explicit IP address.
+			if !hasSubnet && endpoint.IPAMConfig != nil {
+				endpoint.IPAMConfig.IPv4Address = ""
+				endpoint.IPAMConfig.IPv6Address = ""
+			}
+
 			ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 			defer cancel()
 			err = newContainer.Engine.apiClient.NetworkConnect(ctx, networkName, name, endpoint)


cluster/watchdog.go
```diff
@@ -107,7 +107,7 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 			clusterNetworks := w.cluster.Networks().Uniq()
 			for networkName, endpoint := range c.Info.NetworkSettings.Networks {
 				net := clusterNetworks.Get(endpoint.NetworkID)
-				if net != nil && net.Scope == "global" {
+				if net != nil && (net.Scope == "global" || net.Scope == "swarm") {
```
nishanttotla: Why is `net.Scope == "swarm"` required here? Isn't this something we were discussing in another PR as well?

cluster/watchdog.go
```diff
@@ -121,6 +121,9 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 			}
 		}
 
+		// Clear out the network configs, since we cannot create a
+		// container with multiple networks.
+		c.Config.NetworkingConfig.EndpointsConfig = nil
```
nishanttotla: One of the unfortunate things about this piece of code is that some information, particularly endpoints are replicated in multiple locations. Make doubly sure that this works on testing. ie `c.Config.NetworkingConfig.EndpointsConfig` is the data structure you want to change.

cluster/watchdog.go
wsong: When running Swarm with Swarm mode enabled, we need to disconnect and reconnect swarm-scoped networks as well, not just global-scoped ones. swarm-scoped networks are, in effect, globally-scoped ones.

cluster/watchdog.go
wsong: It worked when I tested this manually on a 1.13 cluster.

dongluochen: When the network configs is wiped, `newContainer` wouldn't get this config. The re-created `newContainer` would have a different `NetworkingConfig.EndpointsConfig`. I suspect `newContainer` should restore the original's `Config.NetworkingConfig.EndpointsConfig`. Can you take a look?

wsong: `newContainer` is not after this function returns; we just use it to get the new container's ID and engine object.

dongluochen: The created container would not have the original's `NetworkingConfig.EndpointsConfig`, isn't it?

wsong: That's correct, but it'll get updated when we do the `NetworkConnect` calls. The reason why we can't create the new container with the `EndpointsConfig` object is that you can't create a container that's attached to multiple networks.

dongluochen: I think `NetworkConnect` would add the `EndpointsCofnig` for the network to be connected. Are there any other config that's not **global**(or **swarm**)?

wsong: It appears the only other type of scope is `local`:

https://github.com/docker/docker/blob/5497f4758d14f6f65c722838d342f634e353b08e/api/server/router/network/network_routes.go#L182
https://github.com/docker/libnetwork/blob/82905ed5312c9b3403e5266aab5c2b2dd346ce5a/datastore/datastore.go#L113

dongluochen: We need to preserve the `local` type then.

dongluochen: I mean, the `local` type configuration should be used when creating `newContainer`.

wsong: We weren't disconnecting/reconnecting networks for local networks before. In fact, doing so for the `local` networks is almost guaranteed to fail since we're rescheduling this container on a new node, which will not have the local networks the container previously relied upon.

dongluochen: Let's say the original container has a host TCP port 2388, and is connected to an overlay network `ovnet`. With this change, the original host port is not created on `newContainer`, right?

wsong: Host ports are stored in `HostConfig.PortBindings`, not in the `EndpointsConfig` map.

dongluochen: So what goes into `local` in `Config.NetworkingConfig.EndpointsConfig`?

wsong: Bridge networks can have their scope field set to `local`, indicating that they only exist on a single node.

dongluochen: That's another trouble in container rescheduling. Should we check if this container has bridge (local) network setup and refuse to reschedule it because it's node dependent, or ignore bridge network?

wsong: Prior to this change, we essentially ignored the trouble; since we attempted to preserve the network configuration, if the container was connected to a bridge network, we'd try to recreate it on a new node with that same network config. If there was a network with the same name, it'd work; otherwise, it'd fail.

I can keep around networks that do not have the `swarm` or `global` scope if you'd like this change to have as little impact as possible.

dongluochen: > keep around networks that do not have the swarm or global scope

I prefer that. After all there is default bridge network named `bridge`. And users may create same bridge network on every node.

wsong: Done.

cluster/watchdog.go
```diff
@@ -121,6 +121,18 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 			}
 		}
 
+		// Clear out the network configs that we're going to reattach
+		// later.
+		endpointsConfig := map[string]*network.EndpointSettings{}
+		for k, v := range c.Config.NetworkingConfig.EndpointsConfig {
+			_, nameExists := globalNetworks[v.Name]
```
dongluochen:
```
13.56s$ GOOS=darwin go build
cluster/watchdog.go:128: v.Name undefined (type *network.EndpointSettings has no field or method Name)
cluster/watchdog.go:129: v.ID undefined (type *network.EndpointSettings has no field or method ID)
The command "GOOS=darwin go build" exited with 2.
```

wsong: Fixed.

dongluochen:
@wsong Test failure on engine-1.9. PTAL. 

```
# [signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x61228a]
# 
# goroutine 406 [running]:
# panic(0xc17ae0, 0xc4200120a0)
# 	/usr/local/go/src/runtime/panic.go:500 +0x1a1
# github.com/docker/swarm/cluster.(*Watchdog).rescheduleContainers(0xc4201f10c0, 0xc4204202c0)
# 	/go/src/github.com/docker/swarm/cluster/watchdog.go:129 +0x55a
# created by github.com/docker/swarm/cluster.(*Watchdog).Handle
# 	/go/src/github.com/docker/swarm/cluster/watchdog.go:29 +0x12f
```

wsong:
Whoops, forgot some parens. Fixed.

dongluochen:
I'm a little puzzled by this in [engine-1.10 test](https://jenkins.dockerproject.org/job/Swarm-PRs%20(engine%201.10)/1789/console). Don't have time to dig into it now. 

```
# time="2017-02-01T22:14:49Z" level=info msg="New leader elected: 127.0.0.1:6702" 
# 2017/02/01 22:14:49 http: panic serving 127.0.0.1:41876: runtime error: invalid memory address or nil pointer dereference
# goroutine 30 [running]:
# net/http.(*conn).serve.func1(0xc420123400)
# 	/usr/local/go/src/net/http/server.go:1491 +0x12a
# panic(0xc17ae0, 0xc4200120a0)
# 	/usr/local/go/src/runtime/panic.go:458 +0x243
# github.com/docker/swarm/api.(*Replica).ServeHTTP(0xc420053a90, 0x110a1e0, 0xc420126410, 0xc4201345a0)
# 	/go/src/github.com/docker/swarm/api/replica.go:38 +0x277
# github.com/docker/swarm/api.(*dispatcher).ServeHTTP(0xc420310c40, 0x110a1e0, 0xc420126410, 0xc4201345a0)
# 	/go/src/github.com/docker/swarm/api/server.go:33 +0x5a
# net/http.serverHandler.ServeHTTP(0xc420123100, 0x110a1e0, 0xc420126410, 0xc4201345a0)
# 	/usr/local/go/src/net/http/server.go:2202 +0x7d
# net/http.(*conn).serve(0xc420123400, 0x110b0e0, 0xc4203ee740)
# 	/usr/local/go/src/net/http/server.go:1579 +0x4b7
# created by net/http.(*Server).Serve
# 	/usr/local/go/src/net/http/server.go:2293 +0x44d
```

nishanttotla:
@dongluochen you're right, this is actually a little concerning. Let me look into it. It's not related to this PR though, so we can restart CI one more time and investigate the issue in parallel.

wsong:
Looks like that failure occurred again, but I'm quite baffled as to how it's possible. The line it's failing on is:

https://github.com/docker/swarm/blob/master/api/replica.go#L38

```
p.handler.ServeHTTP(w, r)
```

But there's no way that `p.handler` could ever be nil. `Replica` objects are always instantiated with a non-nil `handler`. I suppose the receiver of that method (i.e. `p`) could be nil as well, but again, I don't see how; we should never reach a code path where either of those variables could be nil.

nishanttotla:
Uh oh, this is quite concerning. I agree that I don't see how we could have an exception on that line. I couldn't reproduce locally :-/

dongluochen:
I suspect it's related to leader change. The replica's data structure might change because of replica update, or indirectly underlayer connection change. May need to look into that. I think it's not related to this change. 

```
# time="2017-02-01T22:14:49Z" level=info msg="New leader elected: 127.0.0.1:6702" 
```

dongluochen:
I opened #2622 to track the panic. I don't see it related to this PR. 

@wsong Can you clean up the debug message in api/replica.go?

wsong:
Removed debug message.

api/replica.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+
+	log "github.com/Sirupsen/logrus"
```
dongluochen: Please revert the change to this file.

wsong: Whoops, done.

dongluochen:
LGTM

PR #2634, opened 2017-02-22
Set timeouts in HTTP clients for connecting to engines

wsong:
Previously, we were only using the timeouts in the Dial function of the HTTP
clients we use to connect to Docker engines. This means that if we can
successfully connect to a server but it hangs when it's processing a request,
we will never disconnect from that engine. This can result in the Swarm manager
using up all its file descriptors on an engine that can process HTTP requests
but which cannot return responses.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -185,7 +185,7 @@ func (e *Engine) Connect(config *tls.Config) error {
 	e.IP = addr.IP.String()
 
 	// create the HTTP Client and URL
-	httpClient, url, err := NewHTTPClientTimeout("tcp://"+e.Addr, config, time.Duration(requestTimeout), nil)
+	httpClient, url, err := NewHTTPClientTimeout("tcp://"+e.Addr, config, time.Duration(requestTimeout))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


wsong:
This change is a bit scary. As I mentioned in the initial commit, we want to implement this to handle cases where the daemon successfully accepts incoming connections but then hangs when trying to process them. I'm not sure, however, if there is a case where we actually do want a very long-lived HTTP connection (maybe image loads or something like that)?

dongluochen:
> there is a case where we actually do want a very long-lived HTTP connection (maybe image loads or something like that)

There are long-lived HTTP connections like `docker logs -f xxx`.

alexmavr:
How about we tie up the context of incoming HTTP requests to outgoing ones? that way, if an incoming http request is cancelled by the caller, the swarm manager would cancel all ongoing API calls to the engines that correspond to that request.

This would only affect synchronous calls such as logs, ps inspect etcd

dongluochen:
@alexmavr That would require adding timeouts to requests, which is not available today. 

We address caller termination problem by subscribing to caller's CloseNotifier. Here is an example, https://github.com/docker/swarm/pull/2578.

wsong:
I'm going to close this PR for now, since this doesn't seem like the right approach. I think, as Alex suggested, we'll have to make sure to pass around a single context object for each engine request that's initiated by a user request and make sure that if the user connection gets closed, we clean up the engine connections (this is similar to #2578, as Dong pointed out).

For connections initiated by Swarm itself, we'll just have to make sure they have a sensible timeout.

PR #2636, opened 2017-02-24
fix typo in swarm/clster

kevinetc123:
Signed-off-by: kevinetc123 <kaiwentan@harmonycloud.cn>


dongluochen:
@kevinetc123 Thanks for the PR. While godoc says we should include punctuation in comments, in practice it's irrelevant for single line comments. Developers from different background have their own styles. You don't need to correct this issue, especially for old code.

PR #2641, opened 2017-03-02
Added progress reporting during image pulls

wsong:


wsong:
Added progress reporting for image imports and loads as well.

wsong:
It seems like Jenkins doesn't like that we're now importing the `docker/docker/pkg/term` library (which is apparently used in the `vendor/github.com/docker/docker/pkg/jesonmessage` library); it seems like we're not using the right type of build machine, maybe?

We can switch to just vendoring in the `jsonmessage` structs directly, if you guys would prefer.

dongluochen:
@wsong It'd help if you can describe your design first. What is it trying to achieve, any change to CLI/API? 

The build fails because `docker/docker/pkg/term` uses `cgo` which requires `gcc` to build. 
```
# github.com/docker/swarm/vendor/github.com/docker/docker/pkg/term
exec: "gcc": executable file not found in $PATH
Build step 'Execute shell' marked build as failure
```

wsong:
Our goal here is just to show incremental progress notifications during very large image pulls, loads or imports.

dongluochen:
Sounds cool. Can you describe the expected output? 

FYI, there is some similarity with https://github.com/docker/docker/pull/31144. 

Swarm doesn't support `cgo` for now.

wsong:
The output should be more or less the same as doing a `docker load` against a plain engine. If you do this without Swarm, you see messages like "Pulling image..." on the CLI. This PR just passes those images through Swarm to the client.

dongluochen:
IIRC, Swarm doesn't want to support `cgo` because it introduces a lot of overhead (and maybe cross platform compilation problem). @vieux might have more insight in this.

wsong:
That's fine. Should I just vendor in the `jsonmessage` structs? We just need them to deserialize the JSON we get back from the daemon.

dongluochen:
That sounds good to me.

wsong:
All right, reverted vendor/ changes and just copy-and-pasted the structs in. PTAL.

wsong:
Looks like engine 1.9 reported progress in a different format, requiring us to deserialize the progress messages differently (in particular, it looks like they didn't have line breaks in between messages).

Would you guys be okay with just logging a warning if we fail to deserialize a progress message? This issue should only occur with very old engines, and users will just miss out on incremental progress indicators. They should still see any errors.

dongluochen:
> Would you guys be okay with just logging a warning if we fail to deserialize a progress message? 

I think this is a good compromise.

dongluochen:
LGTM

Issue #1915, opened 2016-03-01
zookeeper fails to add nodes back in integration test


dongluochen:
In [Jenkins test](https://jenkins.dockerproject.org/job/Swarm-PRs%20%28engine%20master%29/2015/console) frequently zk fails to add nodes back at (https://github.com/docker/swarm/blob/master/test/integration/discovery/zk.bats#L90). 2 nodes are expected to register, but only 1 does. 

```
  # Check that we can add instances back to the cluster
        start_docker 2                                                                                                                       
        swarm_join "$DISCOVERY"
        retry 10 1 discovery_check_swarm_info 2
```

```
not ok 56 zk discovery: node removal
# (from function `retry' in file discovery/../helpers.bash, line 80,
#  in test file discovery/zk.bats, line 90)
#   `retry 10 1 discovery_check_swarm_info 2' failed
...
#2016/03/01 22:45:35 Connected to 127.0.0.1:7873
#2016/03/01 22:45:35 Authenticated: id=95477584511303685, timeout=10000
# time="2016-03-01T22:45:35Z" level=info msg="Registered Engine node-2 at 127.0.0.1:5864" 
# time="2016-03-01T22:45:35Z" level=debug msg="Watch triggered with 2 nodes" discovery=zk 
# time="2016-03-01T22:45:35Z" level=debug msg="HTTP request received" method=GET uri="/v1.23/info" 
# Containers: 0 Running: 0 Paused: 0 Stopped: 0 Images: 0 Server Version: swarm/1.1.3 Role: primary Strategy: spread Filters: health, port, dependency, affinity, constraint 
Nodes: 1 
node-2: 127.0.0.1:5864 └ Status: Healthy └ Containers: 0 └ Reserved CPUs: 0 / 2 └ Reserved Memory: 0 B / 8.187 GiB └ Labels: executiondriver=native-0.2, kernelversion=3.13.0-79-generic, operatingsystem=Debian GNU/Linux 8 (jessie) (containerized), storagedriver=aufs └ Error: (none) └ UpdatedAt: 2016-03-01T22:45:35Z Plugins: Volume: Network: Kernel Version: 3.13.0-79-generic Operating System: linux Architecture: amd64 CPUs: 2 Total Memory: 8.187 GiB Name: 429cda8b0278 Docker Root Dir: Debug mode (client): false Debug mode (server): false Experimental: true
```

abronan:
I don't see this issue anymore on master. Let's close for now.

[...]

PR #2642, opened 2017-03-03
Move to the docker/swarm/1915-4 branch of libkv which fixes issue #1915

convergentio:
A race in the libkv zookeeper store implementation has been causing the integration tests to fail intermittently.

The race has been fixed in libkv; the corresponding pull request is here: https://github.com/docker/libkv/pull/148

This cset updates the swarm godeps to pick up the fix in libkv.


convergentio:
Sounds good.  Once the change to docker/libkv has been upstreamed I'll update this PR.

PR #2502, opened 2016-09-29
Check port conflict with pending containers

dongluochen:
Swarm checks port conflict with existing containers at create time. In concurrent container creation, it should also check **pending** containers which are under construction. This corner case is amplified by Docker compose.   

Fix #2499. 

Signed-off-by: Dong Chen dongluo.chen@docker.com

PR #2643, opened 2017-03-03
Fix bug: ports of pending containers are not filtered when scheduling

jimmycmh:
The port filter uses Info structure to identify the ports used by containers(Info.HostConfig.PortBindings, Info.NetworkSettings and Info.Config.ExposedPorts), but pending containers are initialized with empty Info. 

Signed-off-by: jimmycmh <menghui.chen@alibaba-inc.com>


dongluochen:
@jimmycmh Can you take a look at https://github.com/docker/swarm/pull/2502/? `c.Config` is used to do the port filtering. 

```
if c.Config != nil && p.compare(requested, c.Config.HostConfig.PortBindings) 
```

It'd be nice if we can have test case for it.

PR #2652, opened 2017-03-16
data race from concurrent write/flush 

dongluochen:
Concurrent write/flush to http.ResponseWriter or http.Flusher would cause data race. This change sequentializes event processing. Normally this shouldn't be a performance bottleneck. If there are too many events, we could change this to batch processing, flush when there are no more events, or a small timer (20 milliseconds) expires. 


alexmavr:
LGTM

PR #2657, opened 2017-03-22
Add node filters for network and volume lists

alexmavr:
This PR adds a `node=<hostname>` filter to the networks and volumes lists, effectively mirroring the node filter in the container list API.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -332,6 +332,7 @@ func getVolumes(c *context, w http.ResponseWriter, r *http.Request) {
 	}
 
 	names := filters.Get("name")
+	nodes := filters.Get("node")
 	for _, volume := range c.cluster.Volumes() {
 		// Check if the volume matches any name filters
 		found := false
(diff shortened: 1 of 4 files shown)


dongluochen:
The test failure on engine-1.9 is related. You can use `echo ${output}` after `run` to see the errors locally or in CI. 

```
not ok 100 docker network ls --filter node
# (in test file api/network.bats, line 51)
#   `[ "${#lines[@]}" -eq 4 ]' failed
```

alexmavr:
I can't seem to run engine-1.9 locally as `DOCKER_VERSION=1.9 ./run.sh` fails with 
```
Tag 1.9 not found in repository docker.io/dockerswarm/dind-master
```

Is there some other way to trigger the dind test runner to use a different engine version? For now, I pushed one additional commit to check the output on jenkins

dongluochen:
Try this. 

```
export DOCKER_IMAGE=dockerswarm/dind
export DOCKER_VERSION=1.9.1
./test/integration/test.sh
```

alexmavr:
@dongluochen turns out network filters were introduced in 1.10, updated the test to check for that and amended my commit. The only failing build seems to be unrelated to the changes

dongluochen:
LGTM

PR #2659, opened 2017-03-23
Set a timeout for ImageRemove, ImageList and VolumeRemove calls

wsong:
These calls are made while holding the cluster RWLock, so we need to make sure
that we don't hold that lock forever

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -585,7 +585,10 @@ func (e *Engine) RemoveImage(name string, force bool) ([]types.ImageDelete, erro
 		Force:         force,
 		PruneChildren: true,
 	}
-	dels, err := e.apiClient.ImageRemove(context.Background(), name, rmOpts)
+
+	ctx, cancelFunc := context.WithDeadline(context.Background(), time.Now().Add(time.Minute))
+	defer cancelFunc()
+	dels, err := e.apiClient.ImageRemove(ctx, name, rmOpts)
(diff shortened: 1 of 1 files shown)


wsong:

This PR fixes the issue of cluster locks being held forever when we delete images or volumes and an `ImageRemove` call hangs forever for some reason, but I'm a little confused why the `RemoveImages` and `RemoveVolumes` methods in `cluster/swarm/cluster.go` obtain the cluster lock in the first place. What is that lock supposed to be protecting?

cluster/engine.go
```diff
@@ -644,8 +650,11 @@ func (e *Engine) RemoveVolume(name string) error {
 
 // RefreshImages refreshes the list of images on the engine.
 func (e *Engine) RefreshImages() error {
+	ctx, cancelFunc := context.WithDeadline(context.Background(), time.Now().Add(time.Minute))
```
dongluochen: Let's make a constant for `ImageOperationTimeout` and use it in these functions.

dongluochen:
> why the RemoveImages and RemoveVolumes methods in cluster/swarm/cluster.go obtain the cluster lock in the first place. 

It's trying to block changes to cluster for the whole operation, not sure if that's necessary.

wsong:
Closing this in favor of https://github.com/docker/swarm/pull/2660.

PR #2660, opened 2017-03-24
cluster lock should only protect cluster data access

dongluochen:
Cluster lock should protec cluster data only, not be locked during engine operations.


dongluochen:
Need to run some tests to validate there is no dependence on this.

dongluochen:
@wsong ran some tests for this PR. 

Cluster shouldn't be locked during engine operations. The lock should be used to protect cluster data only. While each engine operation may fail, they should fail gracefully.

Ping @docker/core-swarm-maintainers.

PR #2668, opened 2017-03-30
Use engine ID + addr instead of Swarm mode ID for engine IDs

wsong:

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -496,15 +496,11 @@ func (e *Engine) updateSpecs() error {
 	// Swarm/docker identifies engine by ID. Updating ID but not updating cluster
 	// index will put the cluster into inconsistent state. If this happens, the
 	// engine should be put to pending state for re-validation.
-	var infoID string
-	if info.Swarm.NodeID != "" {
-		// Use the swarm-mode node ID if it's available, since it's
-		// guaranteed to be unique, even if the daemon has a copied
-		// /etc/docker/key.json file from another machine.
-		infoID = info.Swarm.NodeID
-	} else {
-		infoID = info.ID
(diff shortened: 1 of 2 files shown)


dongluochen:
Please take a look at CI failure.

wsong:
Seems like an unrelated issue?

```
--- FAIL: TestTagImage (0.00s)
	Error Trace:	cluster_test.go:269
	Error:		Expected nil, but got: &errors.errorString{s:"No such image: busybox"}
```

dongluochen:
It's related because https://github.com/docker/swarm/blob/master/cluster/swarm/cluster_test.go#L246 manually set engine ID. `Connect` function will generate a different ID. So the engine would be in conflicting state.

wsong:
Ah, got it. Fixed.

dongluochen:
LGTM. 

@nishanttotla This changes the format of engine ID. We should notify users at the release notes of this.

wsong:
To give a little more context for this change:

Previously, I made a change to use the Swarm mode node ID as the engine ID if Swarm mode was enabled on a cluster. This was a workaround for a common issue where a user would clone a VM with a Docker installation; if two Docker engines have the same `/etc/docker/daemon.json` file, they'll have the same engine ID, so users would frequently wind up with two nodes with the same engine ID if they cloned a VM's filesystem.

However, using the Swarm mode node ID turned out to not be a very good solution, because you can have a node leave a Swarm mode cluster and rejoin with a new ID. In this case, the Swarm manager would not allow this node to rejoin because it would detect that its ID had changed.

This PR switches us back to using the engine ID, but also appends the node address to avoid the duplicated ID problem.

PR #2697, opened 2017-05-05
event stream retry interval should be capped

dongluochen:
Docker event stream may be broken from time to time. The current backoff strategy is borrowed from engine refresh interval. The default settings is a minimum of 30 seconds. This is too long and results in event loss. Set a cap of 10 seconds for the backoff. 

cc @wsong @nishanttotla.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -217,8 +217,13 @@ func (e *Engine) StartMonitorEvents() {
 		if err := <-ec; err != nil {
 			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).WithError(err).Error("error monitoring events, will restart")
 			if !strings.Contains(err.Error(), "EOF") {
-				// failing node reconnect should use back-off strategy
-				<-e.refreshDelayer.Wait(e.getFailureCount())
+				// failing node reconnect should use back-off strategy to avoid frequent reconnect
+				retryInterval := e.getFailureCount() + 1
+				// maximum retry interval of 10 seconds
+				if retryInterval > 10 {
+					retryInterval = 10
+				}
+				<-time.After(time.Duration(retryInterval) * time.Second)
 			}
 			e.StartMonitorEvents()
 		}
@@ -1525,6 +1530,7 @@ func IsConnectionError(err error) bool {
 		strings.Contains(err.Error(), "annot connect to the Docker daemon")
 }
 
+// RefreshEngine refreshes container records from an engine
 func (e *Engine) RefreshEngine(hostname string) error {
 	if hostname != e.Name {
 		return fmt.Errorf("invalid engine name during refresh: %s vs %s", hostname, e.Name)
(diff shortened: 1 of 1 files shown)


cluster/engine.go
```diff
@@ -217,8 +217,13 @@ func (e *Engine) StartMonitorEvents() {
 		if err := <-ec; err != nil {
 			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).WithError(err).Error("error monitoring events, will restart")
 			if !strings.Contains(err.Error(), "EOF") {
-				// failing node reconnect should use back-off strategy
-				<-e.refreshDelayer.Wait(e.getFailureCount())
+				// failing node reconnect should use back-off strategy to avoid frequent reconnect
+				retryInterval := e.getFailureCount() + 1
```
wsong: I believe that, by default, `retryInterval` is always going to be at least 30, right?

https://github.com/docker/swarm/blob/master/cli/flags.go#L71

So with the default settings, this will just always be set to 10.

dongluochen: No. `e.getFailureCount()` returns how many failure attempts have happened when communicating with this engine. It's integer like 1, 2, 3 ...

wsong: Oh, I see. So this change gets rid of the random backoff logic?

https://github.com/docker/swarm/blob/master/cluster/engine.go#L99

I don't really have any idea how useful that random backoff logic is, but I feel like it might be nice to keep around.

dongluochen: The random delay was to avoid synchronization on engine reconnect. For example, hundreds of nodes reboot due to power interrupt. With the randomization, their reconnect (and subsequent refresh) would be spread out. The event stream reconnect doesn't have this problem. So I don't think randomization has effect here.

dongluochen:
ping @nishanttotla @allencloud

dongluochen:
A broken engine can respond with `EOF` continuously. Don't exclude `EOF` from backoff strategy.

nishanttotla:
@dongluochen I don't follow why we need to exclude EOF here, can you explain?

nishanttotla:
I understand the rationale of the overall change. I'm assuming the 10 second maximum is the same number the engine uses? Is there any reason to increase the maximum for larger clusters?

dongluochen:
> I don't follow why we need to exclude EOF here, can you explain?

EOF was excluded because we don't know why a daemon would send EOF when it's not being shutdown. Since the daemon is live most of the time, we were trying to reconnect right away. 

But I see some logs that the remote daemon might be responding with EOF continuously. So it should be covered by a backoff strategy, not to do a lot of meaningless retries.

dongluochen:
>  I'm assuming the 10 second maximum is the same number the engine uses? 

The engine backoff strategy is larger, multiplying `--engine-refresh-min-interval`. 

> Is there any reason to increase the maximum for larger clusters?

We change this backoff with different cluster size. This should be lightweight. If a node is offline it'd eventually be removed from discovery.

dongluochen:
Kick a rebuild.

dongluochen:
Confirm test failure are not related.

PR #2701, opened 2017-05-08
Don't throw away most of the JSONMessage content when pulling images

wsong:
Previously, we were basically crafting our own JSONMessage objects, but we
should be using the ones the engines pass back to us so that we preserve
progress information and so on.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -743,24 +743,27 @@ func postImagesCreate(c *context, w http.ResponseWriter, r *http.Request) {
 		errorFound := false
 		nonOSErrorFound := false
 		successfulPull := false
-		callback := func(what, status string, err error) {
-			if err != nil {
+		callback := func(msg cluster.JSONMessageWrapper) {
+			msg.Msg.ID = msg.EngineName
+			if msg.Msg.Status != "" {
+				msg.Msg.Status = fmt.Sprintf("Pulling %s... : %s", image, msg.Msg.Status)
+			} else {
+				msg.Msg.Status = fmt.Sprintf("Pulling %s...", image)
+			}
+			// If we get a successful pull on any node, mark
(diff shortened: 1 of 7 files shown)


api/handlers.go
```diff
@@ -743,24 +743,25 @@ func postImagesCreate(c *context, w http.ResponseWriter, r *http.Request) {
...
+		callback := func(msg cluster.JSONMessageWrapper) {
+			msg.Msg.ID = msg.EngineName
+			if msg.Msg.Status != "" {
+				msg.Msg.Status = fmt.Sprintf("Pulling %s... : %s", image, msg.Msg.Status)
+			} else {
+				msg.Msg.Status = fmt.Sprintf("Pulling %s...", image)
+			}
+			if msg.Success {
+				successfulPull = true
+			}
```
dongluochen: Are you trying to achieve `successfulPull = msg.Success`?

dongluochen:
CI failed. 

```
FAIL	github.com/docker/swarm/cluster/swarm [build failed]
```

wsong:
Build should be passing now.

api/handlers.go
dongluochen: If you intention is `successfulPull = msg.Success`, you should change it. Otherwise maybe good to add a comment on corner cases.

wsong: We want `successfulPull` to be true if `msg.Success` is true for any message, not all messages. If we did `successfulPull = msg.Success`, then `successfulPull` would just be set to the value of the most recent `msg.Success`. I've added a comment to clarify.

dongluochen:
LGTM

wsong:
Ping @nishanttotla @allencloud; can you folks take a look at this some time this week?

Issue #2664, opened 2017-03-29
Sometimes Swarm returns error Exec not found


garagatyi:
We use Swarm in [Codenvy](https://github.com/codenvy/codenvy) system built on top of [Eclipse Che](github.com/eclipse/che).
And we use Swarm instead of Docker to scale infrastructure.
We use docker execs to launch additional software in container.
But sometimes Swarm responds with message like:
`Exec b98fb0f4a66de8046125e262ca520f7f6af68783440ac0032cf4baf095626779 not found`
It happens when we try to start created exec.
It happens both under high and low load. Even on devs machines.
But I don't know how to reproduce it. It just happens from time to time.

I see [here](https://github.com/docker/swarm/blob/master/api/handlers.go#L965) that exec ID should be set but Swarm returns error [here](https://github.com/docker/swarm/blob/master/api/utils.go#L77)

This issue fixes codenvy/codenvy#1734

dongluochen:
@garagatyi When this happens, do you find any useful debug info from logs? Can you describe the sequence of your API calls?

dongluochen:
In your dev environment, can you add `GET /exec/{id}/json` to see if the exec ID is properly added? https://docs.docker.com/engine/api/v1.28/#operation/ExecInspect.

garagatyi:
I did add this call recently, but haven't investigated what it shows when erro appears. I can't provide logs details right now. I'll try to provide more details ASAP, but it'll probably take some time since I'm on sick leave.

garagatyi:
@dongluochen I haven't found anything useful in swarm logs. But we use default log level - info. Do you think we can find anything useful in debug log level logs?

Sequence of calls is:
1. Build image
1. Create container
1. Start container 
1. Connect to containers logs
1. Inspect container
1. Create exec 1
1. Start exec 1
1. Create exec 2
1. Start exec 2
....

And it is possible that this sequence is executed concurrently in some amount of threads. Though it may happen when there is a lot of threads executing this sequence I met the problem even when 1 or 2 threads are executed.
Also there are some API calls that happen in background of the system, like get containers list, get images list, get network list, remove container, remove image, remove network, listen events.

From what I can say now `GET /exec/{id}/json` when exec start fails returns non 200 code too.
We are going to add more logs and inspections in our code such as inspect container where this happened. If you have any advice on what we can log in addition please let me know.

dongluochen:
[...]

PR #2702, opened 2017-05-11
[WIP] add mutex to protect cluster.Container

dongluochen:
This is related to #2664. `cluster.Container` access is not protected. 

cc @garagatyi @wsong 


dongluochen:
It's WIP. I think #2664 is unrelated to this. It may not be needed.

PR #2721, opened 2017-06-02
Immediately flush response headers for ContainerWait requests

jlhawn:
Some Docker API endpoints (specifically the latest updates to the `ContainerWait` API in Docker API v1.30) require that response headers are immediately flushed to the client in order to have correct behavior.

Fixes an issue where `docker run ...` hangs on the latest version of Docker CLI (v17.06.0-ce-rc1).

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1484,3 +1484,25 @@ func notImplementedHandler(c *context, w http.ResponseWriter, r *http.Request) {
 func optionsHandler(c *context, w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusOK)
 }
+
+// headerFlusher is a convenient wrapper around http.ResponseWriter which
+// always flushes response headers to the client immediately.
+type headerFlusher struct {
+	http.ResponseWriter
+}
+
+func (h headerFlusher) WriteHeader(status int) {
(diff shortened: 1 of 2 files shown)


wsong:
LGTM

dongluochen:
LGTM if CI doesn't fail from this change. (This may have side effect of sending header alone while the whole response is already available. It may have performance penalty.)

dongluochen:
`volume` tests are failing for engine-1.10. Should validate it first. 

```
not ok 159 docker volume ls
# (in test file api/volume.bats, line 22)
#   `[ "${#lines[@]}" -eq 2 ]' failed
```

dongluochen:
cc @nishanttotla.

jlhawn:
I've updated the patch to only apply immediate header flushing to the `POST /containers/{id}/wait` endpoint.

dongluochen:
This is better. 
LGTM if CI doesn't fail from this. (Docker events failure is unrelated.)

jlhawn:
> Docker events failure is unrelated

Yeah, it looks like `engine-1.10` errors on that case:
```
not ok 76 docker events
```

dongluochen:
LGTM

PR #2723, opened 2017-06-05
Allow control over where volumes get created with a whitelist label

wsong:


nishanttotla:
@wsong the new test seems to fail.

wsong:
Test should be passing now. PTAL

nishanttotla:
Failling `engine-master` tests should be fixed by https://github.com/aluzzardi/dind/pull/7 (we'll need to wait for the dind image to rebuild and then kick the tests again). The issue was that download links for docker binaries changed recently, causing the existing dind image to break.

PR #2740, opened 2017-06-20
Debugging 17.03 test failures

nishanttotla:
We now have 17.03 tests in the CI, but it seems like a graph driver issue is causing them to fail. Using this PR to track and fix those failures.

### DIFF
--- a/test/integration/api/search.bats
+++ b/test/integration/api/search.bats
@@ -13,7 +13,7 @@ function teardown() {
 
 	local version="new"
 	run docker --version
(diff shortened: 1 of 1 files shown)


nishanttotla:
Issue resolved. Jenkins config had incorrect env value `DOCKER_VERSION=17.03` instead of `DOCKER_VERSION=17.03-dind`

nishanttotla:
`network inspect` test failures are fixed in https://github.com/docker/swarm/pull/2742

PR #2745, opened 2017-06-21
Add automatic constraint logic when ContainerCreate fails because of an OSType mismatch

simonferquel:
If the swarm contains hybrid nodes (ie: linux nodes and windows nodes), if the client does not specify an ostype constraint when creating a container, it fails randomly because there can be a mismatch between the image os and the engine os. 
This PR automatically detect OS mismatch when creating a container, and retries the operation with an ostype constraint.

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -1431,7 +1431,7 @@ func postBuild(c *context, w http.ResponseWriter, r *http.Request) {
 		if msg.Err != nil {
 			errorFound = true
 			errorMessage = msg.Err.Error()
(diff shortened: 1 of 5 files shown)


nishanttotla:
@simonferquel thanks for your contribution. I've added this PR to the 1.2.8 milestone. Once the upcoming release is done, I'll review and merge it (I've already gone over the code and it looks acceptable, but I need to look in more detail and wait for tests to pass).

nishanttotla:
@simonferquel could you please rebase this PR?

simonferquel:
done

Issue #2718, opened 2017-06-01
Info hangs while events are being flushed to slow readers


alexmavr:
In the events handler, swarm managers are currently holding a write lock while flushing events to listeners (https://github.com/docker/swarm/blob/master/api/events.go#L120-L142). If this flush operation takes more time than expected, such as due to a slow or inactive reader, then this causes all `/info` requests on the leader swarm manager to hang for that duration, as the handler for info attempts to calculate the number of listeners, which requires a read lock on the events handler (https://github.com/docker/swarm/blob/master/api/events.go#L149).

Steps to reproduce:
1) Provision 1 classic swarm manager and 1 docker node
2) Create several event listeners from different nodes using `docker events`, and background these processes on the shell using `Ctrl+Z`. This will maintain the TCP connection active, but no events will be actively read by these listeners.
3) Populate events on the cluster by continuously running `docker -H manager:port run --rm hello-world` with 0.1 second gap.
4) Perform a `docker -H manager:port info` and expect it to hang

Potential courses of action for a resolution:
1) Remove the number of event listeners from the response of Info. The CLI doesn't pretty-print that field, so the use case is fairly limited. Technically this is a functionality regression so it should be the least preferred resolution
2) Remove the read lock on the `Size` method of the event handler. Instead, we can maintain an integer counter that is incremented and decremented appropriately together with entries to `eh.ws`. We can use a separate lock to ensure atomicity of the `delete/decrement`, `add/increment` operations, and have the `Size` operation hold that lock instead, which will have a much smaller critical section.
3) Flush the `http.ResponseWriter` outside of the critical section of the `eventsHandler.Handle` method. This may result in events appearing after delays, or in large chunks, so I'm not sure if that's the easiest approach

aaronlehmann:
I think the right way to fix this is to maintain separate event queues for each listener (perhaps with go-events), or separate positions in a shared queue. Then each listener should have its own goroutine that reads from the queue and pushes events to the socket. Trying to flush an event to every listener before moving on to the next event is a big problem.

I don't think (1) or (2) are acceptable solutions because they would allow slow readers to block delivery of events to other event readers. (3) probably suffers from this as well unless `Handle` is reworked to write to the various sockets in separate goroutines.

cc @dongluochen @wsong

dongluochen:
> Flush the http.ResponseWriter outside of the critical section of the eventsHandler.Handle method. This may result in events appearing after delays, or in large chunks, so I'm not sure if that's the easiest approach

[...]

PR #2747, opened 2017-06-22
Moving event handling for external subscribers to use go-events

nishanttotla:
Fix https://github.com/docker/swarm/issues/2718

This PR starts to move event handling for external subscribers (API users) to use the battle-tested github.com/docker/go-events repo instead of `cluster.EventHandlers` defined internally in Swarm. To be more precise, it uses the `Queue` abstraction defined in the `watch` package in github.com/docker/swarmkit.

For now, event listening by the watchdog (for rescheduling) continues to be done by the old code.

### DIFF
--- a/api/events.go
+++ b/api/events.go
@@ -80,29 +80,31 @@ func (eh *eventsHandler) cleanupHandler(remoteAddr string) {
 
 }
 
-// Handle writes information about a cluster event to each remote address in the cluster that has been added to the events handler.
-// After an unsuccessful write to a remote address, the associated channel is closed and the address is removed from the events handler.
-func (eh *eventsHandler) Handle(e *cluster.Event) error {
+// normalizeEvent takes a cluster Event and ensures backward compatibility
+// and all the right fields filled up
+func normalizeEvent(receivedEvent *cluster.Event) ([]byte, error) {
+	// make a local copy of the event
+	e := *receivedEvent
+	// make a fresh copy of the Actor.Attributes map to prevent a race condition
+	e.Actor.Attributes = make(map[string]string)
+	for k, v := range receivedEvent.Actor.Attributes {
+		e.Actor.Attributes[k] = v
+	}
+
 	// remove this hack once 1.10 is broadly adopted
-	from := e.From
 	e.From = e.From + " node:" + e.Engine.Name
 
-	// Attributes will be nil if the event was sent by engine < 1.10
-	if e.Actor.Attributes == nil {
-		e.Actor.Attributes = make(map[string]string)
-	}
 	e.Actor.Attributes["node.name"] = e.Engine.Name
 	e.Actor.Attributes["node.id"] = e.Engine.ID
 	e.Actor.Attributes["node.addr"] = e.Engine.Addr
 	e.Actor.Attributes["node.ip"] = e.Engine.IP
 
-	data, err := json.Marshal(e)
-	e.From = from
+	data, err := json.Marshal(&e)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
-	// remove the node field once 1.10 is broadly adopted & interlock stop relying on it
+	// remove the node field once 1.10 is broadly adopted & interlock stops relying on it
 	node := fmt.Sprintf(",%q:{%q:%q,%q:%q,%q:%q,%q:%q}}",
 		"node",
 		"Name", e.Engine.Name,
@@ -114,6 +116,17 @@ func (eh *eventsHandler) Handle(e *cluster.Event) error {
 	// insert Node field
 	data = data[:len(data)-1]
 	data = append(data, []byte(node)...)
+
+	return data, nil
+}
+
+// Handle writes information about a cluster event to each remote address in the cluster that has been added to the events handler.
+// After an unsuccessful write to a remote address, the associated channel is closed and the address is removed from the events handler.
+func (eh *eventsHandler) Handle(e *cluster.Event) error {
+	data, err := normalizeEvent(e)
+	if err != nil {
+		return err
+	}
 
 	var failed []string
 
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -13,6 +13,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -41,7 +42,7 @@ var (
 func getInfo(c *context, w http.ResponseWriter, r *http.Request) {
 	info := apitypes.Info{
 		Images:            len(c.cluster.Images().Filter(cluster.ImageFilterOptions{})),
-		NEventsListener:   c.eventsHandler.Size(),
+		NEventsListener:   int(atomic.LoadUint64(c.listenerCount)),
 		Debug:             c.debug,
 		MemoryLimit:       true,
 		SwapLimit:         true,
@@ -893,6 +894,9 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
 	eventsChan, cancelFunc := c.watchQueue.Watch()
 	defer cancelFunc()
 
+	atomic.AddUint64(c.listenerCount, 1)
+	defer atomic.AddUint64(c.listenerCount, ^uint64(0))
+
 	// create timer for --until
 	var (
 		timer   *time.Timer
@@ -903,7 +907,10 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
 		timer = time.NewTimer(dur)
 		timerCh = timer.C
 	}
-	closeNotify := r.Context().Done()
+	var closeNotify <-chan bool
+	if closeNotifier, ok := w.(http.CloseNotifier); ok {
+		closeNotify = closeNotifier.CloseNotify()
+	}
 
 	for {
 		select {
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -890,8 +890,49 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
 		f.Flush()
 	}
 
-	c.eventsHandler.Add(r.RemoteAddr, w)
-	c.eventsHandler.Wait(r.RemoteAddr, until)
+	eventsChan, cancelFunc := c.watchQueue.Watch()
+	defer cancelFunc()
+
+	// create timer for --until
+	var (
+		timer   *time.Timer
+		timerCh <-chan time.Time
+	)
+	if until > 0 {
+		dur := time.Unix(until, 0).Sub(time.Now())
+		timer = time.NewTimer(dur)
+		timerCh = timer.C
+	}
+	closeNotify := r.Context().Done()
+
+	for {
+		select {
+		case eChan, ok := <-eventsChan:
+			if !ok {
+				return
+			}
+			e, ok := eChan.(*cluster.Event)
+			if !ok {
+				break
+			}
+			data, err := normalizeEvent(e)
+			if err != nil {
+				return
+			}
+			_, err = w.Write(data)
+			if err != nil {
+				log.Debugf("failed to write event to output stream %s", err.Error())
+				return
+			}
+			if f, ok := w.(http.Flusher); ok {
+				f.Flush()
+			}
+		case <-closeNotify:
+			return
+		case <-timerCh:
+			return
+		}
+	}
 }
 
 // POST /containers/{name:.*}/start
--- a/api/primary.go
+++ b/api/primary.go
@@ -3,18 +3,26 @@ package api
 import (
 	"crypto/tls"
 	"net/http"
+	"time"
 
 	"net/http/pprof"
 
 	log "github.com/Sirupsen/logrus"
 	"github.com/docker/swarm/cluster"
+	"github.com/docker/swarmkit/watch"
 	"github.com/gorilla/mux"
+)
+
+const (
+	defaultEventQueueLimit   = 10000
+	defaultEventQueueTimeout = 10 * time.Second
 )
 
 // Primary router context, used by handlers.
 type context struct {
 	cluster       cluster.Cluster
 	eventsHandler *eventsHandler
+	watchQueue    *watch.Queue
 	statusHandler StatusHandler
 	debug         bool
 	tlsConfig     *tls.Config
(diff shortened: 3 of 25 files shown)


nishanttotla:
@aaronlehmann @alexmavr this PR is a rough pattern of the changes. The relevant commit is https://github.com/docker/swarm/pull/2747/commits/d0f5ce35070d4c6eaf68130bb4689cfd0ad0f8c5 (the others are vendoring). There's a race condition somewhere that I'm trying to debug now, but let me know if this is a reasonable way to use the queue.

I have let the old event handler stay, because the rescheduling code and Mesos integration uses it, but will work on removing it later.

Still remaining:
1. Closing the queue. Thinking about the right place to do it (it should be someplace where the API server on the primary manager is shut down)
2. Handling `--until`. I still have to look up how to use that with the Queue.

api/handlers.go
```diff
@@ -890,8 +890,28 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
...
+			if e, ok := eChan.(*cluster.Event); ok {
+				fmt.Println("NEW HANDLE CALLING NORMALIZE EVENT", e.Time, e.Action, e.From)
+				if data, err := normalizeEvent(e); err != nil {
+				} else {
+					_, err := fmt.Fprint(w, string(data))
+					if f, ok := w.(http.Flusher); ok {
+						f.Flush()
+					}
+					if err != nil {
+						log.Debugf("failed to write event to output stream %s", err.Error())
```
aaronlehmann: If there was an error writing to the socket, this code should return from the function. Currently this loop will keep receiving events and trying to write them forever.

aaronlehmann:
This looks like the right approach to me.

nishanttotla:
@aaronlehmann @alexmavr I've updated the PR, and fixed all outstanding issues, barring one (getting an accurate counter for the number of listeners, which at this point will return a lower estimate).

For reference and help with reviewing, the main event handler code is in `handlers.go`: https://github.com/docker/swarm/pull/2747/files#diff-379151c15066f365334797e959e6ea94R893

The watch queue itself and related functions are part of `cluster.go`:
https://github.com/docker/swarm/pull/2747/files#diff-423777b5183d89a8b517694ebcebfd6a

The rest of it should be hopefully straightforward. Also ping @dhiltgen @wsong @aluzzardi

api/events.go
```diff
@@ -80,13 +80,15 @@ func (eh *eventsHandler) cleanupHandler(remoteAddr string) {
...
-func (eh *eventsHandler) Handle(e *cluster.Event) error {
+// normalizeEvent takes a cluster Event and ensures backward compatibility
+// and all the right fields filled up
+func normalizeEvent(e *cluster.Event) ([]byte, error) {
 	// remove this hack once 1.10 is broadly adopted
 	from := e.From
 	e.From = e.From + " node:" + e.Engine.Name
 
+	// Lock access to the attributes map
+	e.Lock()
```
aaronlehmann: Shouldn't this operate on a copy of `e`, so that the event itself (which is shared between all the subscribers) is not modified by each subscriber?

api/handlers.go
```diff
@@ -890,8 +890,47 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
 		f.Flush()
 	}
 
-	c.eventsHandler.Add(r.RemoteAddr, w)
-	c.eventsHandler.Wait(r.RemoteAddr, until)
+	eventsChan, cancelFunc := c.watchQueue.Watch()
+	defer cancelFunc()
+
+	// create timer for --until
+	timer := time.NewTimer(0)
```
aaronlehmann: We shouldn't create the timer in the case where `--until` is not specified. It would fire immediately, and the `select` would always be able to take the `<-timer.C` branch, hogging CPU.

How about something like this:

```go
var (
        timer *time.Timer
        timerCh <-chan time.Time
)
if until > 0 {
        dur := time.Unix(until, 0).Sub(time.Now())
        timer = time.NewTimer(dur)
        timerCh = timer.C
}
```

Then have a `case <-timerCh:` in the select below.

api/events.go
nishanttotla: @aaronlehmann yeah, I intended to do that, but must have forgotten. I think doing a deepcopy of the event is the way to go.

api/handlers.go
```diff
@@ -890,8 +890,47 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
...
+		closeNotify = closeNotifier.CloseNotify()
+	}
+
+	for {
+		select {
+		case eChan := <-eventsChan:
+			if e, ok := eChan.(*cluster.Event); ok {
+				if data, err := normalizeEvent(e); err != nil {
+					return
+				} else {
```
aaronlehmann: Don't need an `else` because of the `return` above.

api/handlers.go
nishanttotla: This is a better idea.

nishanttotla:
@aaronlehmann addressed your comments.

api/events.go
```diff
@@ -80,29 +80,34 @@ func (eh *eventsHandler) cleanupHandler(remoteAddr string) {
...
-	}
 	e.Actor.Attributes["node.name"] = e.Engine.Name
 	e.Actor.Attributes["node.id"] = e.Engine.ID
 	e.Actor.Attributes["node.addr"] = e.Engine.Addr
 	e.Actor.Attributes["node.ip"] = e.Engine.IP
 
-	data, err := json.Marshal(e)
+	data, err := json.Marshal(&e)
+
 	e.From = from
```
aaronlehmann: This line seems to be a noop. It happens after the marshalling, so it doesn't affect the data returned by this function.

Since the old code used to mutate the event, let's make sure nothing depends on this being set in the `Event` structure.

api/events.go
```diff
@@ -115,6 +120,17 @@ func (eh *eventsHandler) Handle(e *cluster.Event) error {
 	data = data[:len(data)-1]
 	data = append(data, []byte(node)...)
```
aaronlehmann: Unrelated to the changes in this PR, but this is really weird. Not sure why `node` can't just be a struct field that gets serialized the normal way.

nishanttotla: I'm not sure about this myself, but my guess is that this is related to backward compatibility. It's been around long enough that it should be removable now though. I can address that later.

api/events.go
aaronlehmann: Oh, I guess this was actually to restore the `From` field to the original contents, so it looks like it's not needed anymore, since we're acting on a copy.

nishanttotla: You have a point, but this code is weird. I suspect that this line was needed to reset the changes done by
```
	from := e.From
	e.From = e.From + " node:" + e.Engine.Name
```
above. Since we're not mutating the event anymore, it shouldn't be required to set back to the old value anymore. I'll remove this.

aaronlehmann:
LGTM (not a maintainer)

aaronlehmann:
Note that this should probably use the timeout and queue size limit features introduced in https://github.com/docker/swarmkit/pull/2285 when that PR is ready. That would protect us against introducing out-of-memory regressions in situations where listeners are too slow to consume events, and those events would pile up.

api/events.go
```diff
@@ -80,29 +80,31 @@ func (eh *eventsHandler) cleanupHandler(remoteAddr string) {
...
 }
 
-// Handle writes information about a cluster event to each remote address in the cluster that has been added to the events handler.
-// After an unsuccessful write to a remote address, the associated channel is closed and the address is removed from the events handler.
-func (eh *eventsHandler) Handle(e *cluster.Event) error {
+// normalizeEvent takes a cluster Event and ensures backward compatibility
+// and all the right fields filled up
+func normalizeEvent(receivedEvent *cluster.Event) ([]byte, error) {
+	// make a local copy of the event
+	e := *receivedEvent
```
stevvooe: Are there any slices or maps here?

nishanttotla: @stevvooe yes, there's only one map that's written to. For that, I've done a deepcopy in the lines below to prevent a race condition.

api/handlers.go
```diff
@@ -890,8 +890,47 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
...
+	var closeNotify <-chan bool
+	if closeNotifier, ok := w.(http.CloseNotifier); ok {
+		closeNotify = closeNotifier.CloseNotify()
+	}
+
+	for {
+		select {
+		case eChan := <-eventsChan:
+			if e, ok := eChan.(*cluster.Event); ok {
+				data, err := normalizeEvent(e)
```
stevvooe: Move the "meat" outside the if and just break on getting the wrong object.

api/handlers.go
```diff
@@ -890,8 +890,47 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
...
+		timerCh = timer.C
+	}
+	var closeNotify <-chan bool
+	if closeNotifier, ok := w.(http.CloseNotifier); ok {
+		closeNotify = closeNotifier.CloseNotify()
+	}
+
+	for {
+		select {
+		case eChan := <-eventsChan:
```
stevvooe: Is `eChan` not an event?

api/handlers.go
```diff
@@ -890,8 +890,47 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
...
+			if e, ok := eChan.(*cluster.Event); ok {
+				data, err := normalizeEvent(e)
+				if err != nil {
+					return
+				}
+				_, err = fmt.Fprint(w, string(data))
```
stevvooe: Why not write directly here, rather than reallocate the data as a string? `w.Write(data)` should be sufficient.

api/handlers.go
```diff
@@ -890,8 +890,47 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
...
+		timerCh <-chan time.Time
+	)
+	if until > 0 {
+		dur := time.Unix(until, 0).Sub(time.Now())
+		timer = time.NewTimer(dur)
+		timerCh = timer.C
+	}
+	var closeNotify <-chan bool
+	if closeNotifier, ok := w.(http.CloseNotifier); ok {
+		closeNotify = closeNotifier.CloseNotify()
```
stevvooe: Depending on go version, you can now use `r.Context().Done()` to listen to this cancellation.

api/handlers.go
stevvooe: @aaronlehmann Does this need "closed" detection? For example, will the watch ever shutdown this channel?

aaronlehmann: Currently it will not, but https://github.com/docker/swarmkit/pull/2285 adds this as an option. So it probably makes sense to add the check here in anticipation of using that later.

cluster/mesos/cluster.go
```diff
@@ -168,12 +170,20 @@ func NewCluster(scheduler *scheduler.Scheduler, TLSConfig *tls.Config, master st
 
 // Handle callbacks for the events
 func (c *Cluster) Handle(e *cluster.Event) error {
+	// publish event to the watchQueue for external subscribers
+	c.watchQueue.Publish(e)
```
stevvooe: Why don't we adapt this to the eventHandlers list? I am not sure that making this dependent on the watch queue makes sense.

api/events.go
stevvooe: While it is slightly wasteful if you're not writing to it, it might be worthwhile to just copy the whole structure here to avoid future bugs. These can be nasty in practice.

cluster/mesos/cluster.go
nishanttotla: @stevvooe I agree, but I also think that the eventual queue we will used will be moved upstream to go-events. This is supposed to be a temporary fix.

`eventHandlers' will go away actually, and all events should be handled using the watchQueue. For now though, some internal rescheduling functionality uses events which I did not want to mess with in this PR, that's why we have this case.

api/handlers.go
nishanttotla: `eChan` has an interface type, which we cast to `*cluster.Event` below.

api/handlers.go
nishanttotla: I'm not sure the request context is set correctly in all cases in Swarm, some of the code is from before contexts were introduced in Go. I'll check.

cluster/mesos/cluster.go
stevvooe: @nishanttotla That still doesn't explain why this isn't just a handler. You can do this with a closure to adapt `Publish` to `Handler`.

api/handlers.go
stevvooe: This should be from `net/http`, so should be wired to the `CloseNotifier`.

cluster/mesos/cluster.go
nishanttotla: @stevvooe not sure I follow exactly here. What do you envision this handler to look like?

api/primary.go
```diff
@@ -122,16 +124,22 @@ func NewPrimary(cluster cluster.Cluster, tlsConfig *tls.Config, status StatusHan
 	// The new eventsHandler is initialized with no writers or channels.
 	// This is in api/events.go
 	eventsHandler := newEventsHandler()
+	// eventsQueue uses the watch package from SwarmKit, which is based on
+	// the go-events package. See https://github.com/docker/swarm/issues/2718
+	// for context
+	eventsQueue := watch.NewQueue()
```
alexmavr: Now that https://github.com/docker/swarmkit/pull/2285 is merged, let's revendor swarmkit and change this constructor to:
```
NewQueue(WithTimeout(10 * time.Second), WithLimit(10000), WithCloseOutChan())
```

We probably want to define these values as config vars and plumb them through the standard config model used by classic swarm. 
Assuming 1KB per event (gross overestimation), the size limit of `10000` for the queue should occupy no more than 10MB per listener in memory. For the edge case of 100 simultaneous inefficient listeners, this would lead up to to an additional 1 GB of memory

nishanttotla: @alexmavr done. PTAL.

api/handlers.go
```diff
@@ -890,8 +890,49 @@ func getEvents(c *context, w http.ResponseWriter, r *http.Request) {
...
+		select {
+		case eChan, ok := <-eventsChan:
+			if !ok {
+				return
+			}
+			e, ok := eChan.(*cluster.Event)
+			if !ok {
+				break
+			}
+			data, err := normalizeEvent(e)
```
alexmavr: Should we consider event filtering here?

nishanttotla: @alexmavr maybe we could do it as a follow up?

nishanttotla:
@alexmavr added a counter to track number of active listeners, PTAL.

nishanttotla:
Merging this now, since it's been reviewed and it worked in testing, both manual and `docker-e2e`. Any downstream issues can be fixed in a follow-up.

cluster/mesos/cluster.go
stevvooe: Not sure what you mean: just make an event handler. I am not sure how this got merged in its current state. It just doesn't make any sense to have both a handler set and a publish set. I'd recommend cleaning this up before it gets too far. It is going to be a pain to remove later and will confuse future contributors.

stevvooe:
@nishanttotla I'd recommend taking time to cleanup the double handler logic. That will result in a bug in the future if one adds some to the task queue and the handler set.

cluster/mesos/cluster.go
nishanttotla: I will clean that up soon. The handler will entirely go away as it is, and only the publish set will remain.

PR #2751, opened 2017-06-28
Support asterisks in image reference filter

wsong:


nishanttotla:
Let's take this PR in for the next release (not the upcoming one). This will need to be rebased once for the updated API.

One thing that concerns me here is that here's a lot of old code (pre `github.comdocker/distribution`) that directly handles image strings. I want to be extra careful that mixing that and the `distribution` package will not cause problems.

nishanttotla:
@wsong can you check if the test failures are related?

wsong:
Test failures all appear unrelated

PR #2759, opened 2017-07-07
Make sure we don't read from Request.Body after hijacking

wsong:
Apparently, reading from Request.Body after a hijack is not allowed:

https://github.com/golang/go/issues/20933

So we need to make sure that we stop messing with the request after the
hijack call. I also ensured that we read any buffered data that might be there.


nishanttotla:
@wsong can you rebase?

wsong:
Rebased.

PR #2760, opened 2017-07-10
Make sure that correct API versions are set on the client

nishanttotla:
Docker recently made changes around how it handles setting of API versions. This was causing Swarm versions to be stuck at 1.24.

This change fixes that.


nishanttotla:
Ping @allencloud

nishanttotla:
https://github.com/docker/swarm/pull/2756 should be merged before this PR, because that PR completely removes all uses of the `samalba/dockerclient` client, which is still stuck on API version 1.21.

PR #2767, opened 2017-07-12
Remove OldContainerConfig struct

nishanttotla:
This was created to facilitate #1879 but it doesn't make sense to keep around anymore, after Docker versions 1.6/1.7 are long deprecated.


nishanttotla:
@thaJeztah it seems like API version in integration tests need to be updated. That explains why some of these tests are failing consistently.

PR #2771, opened 2017-07-18
Supports image load on hybrid cluster

simonferquel:
On hybrid clusters (with linux and windows nodes), `docker load` always fail because a linux node cannot load a windows image and vice-versa.

This fix the issue by parsing errors in the engine output, and ignoring OS mismatch errors

### DIFF
--- a/cluster/swarm/utils.go
+++ b/cluster/swarm/utils.go
@@ -1,6 +1,7 @@
 package swarm
 
 import (
+	"regexp"
 	"strings"
 )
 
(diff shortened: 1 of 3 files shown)


cluster/swarm/utils.go
```diff
@@ -33,3 +34,13 @@ func convertMapToKVStrings(values map[string]*string) []string {
 	}
 	return result
 }
+
+var imageEngineOSErrorPattern = regexp.MustCompile(`cannot load (.+) image on (.+)`)
```
nishanttotla: @simonferquel @wsong is this pattern too general for image loading? I'm not aware of load errors so we should be sure about this.

wsong: What do you mean too general? Are you worried that this will catch other errors besides the ones we care about?

nishanttotla: @wsong yes, that's what I'm worried about.

wsong: To the best of my knowledge, that error only occurs during image loads for operating system mismatches, but perhaps the engine folks would know more.

nishanttotla: Seems like this specific text comes only from OS mismatch errors: https://github.com/moby/moby/blob/master/image/tarexport/load.go#L428

This error could change and break this issue, but the engine error text is also somewhat ad-hoc, so this is probably the best we can do.

Issue #2774, opened 2017-07-20
Proposal: Docker Swarm backed by SwarmKit


nishanttotla:
# Proposal

Docker Swarm (aka Swarm Classic) and SwarmKit are tools for container orchestration. Swarm requires the use of an external discovery mechanism for its functioning. Commonly used discovery backends include etcd, consul, and Docker Hub discovery tokens (not recommended in production). SwarmKit on the other hand has discovery built in, and does not require the setting up of a separate discovery backend.

Some projects might want to use both Swarm and SwarmKit, which means that in addition to having SwarmKit’s inbuilt discovery, it is still required to set up an external discovery mechanism (etcd), which adds complexity and points of failure. Moreover, Swarm and SwarmKit might report node states differently, leading to inconsistent views of the cluster.

This issue describes a proposal to make SwarmKit the default discovery backend for Swarm

# High-level UX

Swarm managers are Go processes, and they read cluster membership information off a discovery backend. Starting a Swarm manager currently looks like

```
./swarm manage -H 192.168.99.103:2376 --discovery-opt kv.path=docker/nodes consul://192.168.99.103:8500
```

where `consul://192.168.99.103:8500` specifies the IP address and port for the discovery backend. Swarm agents (which are backed by a Docker Engine each) can then be run to advertise the corresponding Engine’s address to the discovery backend.

```
docker run -d swarm join --advertise=172.30.0.69:2375 consul://192.168.99.103:8500
```

This may not be required if the Engine itself advertises to the backend. The key is that the discovery should be able to access the list of agent Engines.

Since SwarmKit (and its inbuilt discovery) is integrated into the Docker Engine, we require that each Swarm manager be backed by a corresponding Docker Engine, that is also a SwarmKit manager in the cluster. We call it the manager’s “local” Engine. With that requirement, the Swarm manager would be passed a Docker socket so that it can directly query the Docker API at that engine. The workflow may look something like

```
./swarm manage -H 192.168.99.103:2376 unix:///var/run/docker.sock
```

This means the Swarm manager is listening for requests at 192.168.99.103:2376, and using the Docker socket of its “local” Engine for discovery. This may become the default way to run Swarm, as token based discovery is deprecated (https://github.com/docker/swarm/pull/2743) and not considered safe for production.

# Design Details

Here are some high-level design details for how this would be implemented.

[...]

PR #2775, opened 2017-07-20
Adding a SwarmKit based discovery backend

nishanttotla:
Addresses https://github.com/docker/swarm/issues/2774

To start a manager with swarmkit discovery
```
swarm manage -H swarm_manager_ip:port --discovery-engine-port 2377 swarmkit://local_engine_ip:port
```

The default value for `discovery-engine-port` is 2375.

TODO
- [x] Provide a discovery option for port in case agents are exposed on a non-default port
- [x] Add support for cluster level events (or the case when the local engine stops being a manager) [follow-up]
- [x] Handle failure or disconnect of Event stream
- [ ] Make sure connection also works with the docker socket [follow-up]

### DIFF
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -104,7 +104,7 @@ func setupSwarmKitDiscovery(uri string, heartbeat time.Duration, ttl time.Durati
 }
 
 // Initialize the discovery service.
-func createDiscovery(uri string, c *cli.Context) discovery.Backend {
+func createDiscovery(uri string, c *cli.Context, discoveryEnginePort int) discovery.Backend {
 	// this initializes and sets up a SwarmKit discovery backend in case it
 	// is used. The discovery package handles updating the parameters
 	// when discovery.New() is called, but it requires "swarmkit" to be
@@ -123,6 +123,11 @@ func createDiscovery(uri string, c *cli.Context) discovery.Backend {
 	discovery, err := discovery.New(uri, hb, 0, getDiscoveryOpt(c))
 	if err != nil {
 		log.Fatal(err)
+	}
+
+	// if swarmkit based discovery is being used, set the uniform agent port
+	if swarmkitDiscovery, ok := discovery.(*swarmkit.Discovery); ok {
+		swarmkitDiscovery.SetAgentPort(discoveryEnginePort)
 	}
 
 	return discovery
@@ -281,11 +286,16 @@ func manage(c *cli.Context) {
 		FailureRetry:       failureRetry,
 	}
 
+	discoveryEnginePort := c.Int("discovery-engine-port")
+	if discoveryEnginePort <= 0 {
+		log.Fatal("invalid swarmkit discovery port")
+	}
+
 	uri := getDiscovery(c)
 	if uri == "" {
 		log.Fatalf("discovery required to manage a cluster. See '%s manage --help'.", c.App.Name)
 	}
-	discovery := createDiscovery(uri, c)
+	discovery := createDiscovery(uri, c, discoveryEnginePort)
 	s, err := strategy.New(c.String("strategy"))
 	if err != nil {
 		log.Fatal(err)
--- a/cli/manage.go
+++ b/cli/manage.go
@@ -17,6 +17,7 @@ import (
 	"github.com/docker/swarm/cluster"
 	"github.com/docker/swarm/cluster/mesos"
 	"github.com/docker/swarm/cluster/swarm"
+	"github.com/docker/swarm/discovery/swarmkit"
 	"github.com/docker/swarm/scheduler"
 	"github.com/docker/swarm/scheduler/filter"
 	"github.com/docker/swarm/scheduler/strategy"
@@ -97,8 +98,19 @@ func loadTLSConfig(ca, cert, key string, verify bool) (*tls.Config, error) {
 	return config, nil
 }
 
+func setupSwarmKitDiscovery(uri string, heartbeat time.Duration, ttl time.Duration, m map[string]string) error {
+	discovery := &swarmkit.Discovery{}
+	return discovery.Initialize(uri, heartbeat, ttl, m)
+}
+
 // Initialize the discovery service.
 func createDiscovery(uri string, c *cli.Context) discovery.Backend {
+	// this initializes and sets up a SwarmKit discovery backend in case it
+	// is used. The discovery package handles updating the parameters
+	// when discovery.New() is called, but it requires "swarmkit" to be
+	// present as a discovery backend.
+	setupSwarmKitDiscovery("", 0, 0, nil)
+
 	hb, err := time.ParseDuration(c.String("heartbeat"))
 	if err != nil {
 		log.Fatalf("invalid --heartbeat: %v", err)
@@ -134,6 +146,9 @@ func getDiscoveryOpt(c *cli.Context) map[string]string {
 
 func getCandidateAndFollower(discovery discovery.Backend, addr string, leaderTTL time.Duration) (*leadership.Candidate, *leadership.Follower) {
 	kvDiscovery, ok := discovery.(*kvdiscovery.Discovery)
+	// TODO(swarmkitdiscovery): Here we should also check if this is a
+	// swarmkit discovery object, and allow for leader election when
+	// it is supported by it
 	if !ok {
 		log.Fatal("Leader election is only supported with consul, etcd and zookeeper discovery.")
 	}
--- a/discovery/swarmkit/swarmkit.go
+++ b/discovery/swarmkit/swarmkit.go
@@ -10,6 +10,7 @@ import (
 	"golang.org/x/net/context"
 
 	"github.com/docker/docker/api/types"
+	"github.com/docker/docker/api/types/filters"
 	"github.com/docker/docker/api/types/swarm"
 	"github.com/docker/docker/client"
 	"github.com/docker/docker/pkg/discovery"
@@ -52,7 +53,7 @@ func (s *Discovery) Initialize(uri string, heartbeat time.Duration, ttl time.Dur
 	s.heartbeat = heartbeat
 	s.ttl = ttl
 
-	// create a client to connect with the Swarm Mode API
+	// create a client to connect with the Docker API
 	httpClient, _, err := cluster.NewHTTPClientTimeout("tcp://"+uri, nil, time.Duration(30*time.Second), nil)
 	if err != nil {
 		return err
@@ -95,9 +96,19 @@ func (s *Discovery) Watch(stopCh <-chan struct{}) (<-chan discovery.Entries, <-c
 	ticker := time.NewTicker(s.heartbeat)
 	errCh := make(chan error)
 
+	// eventsOptions sets a filter to only retrieve cluster level events
+	eventFilter := filters.NewArgs()
+	eventFilter.Add("type", "node")
+	eventsOptions := types.EventsOptions{
+		Filters: eventFilter,
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	eventsChan, eventsErrChan := s.apiClient.Events(ctx, eventsOptions)
+
 	go func() {
 		defer close(ch)
 		defer close(errCh)
+		defer cancel()
 
 		// Send the initial entries if available.
 		currentEntries, err := s.fetch()
@@ -107,7 +118,7 @@ func (s *Discovery) Watch(stopCh <-chan struct{}) (<-chan discovery.Entries, <-c
 			ch <- currentEntries
 		}
 
-		// Periodically send updates.
+		// Periodically send updates, unless an event is received
 		for {
 			select {
 			case <-ticker.C:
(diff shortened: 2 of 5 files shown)


nishanttotla:
cc @wsong @alexmavr do you have initial comments on this?

Obviously more work is needed in addition to the TODO list above, but this works for the case of a single manager when leader election is not required. In a follow-up PR, I will add a subscription to the events stream to look for node inventory changes which will allow us to immediately access any updates without having to depend on the discovery heartbeat.

fbuecklers:
It would be cool if swarmkit based discovery could also support [node availability](https://docs.docker.com/engine/swarm/manage-nodes/#change-node-availability) so that running task are still visible, but new containers aren't scheduled on the node. 
That allows to remove nodes from.the cluster seamensly.

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,138 @@
...
+	// the uri may be a manager address or a docker socket
+	if uri == "" {
+		return errors.New("Swarm Mode manager address not provided")
+	}
+	s.uri = uri
+	s.heartbeat = heartbeat
+	s.ttl = ttl
+
+	// create a client to connect with the Swarm Mode API
+	httpClient, _, err := cluster.NewHTTPClientTimeout("tcp://"+uri, nil, time.Duration(30*time.Second), nil)
```
wsong: So with this approach, each Swarm manager is tied to a specific Swarmkit manager, right? And if that Swarmkit manager goes down, discovery will no longer work on this manager, right?

nishanttotla: @wsong that is accurate (outlined in #2774). It's pretty much an addition to the SwarmKit manager (and could potentially be run as a service).

nishanttotla:
@alexmavr @wsong would like to get your reviews. This PR currently also looks at node events and immediately updates the node inventory when something happens rather than wait for the discovery period.

nishanttotla:
@wsong that's right, leader election isn't part of this PR. I've been working on it on the side, but haven't opened a PR yet. I think it is okay to do it in steps.

cli/manage.go
```diff
@@ -97,8 +98,19 @@ func loadTLSConfig(ca, cert, key string, verify bool) (*tls.Config, error) {
...
+}
+
 // Initialize the discovery service.
-func createDiscovery(uri string, c *cli.Context) discovery.Backend {
+func createDiscovery(uri string, c *cli.Context, discoveryEnginePort int) discovery.Backend {
+	// this initializes and sets up a SwarmKit discovery backend in case it
+	// is used. The discovery package handles updating the parameters
+	// when discovery.New() is called, but it requires "swarmkit" to be
+	// present as a discovery backend.
+	setupSwarmKitDiscovery("", 0, 0, nil)
```
alexmavr: Why do we call this here? couldn't we wrap the `discovery.New` method in a way that detects that the backend is swarmkit and performs the initialization at that time?

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,162 @@
...
+)
+
+const defaultAgentPort = "2375"
+
+// Discovery is exported.
+type Discovery struct {
+	heartbeat  time.Duration
+	ttl        time.Duration
+	httpClient *http.Client
+	apiClient  *client.Client
```
alexmavr: nit: We could use the `CommonAPIClient` interface instead

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,162 @@
...
+	nodeList, err := s.apiClient.NodeList(ctx, types.NodeListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("Failed to fetch entries, Discovery service returned %s", err.Error())
+	}
+
+	var addrs []string
+
+	for _, n := range nodeList {
+		// report all nodes that are not in drain mode
+		if n.Spec.Availability != swarm.NodeAvailability("drain") {
```
alexmavr: I'm not sure whether we should exclude nodes that are on Drain from the node inventory entirely. This could lead to mis-reporting of the inventory and operational confusion. If we want to exclude "drain" nodes from scheduling, we should consider adding a separate mechanism that abstracts this behavior out. For example, a node could be marked as "excluded" in the memory of the swarm manager, and a scheduling filter can prevent scheduling on excluded nodes. This will still allow us to view resources present on that node and consider it part of the inventory. The `Watch` method below could make sure to mark/unmark excluded nodes as that information becomes available through the events stream.

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,162 @@
...
+
+				// Check if the list of nodes has really changed.
+				if !newEntries.Equals(currentEntries) {
+					ch <- newEntries
+				}
+				currentEntries = newEntries
+			case err := <-eventsErrChan:
+				errCh <- err
+				continue
+			case <-stopCh:
```
alexmavr: The context of the `Events` call above should be cancelled when a value is read from `stopCh`, otherwise we'll be leaking a TCP connection and a goroutine

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,162 @@
...
+func (s *Discovery) Initialize(uri string, heartbeat time.Duration, ttl time.Duration, _ map[string]string) error {
+	// the uri may be a manager address or a docker socket
+	if uri == "" {
+		return errors.New("Swarm Mode manager address not provided")
+	}
+	s.uri = uri
+	s.heartbeat = heartbeat
+	s.ttl = ttl
+
+	// create a client to connect with the Swarm Mode API
```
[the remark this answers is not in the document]
nishanttotla: Good point. I think in this line though, it should just be called Docker API.

discovery/swarmkit/swarmkit.go
nishanttotla: I think it really depends on what kind of usage we wish to expose with the inventory. I see your point that it's worthwhile giving users full access to prevent confusion. Do you also think it's not the job of Swarm classic to be respecting drain constraints, since this is not a Swarm classic abstraction per se?

cli/manage.go
nishanttotla: I suppose we can, I just followed the pattern used by hub based discovery. Maybe that can do with a refactor as well.

discovery/swarmkit/swarmkit.go
fbuecklers: I hope that is okay, that I add my opinion here. I am really waiting for this feature. Our use case is that we run a custom healtchecker periodicaly over all running containers to see if they are still up to reschedule them if a node or a process fails. But for rolling node updates it is really cool to mark a node as draining. If the node drops immediately when it is marked as draining our healtchecker will reschedule all containers immediately. That's really bad for us since we do not have a phase where we actually can drain the node (rescheduling all running containers).

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,162 @@
...
+func init() {
+	Init()
+}
+
+// Init is exported.
+func Init() {
+	discovery.Register("swarmkit", &Discovery{})
+}
+
+func (s *Discovery) SetAgentPort(discoveryEnginePort int) {
```
vikstrous: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,162 @@
...
+					errCh <- err
+					continue
+				}
+
+				// Check if the list of nodes has really changed.
+				if !newEntries.Equals(currentEntries) {
+					ch <- newEntries
+				}
+				currentEntries = newEntries
+			case <-eventsChan:
```
vikstrous: Why discard the events and not use them?

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,162 @@
...
+	ticker := time.NewTicker(s.heartbeat)
+	errCh := make(chan error)
+
+	// eventsOptions sets a filter to only retrieve cluster level events
+	eventFilter := filters.NewArgs()
+	eventFilter.Add("type", "node")
+	eventsOptions := types.EventsOptions{
+		Filters: eventFilter,
+	}
+	eventsChan, eventsErrChan := s.apiClient.Events(context.Background(), eventsOptions)
```
vikstrous: What if the connection ever breaks? It looks like there's nothing to re-connect

nishanttotla: @vikstrous Good point, this must be handled by the discovery backend. Perhaps we should check for connection errors when they happen and reattempt a connection, with some sort of backoff mechanism.

discovery/swarmkit/swarmkit.go
nishanttotla: Open to suggestions here. I commented above that we might want to do something more/different with this case, but I'm not yet sure what. Maybe just information logging if nothing more?

nishanttotla:
@vikstrous regarding your comment about two primaries, I think that part needs to be handled by leader election, which is not a part of this PR. For now, this backend does not support leader election. The proposal in #2774 has some ideas on that front, and that will be follow up work after this PR is done.

vikstrous:
Oh, I guess I misunderstood what "discovery" does. This PR uses swarm mode to keep track of the list of members of the cluster, but not the primary?

nishanttotla:
@vikstrous that's correct, this PR does not yet understand leadership or primaries. The next step will be to make it so that it respects swarm mode leadership.

Essentially, the swarm classic manager should just be an add-on to the swarm mode manager, that is capable of some extra functionality.

discovery/swarmkit/swarmkit.go
nishanttotla: @fbuecklers thanks for sharing your use case. Sounds like you would also benefit if we don't make assumptions around drain nodes. I've removed this check from this PR.

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,160 @@
...
+					errCh <- err
+					continue
+				}
+
+				// Check if the list of nodes has really changed.
+				if !newEntries.Equals(currentEntries) {
+					ch <- newEntries
+				}
+				currentEntries = newEntries
+			case err := <-eventsErrChan:
```
nishanttotla: @vikstrous perhaps in this case is where we should try to reconnect to the event stream?

nishanttotla:
@wsong yes, sorry I missed your comment above. I was planning to do that as a follow up (mostly integration tests), but even if done with this PR I'd like to have agreement on the design/code first.

discovery/swarmkit/swarmkit.go
```diff
@@ -0,0 +1,160 @@
...
+	d.ttl = ttl
+
+	// create a client to connect with the Docker API
+	httpClient, _, err := cluster.NewHTTPClientTimeout("tcp://"+uri, nil, time.Duration(30*time.Second), nil)
+	if err != nil {
+		return err
+	}
+	d.httpClient = httpClient
+
+	apiClient, err := client.NewClient("tcp://"+uri, "", d.httpClient, nil)
```
wsong: Shouldn't we do a version negotiation here?

discovery/swarmkit/swarmkit.go
wsong: Yes, if we disconnect from the event stream somehow we should reconnect.

discovery/swarmkit/swarmkit.go
nishanttotla: You're right. Added a call to `NegotiateAPIVersion`.

discovery/swarmkit/swarmkit.go
nishanttotla: Right now, I've just added a reconnection if any error is received on the error channel for the event stream.

I think this retry should only happen a few (maybe 5) times with some delay. I'll add that, but does this pattern look okay, or should we reconnect to the events stream another way?

discovery/swarmkit/swarmkit.go
```diff
@@ -129,7 +129,8 @@ func (d *Discovery) Watch(stopCh <-chan struct{}) (<-chan discovery.Entries, <-c
 					ch <- newEntries
 				}
 				currentEntries = newEntries
-			case <-eventsChan:
+			case e := <-eventsChan:
+				log.Debugf("SwarmKit discovery: received node event with action: %s", e.Action)
```
wsong: Is this going to be super spammy in big clusters?

discovery/swarmkit/swarmkit.go
```diff
@@ -142,7 +143,18 @@ func (d *Discovery) Watch(stopCh <-chan struct{}) (<-chan discovery.Entries, <-c
 				}
 				currentEntries = newEntries
 			case err := <-eventsErrChan:
-				errCh <- err
+				log.Debugf("SwarmKit discovery: events stream failed with error: %s", err.Error())
```
wsong: Do you think this should be logged at a WARN level instead? It should be pretty rare, right?

discovery/swarmkit/swarmkit.go
wsong: This general approach looks good.

discovery/swarmkit/swarmkit.go
nishanttotla: We only receive events with type `node` (this is ensured by the events filter above). So as long as we don't have a lot of membership changes, this case won't be triggered. Perhaps it's still worth removing since it works fine and is not an error case.

discovery/swarmkit/swarmkit.go
nishanttotla: Yeah it should be rare (most commonly will happen if the node itself goes down). If that happens (and once we have leader election), consequences of node failure will be handled in a different place.

PR #2778, opened 2017-07-25
Add 'no route to host' as extra failure case for engine connection

nishanttotla:


nishanttotla:
Ping @wsong

PR #2780, opened 2017-08-04
Reschedule containers with a new IP when disconnect fails

alexmavr:
When rescheduling containers, it's possible that the `network disconnect` operation may fail for a given network. This may occur due to several reasons such as:
- The network is swarm-scoped but not currently visible by all nodes
- The connected container is not visible to the engine issuing the disconnect operation
- The disconnect operation fails at the network driver layer

When such a disconnect error occurs, the rescheduling logic should clear out the container's previous IP Address fields, so that the follow-up `network connect` operation may allocate a new IP address for the container on the network.

This behavior would allow swarm to perform rescheduling of containers on attachable networks managed by swarm-mode.

### DIFF
--- a/cluster/watchdog.go
+++ b/cluster/watchdog.go
@@ -116,6 +116,14 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 					if err != nil {
 						// do not abort here as this endpoint might have been removed before
 						log.Warnf("Failed to remove network endpoint from old container %s: %v", name, err)
+
+						// When connecting to this network later, avoid
+						// requesting the same IP address.
+						globalNetworks[networkName].IPAddress = ""
+						if globalNetworks[networkName].IPAMConfig != nil {
+							globalNetworks[networkName].IPAMConfig.IPv4Address = ""
+							globalNetworks[networkName].IPAMConfig.IPv6Address = ""
+						}
 					}
 				}
 			}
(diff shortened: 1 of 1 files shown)


cluster/watchdog.go
```diff
@@ -146,23 +154,6 @@ func (w *Watchdog) rescheduleContainers(e *Engine) {
 		// see https://github.com/docker/docker/issues/17750
 		// Add the global networks one by one
 		for networkName, endpoint := range globalNetworks {
-			hasSubnet := false
-			network := w.cluster.Networks().Uniq().Get(networkName)
-			if network != nil {
-				for _, config := range network.IPAM.Config {
-					if config.Subnet != "" {
-						hasSubnet = true
```
nishanttotla: @alexmavr what happens to this Subnet in the cases you have listed above?

alexmavr: The subnet is not mutated in either the previous version or the changes in this PR. What changed, is that in the previous version we would erase the IP address from the container network config IF the network does NOT have subnet specified, as the assumption is that fixed subnets would require fixed IPs. 

However, that condition is flawed for this code block here, as all networks have a subnet allocated at network creation time, and being able to allocate a specific IP address during container creation is not correlated to whether the swarm manager can detect that subnet or not.

Instead, this PR tweaks this logic so that "if the previous network disconnect failed, you cannot attempt to allocate the same IP address". This should be a more concrete condition, as there is a causal relationship between a disconnect operation for a given IP failing, and a network attachment request for that same IP failing in the future.

nishanttotla: This makes sense, thanks for the explanation @alexmavr

PR #2811, opened 2017-10-30
Add --tlsminversion flag

dperny:
Adds a --tlsminversion flag which allows the user to set the minimum TLS version that swarm will use.

There is no automated test for this right now because I don't know how to use `bats` and there isn't even a test for the other TLS flags right now.


nishanttotla:
Closing this in favor of https://github.com/docker/swarm/pull/2817

PR #2812, opened 2017-10-31
Fix No Elected Primary Cluster Manager

dperny:
Blocks calls to ServeHTTP until a primary cluster manager is selected.

Adds logic to route request to ourself if we become the leader while blocking.

/cc @wsong 

Signed-off-by: Drew Erny <drew.erny@docker.com>

### DIFF
--- a/api/replica.go
+++ b/api/replica.go
@@ -1,10 +1,15 @@
 package api
 
 import (
+	// we have to import this as `gocontext` because there is already a type
+	// called `context` declared in primary.go
+	gocontext "context"
+
 	"crypto/tls"
 	"fmt"
 	"net/http"
 	"strings"
+	"sync"
 )
 
 var localRoutes = []string{"/_ping", "/info", "/debug"}
@@ -14,20 +19,98 @@ type Replica struct {
 	handler   http.Handler
 	tlsConfig *tls.Config
 	primary   string
+	// the address of this replica
+	addr string
+	// mu is the mutex used for locking the primary.
+	mu sync.RWMutex
+	// primaryWait is used to block execution of ServeHTTP until the primary is
+	// ready.
+	primaryWait *sync.Cond
+
+	// capture the hijack function for easier unit testing
+	hijack func(*tls.Config, string, http.ResponseWriter, *http.Request) error
 }
 
 // NewReplica creates a new API replica.
-func NewReplica(handler http.Handler, tlsConfig *tls.Config) *Replica {
-	return &Replica{
+func NewReplica(handler http.Handler, tlsConfig *tls.Config, addr string) *Replica {
+	r := &Replica{
 		handler:   handler,
 		tlsConfig: tlsConfig,
+		addr:      addr,
+		hijack:    hijack,
 	}
+	// This seems to be a big confusing, so here's the explanation:
+	// A Cond is used to wait for a certain value to happen. You use it by:
+	//
+	//   1. Acquiring a lock (cond.L.Lock())
+	//   2. Checking the synchronized value
+	//   3. Calling cond.Wait() if the value isn't what you want
+	//      a. wait releases the lock (cond.L.Unlock()) when it is entered
+	//      b. wait reacquires the lock (cond.L.Lock()) when it is left.
+	//   4. Checking the value again (it may have changed!)
+	//   5. Doing your business
+	//   6. Releasing the lock (cond.L.Unlock())
+	//
+	// The issue here is that for optimal performance, because the ServeHTTP
+	// function only reads the primary value, we want to use a RWMutex, so that
+	// getPrimary can acquire a RLock and not block other concurrent handlers,
+	// but SetPrimary can acquire a write lock `Lock` to modify that value.
+	//
+	// NewCond takes a `Locker` interface, which implements `Lock` and
+	// `Unlock`, and doesn't have any conception of RLock/RUnlock. But we need
+	// those methods so that `r.primaryWait.Wait()` releases and reacquires the
+	// correct lock, the Read lock.
+	//
+	// The way we do this is by passing the return of the `RLocker` method.
+	// This returned object has `Lock` and `Unlock` methods which coorespond to
+	// the `RLock` and `RUnlock` methods of the RWMutex. This means that calls
+	// to `Wait` will acquire and release the read lock, not the write lock.
+	//
+	// This way, `SetPrimary` acquires and releases the WRITE lock on mu
+	// (mu.Lock) and `getPrimary` acquires and releases a READ lock on mu
+	// (r.primaryWait.L.Lock).
+	r.primaryWait = sync.NewCond(r.mu.RLocker())
+	return r
 }
 
 // SetPrimary sets the address of the primary Swarm manager
 func (p *Replica) SetPrimary(primary string) {
 	// FIXME: We have to kill current connections before doing this.
+	// this lock on the mutex is a write lock
+	p.mu.Lock()
+	defer p.mu.Unlock()
 	p.primary = primary
+	if p.primary != "" {
+		// broadcast to wake everyone waiting on a nonempty primary
+		p.primaryWait.Broadcast()
+	}
+}
+
+// getPrimary returns the primary if it exists, blocks if it doesn't, and
+// returns an error if the context is canceled. It's packed nicely in a method
+// to cleanly encapsulate synchronized part of ServeHTTP.
+func (p *Replica) getPrimary(ctx gocontext.Context) (string, error) {
+	// lock a read on primary.
+	//
+	// use primaryWait.L just to ensure we're locking and unlocking the correct
+	// lock, so if we change the lock type later, we can't accidentally forget
+	// to update which lock is acquired and released by the Wait.
+	//
+	// see the description in NewReplica for why this is a read lock
+	p.primaryWait.L.Lock()
+	defer p.primaryWait.L.Unlock()
+	// check if the primary is empty, and launch into a wait if it is
+	for p.primary == "" {
+		// give up waiting if the context is canceled, to avoid blocking forever
+		select {
+		case <-ctx.Done():
+			return "", fmt.Errorf("No elected primary cluster manager: %v", ctx.Err())
+		default:
+		}
+		// wait for the primary to become available
+		p.primaryWait.Wait()
+	}
+	return p.primary, nil
 }
 
 // ServeHTTP is the http.Handler.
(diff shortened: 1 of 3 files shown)


api/replica.go
```diff
@@ -14,20 +15,37 @@ type Replica struct {
...
 }
 
 // SetPrimary sets the address of the primary Swarm manager
 func (p *Replica) SetPrimary(primary string) {
 	// FIXME: We have to kill current connections before doing this.
+	p.mu.Lock()
+	defer p.mu.Unlock()
 	p.primary = primary
+	if p.primary != "" {
+		// broadcast to wake everyoen waiting on a nonempty primary
```
wsong: Nitpick: typo on "everyone".

api/replica.go
```diff
@@ -41,12 +59,34 @@ func (p *Replica) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Otherwise, forward.
-	if p.primary == "" {
-		httpError(w, "No elected primary cluster manager", http.StatusInternalServerError)
-		return
+
+	// lock a read on primary. use primaryWait.L just to ensure we're locking
+	// and unlocking the right lock, if something chnages later
```
wsong: Nitpick: typo on "chnages".

api/replica.go
```diff
@@ -14,20 +15,37 @@ type Replica struct {
 	handler   http.Handler
 	tlsConfig *tls.Config
 	primary   string
+	addr      string // the address of this replica
+	mu        sync.RWMutex
```
wsong: We don't seem to be using the `RLock()` method anywhere; is there a reason we need the `RWMutex` instead of a regular mutex?

dperny: We don't call the method directly, but we pass the output of `mu.RLocker` as the argument to `NewCond`, which means that the lock acquired and released by the Cond is a read lock. So down in `ServeHTTP`, we're actually acquiring read locks, despite saying `Lock` and `Unlock`.

wsong: Ah okay, that makes sense. In that case, let's add a comment above line 42 and line 65 indicating that we're getting a read lock and a write lock, respectively.

api/replica.go
```diff
@@ -41,12 +59,34 @@ func (p *Replica) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			httpError(w, "No elected primary cluster manager", http.StatusInternalServerError)
+			p.primaryWait.L.Unlock()
+			return
+		default:
+		}
+		// wait for the primary to become available
+		p.primaryWait.Wait()
 	}
+	primary := p.primary
+	p.primaryWait.L.Unlock()
```
wsong: Add a comment here about how we need to release the lock before we call `hijack` because `hijack` can take an arbitrarily long time.

api/replica_test.go
```diff
@@ -0,0 +1,101 @@
+package api
+
+import (
+	// weird imports for testing
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+
+	// we need to import context under a different name
+	gocontext "context"
```
wsong: Why do we need to alias `context`?

api/replica.go
dperny: i'm factoring the synchronized bits of code out into a separate method so that it's clearer and cleaner.

api/replica_test.go
dperny: because it's a type in this package as well. or a global variable maybe. go vomitted errors, i'd have to double check what error it was.

nishanttotla: Swarm defines its own `context` struct (this was before contexts were a thing in Golang), which might explain the errors.

dperny: there's a type `context` declared in `primary.go`, toward the top.

api/replica.go
```diff
@@ -14,20 +19,66 @@ type Replica struct {
 	handler   http.Handler
 	tlsConfig *tls.Config
 	primary   string
+	addr      string // the address of this replica
```
nishanttotla: nit: can you move the comment above the line, to be consistent with others?

api/replica_test.go
wsong: I realize this is a bit of a pain, but can you just rename the struct in `primary.go` instead? It'll make this code more readable.

api/replica_test.go
```diff
@@ -0,0 +1,101 @@
...
+	return c.Context.Done()
+}
+
+// Checked returns the number of times Done has been called
+func (c *ContextCheckedEnsurer) Checked() int64 {
+	return atomic.LoadInt64(&(c.checked))
+}
+
+// TestReplicaBlockedPrimary checks that an empty primary causes ServeHTTP to
+// block until it's the primary is no longer empty
```
nishanttotla: nit: until its ~the~ primary is no longer...

nishanttotla:
Overall code LGTM (seems to do what we discussed yesterday).

I still need to more carefully look at the test you have created. We also need integration testing for this, either as bats tests or just manually testing out the offending scenarioo.

api/replica_test.go
nishanttotla: I'm fine with doing that, but it exists in a lot of places, and might just make this PR harder to review/unnecessarily cluttered. Maybe let's do it as a follow up?

wsong: All right, let's do it in a follow-up PR.

api/replica.go
```diff
@@ -14,20 +19,66 @@ type Replica struct {
...
+	// mu is the mutex used for locking the primary.
```
anshulpundir: nit: the primary => the primary swarm manager ?

api/replica.go
```diff
@@ -14,20 +19,66 @@ type Replica struct {
...
+	mu sync.RWMutex
```
anshulpundir: nit: add 'primary' prefix to make it consistent with 'primaryWait' below ?

api/replica.go
```diff
@@ -14,20 +19,66 @@ type Replica struct {
...
-func NewReplica(handler http.Handler, tlsConfig *tls.Config) *Replica {
-	return &Replica{
+func NewReplica(handler http.Handler, tlsConfig *tls.Config, addr string) *Replica {
+	r := &Replica{
 		handler:   handler,
 		tlsConfig: tlsConfig,
+		addr:      addr,
+		hijack:    hijack,
 	}
+	r.primaryWait = sync.NewCond(r.mu.RLocker())
```
anshulpundir: nit: can this be done inline in the Replica{} init above ?

api/replica.go
```diff
@@ -14,20 +19,66 @@ type Replica struct {
...
+	p.mu.Lock()
+	defer p.mu.Unlock()
 	p.primary = primary
+	if p.primary != "" {
+		// broadcast to wake everyone waiting on a nonempty primary
+		p.primaryWait.Broadcast()
+	}
+}
+
+// getPrimary returns the primary if it exists or an error if the context
```
anshulpundir: nit: getPrimary returns the primary if it exists or block until the context is done.

api/replica.go
```diff
@@ -14,20 +19,66 @@ type Replica struct {
...
+	r.primaryWait = sync.NewCond(r.mu.RLocker())
+	return r
 }
 
 // SetPrimary sets the address of the primary Swarm manager
 func (p *Replica) SetPrimary(primary string) {
 	// FIXME: We have to kill current connections before doing this.
+	// this lock on the mutex is a write lock so that we can write the primary
+	// out safely
+	p.mu.Lock()
```
anshulpundir: use the primaryWait to lock/unlock to keep it consistent w below ?

api/replica.go
```diff
@@ -14,20 +19,66 @@ type Replica struct {
...
 	p.primary = primary
+	if p.primary != "" {
+		// broadcast to wake everyone waiting on a nonempty primary
+		p.primaryWait.Broadcast()
+	}
+}
+
+// getPrimary returns the primary if it exists or an error if the context
+// finishes. private method so that we can cleanly separate out the synchronized
+// bits
```
anshulpundir: nit: I'm not sure I fully understand the last part.

api/replica.go
nishanttotla: I believe it's convention to use `mu` since it's clear what it's doing based on who owns it.

api/replica_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+// ContextCheckedEnsurer is a wrapper around context that allows the user to
+// ensure that the Done method has been called
+type ContextCheckedEnsurer struct {
+	gocontext.Context
+	// checked is the number of times Done has been called
+	checked int64
+}
+
+// Done increments the "checked" value and
```
nishanttotla: @dperny is this comment incomplete?

api/replica_test.go
```diff
@@ -0,0 +1,101 @@
...
+	// defer cancel so that we don't leave a dangling goroutine if the tests
+	// run for a long time
+	defer cancel()
+	ctx := &ContextCheckedEnsurer{rctx, 0}
+	// switch the request context for a ContextCheckedEnsurer so we can
+	req = req.WithContext(ctx)
+
+	rec := httptest.NewRecorder()
+
+	// create an empty Response
```
nishanttotla: nit: not sure if this comment should be here.

api/replica_test.go
```diff
@@ -0,0 +1,101 @@
...
+		// yeild the proc so that ServeHTTP has a chance to run
+	}
+	// make sure hijack hasn't been called
+	select {
+	case <-called:
+		t.Fatal("ServeHTTP did not block")
+	default:
+	}
+
+	// update the primary to an empty value and make sure that doesn't trigger
```
nishanttotla: @dperny below here, we're actually updating primary to a non-empty value. Not sure I follow this part.

api/replica.go
dperny: No, because this is is a write lock on the mutex, but `primaryWait.L` is a read locker.

api/replica.go
dperny: No, because we need the replica's mutex to the the `RLocker`.

api/replica.go
anshulpundir: Ahh right, I missed that.

dperny:
@nishanttotla @anshulpundir rebased, fixed comment typos, and added a really big comment explaining the use of the `RWMutex` with the `Cond`.

api/replica.go
```diff
@@ -14,20 +19,98 @@ type Replica struct {
...
-func NewReplica(handler http.Handler, tlsConfig *tls.Config) *Replica {
-	return &Replica{
+func NewReplica(handler http.Handler, tlsConfig *tls.Config, addr string) *Replica {
+	r := &Replica{
 		handler:   handler,
 		tlsConfig: tlsConfig,
+		addr:      addr,
+		hijack:    hijack,
 	}
+	// This seems to be a big confusing, so here's the explanation:
```
anshulpundir: Not sure if this was because of my comment, but most of the RW lock and condition variable is pretty standard stuff around synchronization. While I'm not sure if its needed, I'm OK leaving it in there.

My question was more around how golang's mutexes and condition variables were implemented and I didn't realize that the condition variable was only associated with the reader lock from the RW lock.

dperny: Ah but my comment did answer your question, didn't it? The goal of the comment was to answer every question anyone could ever have.

This was supposed to say  "a bit confusion" and not "a big confusing" but the typo is small enough and the result coherent enough that i don't really feel like pushing up another change, if you don't want to make me.

anshulpundir: Yep, its fine.

nishanttotla:
@dperny @vikstrous how did the testing on this PR turn out?

marcusmartins:
DTR has approved the changes: https://github.com/docker/orca/pull/10553

PR #2814, opened 2017-11-03
Only check healthy engines when doing image inspects

wsong:


wsong:
Ping @nishanttotla for review

Issue #2815, opened 2017-11-05
API /networks/{network_id}/connect sometimes return 404 page not found


panduora:
I created a very simple network named `r`, and when I called swarm network connect API, it always return `404 page not found`, however when I called the docker API, it's OK.

After I turn on the debug message, I found the API `/networks/{network_id}/connect` have become `netwo{network.ID}ks/r/connect`, so it turns to be `404 page not found`.

I checked the related code and found [this]( https://github.com/docker/swarm/blob/master/api/handlers.go#L1160):
```
// Set the network ID in the proxied URL path.
r.URL.Path = strings.Replace(r.URL.Path, networkid, network.ID, 1)
```
The docker remote network API called with network name, so I think it's OK to remove this code?

PR #2816, opened 2017-11-05
Modify the way of replacing networkid

panduora:
this will fix https://github.com/docker/swarm/issues/2815


panduora:
@nishanttotla Thanks for your feedback, I've change the way of replacing network id, how about now?

PR #2823, opened 2017-11-29
Add a timeout in RefreshContainers to avoid hanging connections

wsong:


wsong:
Ping @jlhawn @alexmavr for review

cluster/engine.go
```diff
@@ -728,7 +728,9 @@ func (e *Engine) RefreshContainers(full bool) error {
 		All:  true,
 		Size: false,
 	}
-	containers, err := e.apiClient.ContainerList(context.Background(), opts)
+	ctx, cancel := context.WithTimeout(context.TODO(), 30*time.Second)
```
alexmavr: The timeout value could be `requestTimeout` which is already a const in this file set to 30 seconds. That constant reflects the timeout for any request sent to an engine, so it's appropriate for this context

wsong: Good call. Done.

wsong:
Ping @nishanttotla

PR #2826, opened 2017-12-19
Address concurrency issue

dani-docker:
RefreshContainers is not thread safe and can overwrite _e.containers_ map with a dated list of containers. 
The issue can be observed during high concurrent API calls to **create** and **start** containers.
The **start** container call fails with "container ID not found".

In details: 
(Pre-req: RefreshContainers_ and _CreateContainer_ are both updating the same engine). 
Flow:
- _RefreshContainers_ is first called (issue happens with argument True or False, see below for more details) 
- Right after _RefreshContainers_ gets the list of Containers from the engine, _CreateContainer_ generates container **A** on the same worker engine and adds it to the containers map
- At this point, the list of containers _RefreshContainers_ has is missing **A** and before it returns, it overwrites the map with its "dated" list; 
- result, Container A is missing/removed from the list

Eventually, on the next _RefreshContainers_ the list of containers recover any missing container.

Ideally, refreshing the list of containers should be a "stop the world" operation, especially that the time of execution of this call is relatively slow. 
There are 2 areas that can cause the delay and the container list to be inconsistent:

1- When **full** is false and while getting the ContainerList . 
This represents the majority of what I have seen and can be mitigated with the current PR

& 2- When **full** is true and while iterating and inspecting each container in updateContainer.
Wrapping a mutex lock here, can be costly, without refactoring, not sure what is the best way to address it.

Repro:

-Run a script (ex: python) that calls container create/start API in a multithreaded fashion against a cluster ex: 3 managers 2 workers
-Some of the start calls fails with container not found (the failure is very random as it has to be tied to   _RefreshContainers_ running at the same time.

You can also start your swarm with **refresh-on-node-filter** and give the worker name as its value, next run containers/json api with filters node=<your_worker_name> multiple times in parallel with the above script

**N.B.: See later comments, solution changed to address all concurrency issue related to _RefreshContainers_**

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -315,9 +315,11 @@ func (e *Engine) HealthIndicator() int64 {
 }
 
 // setState sets engine state
-func (e *Engine) setState(state engineState) {
-	e.Lock()
-	defer e.Unlock()
+func (e *Engine) setState(state engineState, lockSet bool) {
+	if !lockSet {
+		e.Lock()
+		defer e.Unlock()
+	}
 	e.state = state
 }
 
@@ -354,9 +356,11 @@ func (e *Engine) ValidationComplete() {
 }
 
 // setErrMsg sets error message for the engine
-func (e *Engine) setErrMsg(errMsg string) {
-	e.Lock()
-	defer e.Unlock()
+func (e *Engine) setErrMsg(errMsg string, lockSet bool) {
+	if !lockSet {
+		e.Lock()
+		defer e.Unlock()
+	}
 	e.lastError = strings.TrimSpace(errMsg)
 	e.updatedAt = time.Now()
 }
@@ -370,7 +374,7 @@ func (e *Engine) ErrMsg() string {
 
 // HandleIDConflict handles ID duplicate with existing engine
 func (e *Engine) HandleIDConflict(otherAddr string) {
-	e.setErrMsg(fmt.Sprintf("ID duplicated. %s shared by this node %s and another node %s", e.ID, e.Addr, otherAddr))
+	e.setErrMsg(fmt.Sprintf("ID duplicated. %s shared by this node %s and another node %s", e.ID, e.Addr, otherAddr), false)
 }
 
 // Status returns the health status of the Engine: Healthy or Unhealthy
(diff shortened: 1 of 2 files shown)


dani-docker:
@wsong 
Yes, definitely possible, however, in my testing, this addressed the issue until I started explicitly calling refreshengines with the flag set to true
As I mentioned in earlier, ideally, the whole method should be synchronized, but not sure about the cons ...

wsong:
It shouldn't be that difficult to refactor `RefreshContainers` to hold the lock the whole time. Just update the `updateContainer` method to take in a new boolean arg called `shouldLock` and pass in `false` in `RefreshContainers`. If `shouldLock` is false, `updateContainers` should not attempt to acquire the lock.

dani-docker:
Right, this is doable, however, it still does not address the concurrency issue 100% as the inspect block in `updateContainers` is not holding the lock, so this will be impacted when full is set to true. Unless you want me to expand the RLock to cover the whole block?

https://github.com/dani-docker/swarm/blob/36a4740b54b0449d8592ef2f945095f365935f05/cluster/engine.go#L821-L850

wsong:
That's fine; when `updateContainer` gets called in `RefreshContainers`, the lock will already be held (because `RefreshContainers` will grab the lock right when it starts).

When `updateContainer` gets called as part of `refreshContainer`, it's only updating a single container, so there's no risk of it wiping out the rest of the `e.containers` map.

dani-docker:
yes, I missed the part that lock is already set , makes perfect sense! 
I will update and run few tests and report back. Thanks @wsong !

dani-docker:
PR updated based on the latest comments and some additional work to address subcalls that tries to lock the mutex .
In summary, this solution addresses any race condition related to _RefreshContainers_ .
Fix has been tested using the steps detailed in the PR description, over and over, with no issues!

nishanttotla:
@dani-docker thanks for submitting this PR! 

There's a merge conflict, could you please fix that? Also, let's squash the two commits into one.

dani-docker:
@nishanttotla 
All done, please sanity check the squash and rebase, Thanks!
Looks like the builds are failing to find swarm-prs-stable677 container.

Thanks!

nishanttotla:
Thanks @dani-docker! I'll restart the build, something seems to have gone wrong.

nishanttotla:
@dani-docker the CI is failing:

```
cluster/engine.go:748: too many arguments in call to e.CheckConnectionErr
```

Can you fix that?

dani-docker:
@nishanttotla 
Apologies for the typo, commit is updated.
Thanks!

PR #2827, opened 2017-12-19
Update docker/libkv dependency

jlhawn:
The latest libkv changes include some fixes to the Lock
logic which may resolve some of the "no elected primary
cluster manager" issues - it may have been due to not
detecting that the lock was freed.


jlhawn:
@wsong @nishanttotla PTAL

I'm not sure why the VNDR decided to remove some of the `sys/unix` files :man_shrugging:

PR #2832, opened 2018-01-10
Change 'volume ls' name format to enginename/volumename for non-local drivers

dani-docker:
I am not sure why existing code limited the engine/volume naming convention to `local` drivers only. The same should apply to volume plugins, otherwise, there will be no way to tell which engine the non-local volume belongs to (ofcourse, unless you filter per specific node)

In case I am missing some design cases, ping @vieux who committed the original fix.
ping @nishanttotla @wsong as if this got accepted, it will make #2831 implementation of dangling much cleaner. 

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -366,20 +366,18 @@ func getVolumes(c *context, w http.ResponseWriter, r *http.Request) {
 		}
 
 		tmp := (*volume).Volume
-		if tmp.Driver == "local" {
-			// Check if the volume matches any node filters
-			found = false
-			for _, node := range nodes {
-				if volume.Engine.Name == node {
-					found = true
-					break
-				}
+		// Check if the volume matches any node filters
+		found = false
+		for _, node := range nodes {
+			if volume.Engine.Name == node {
+				found = true
+				break
 			}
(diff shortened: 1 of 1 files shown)


trapier:
Motion to drive this behavior off of `Scope` rather than `Driver`.  `"Scope": "local"` should have hostname prefix.  `"Scope": "global"` should not (and should only appear once in `volume ls` output, but that's a different discussion).

> Supported scopes are `global` and `local`. Any other value in Scope will be ignored, and `local` is used. Scope allows cluster managers to handle the volume in different ways. For instance, a scope of `global`, signals to the cluster manager that it only needs to create the volume once instead of on each Docker host. More capabilities may be added in the future.

Ref: https://docs.docker.com/v17.09/engine/extend/plugins_volume/#volumedrivercapabilities

thaJeztah:
@trapier so IIUC, this pull request is incorrect, and should be closed?

trapier:
My last comment indicates what _should_ happen, but if my recollection serves, `Scope` was not use that way by at least one driver.  I don't recall which.  Whether or not this PR should be closed is probably a better question for its author.

PR #2837, opened 2018-01-25
Fix possible errant time computation

dperny:
In cluster/engine.go, we compute a delta between the manager and the engine. However, we do several blocking operations between the time we get the engine info, and the time we get the manager system. If those blocking operations take an abnormally long time, the user will be warned of a long time split, and the delta will be incorrect.

This patch moves computation of the delta to immediately after the Info call, so that the subsequent blocking operations do not affect it.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -475,6 +475,20 @@ func (e *Engine) updateSpecs() error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


nishanttotla:
Ping @wsong @vieux

PR #2840, opened 2018-02-07
Fail VolumeCreate requests if they fail on any node

wsong:
Also log failures


wsong:
Ping @nishanttotla

wsong:
This was causing some confusion with third-party volume drivers, where volume creates might succeed on one node but fail on others due to problems with a remote volume driver backend.

PR #2846, opened 2018-03-09
Address concurrency issue (Refactor)

dani-docker:
This PR refactors the fix made in https://github.com/docker/swarm/pull/2826
After testing the fix in the real world, with over 500 nodes and thousands of active containers, we noticed that any hiccup or delays on one node, can cause the `RefreshContainers` to slow down and hold the lock while waiting for API calls to return. 
This issue can be seen with read only calls, like `docker info` which iterates through the list of containers but hang trying to acquire the engine lock first while `RefreshContainers` is running.

A quick test:
- on a 3 nodes cluster (1 manager 2 workers)
- ssh to a worker node and add latency/throttle the connection on eth0
`sudo tc qdisc add dev eth0 root netem delay 5s` in my case, I added a ridiculous 5 seconds
- Force the `RefreshEngines` call by running the following against swarm classic
`docker ps -f name="^ucp-"`
- While waiting for the above to return, run `docker info` and noticed that it's getting blocked while waiting for  `docker ps -f name="whatever you set ContainerNameRefreshFilter to"` to return (edited)

The new fix remove the locks from API calls and used a clever logic proposed by @wsong to address the original issue in https://github.com/docker/swarm/pull/282

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -315,11 +315,9 @@ func (e *Engine) HealthIndicator() int64 {
 }
 
 // setState sets engine state
-func (e *Engine) setState(state engineState, lockSet bool) {
-	if !lockSet {
-		e.Lock()
-		defer e.Unlock()
-	}
+func (e *Engine) setState(state engineState) {
+	e.Lock()
+	defer e.Unlock()
 	e.state = state
 }
 
@@ -356,11 +354,9 @@ func (e *Engine) ValidationComplete() {
 }
 
 // setErrMsg sets error message for the engine
-func (e *Engine) setErrMsg(errMsg string, lockSet bool) {
-	if !lockSet {
-		e.Lock()
-		defer e.Unlock()
-	}
+func (e *Engine) setErrMsg(errMsg string) {
+	e.Lock()
+	defer e.Unlock()
 	e.lastError = strings.TrimSpace(errMsg)
 	e.updatedAt = time.Now()
 }
@@ -374,7 +370,7 @@ func (e *Engine) ErrMsg() string {
 
 // HandleIDConflict handles ID duplicate with existing engine
 func (e *Engine) HandleIDConflict(otherAddr string) {
-	e.setErrMsg(fmt.Sprintf("ID duplicated. %s shared by this node %s and another node %s", e.ID, e.Addr, otherAddr), false)
+	e.setErrMsg(fmt.Sprintf("ID duplicated. %s shared by this node %s and another node %s", e.ID, e.Addr, otherAddr))
 }
 
 // Status returns the health status of the Engine: Healthy or Unhealthy
@@ -385,11 +381,9 @@ func (e *Engine) Status() string {
 }
 
 // incFailureCount increases engine's failure count, and sets engine as unhealthy if threshold is crossed
-func (e *Engine) incFailureCount(lockSet bool) {
-	if !lockSet {
-		e.Lock()
-		defer e.Unlock()
-	}
+func (e *Engine) incFailureCount() {
+	e.Lock()
+	defer e.Unlock()
 	e.failureCount++
 	if e.state == stateHealthy && e.failureCount >= e.opts.FailureRetry {
 		e.state = stateUnhealthy
@@ -412,29 +406,23 @@ func (e *Engine) UpdatedAt() time.Time {
 	return e.updatedAt
 }
 
-func (e *Engine) resetFailureCount(lockSet bool) {
-	if !lockSet {
-		e.Lock()
-		defer e.Unlock()
-	}
+func (e *Engine) resetFailureCount() {
+	e.Lock()
+	defer e.Unlock()
 	e.failureCount = 0
 }
 
-func (e *Engine) CheckConnectionErr(err error) {
-	e.checkConnectionErr(err, false)
-}
-
 // CheckConnectionErr checks error from client response and adjusts engine healthy indicators
-func (e *Engine) checkConnectionErr(err error, lockSet bool) {
+func (e *Engine) CheckConnectionErr(err error) {
 	if err == nil {
-		e.setErrMsg("", lockSet)
+		e.setErrMsg("")
 		// If current state is unhealthy, change it to healthy
 		if e.state == stateUnhealthy {
 			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Infof("Engine came back to life after %d retries. Hooray!", e.getFailureCount())
 			e.emitEvent("engine_reconnect")
-			e.setState(stateHealthy, lockSet)
+			e.setState(stateHealthy)
 		}
-		e.resetFailureCount(lockSet)
+		e.resetFailureCount()
 		return
 	}
 
@@ -444,9 +432,9 @@ func (e *Engine) checkConnectionErr(err error, lockSet bool) {
 		// in engine marked as unhealthy. If this causes unnecessary failure, engine
 		// can track last error time. Only increase failure count if last error is
 		// not too recent, e.g., last error is at least 1 seconds ago.
-		e.incFailureCount(lockSet)
+		e.incFailureCount()
 		// update engine error message
-		e.setErrMsg(err.Error(), lockSet)
+		e.setErrMsg(err.Error())
 		return
 	}
 	// other errors may be ambiguous.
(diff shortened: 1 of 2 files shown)


cluster/engine.go
```diff
@@ -741,31 +727,59 @@ func (e *Engine) refreshVolume(IDOrName string) error {
...
+		for i, cachedID := range s {
+			if cachedID == id {
+				return append(s[:i], s[i+1:]...)
+			}
+		}
+		return s
+	}
+	// Remove valid containers, whatever left in currentContainerIDs should be deleted from e.containers
+	for _, c := range containers {
+		currentContainerIDs = removeContainer(currentContainerIDs, c.ID)
```
wsong: It would be a little easier to read (and slightly more efficient) if you created a new array called `missingContainerIDs` or something. Then just iterate through `currentContainerIDs`, and for each ID, if it's not in `containers`, add it to `missingContainerIDs`.

dani-docker: After what I have witnessed in large deployments, I totally agree, performance should be taking with extra care. I compared 3 different approaches. 
- "Slicing" as I initially implemented 
- "nested iteration" 
- "building a map out of slice and query it"

I forced the worst case scenario on all 3 and here is what I got:
https://gist.github.com/dani-docker/897271ce9dead777ac518294cb03d82d

nestedIteration took 1.263614705s 
mapCheck took 3.755131ms 
slicing took 787.9808ms 

Are you ok with the map approach?

dani-docker:
Worse to mention this in case someone ran into it.
So after opening the lock on the `RefreshContainers`, during the following stress test.
- Ran a test harness that creates 800 containers in 40 concurrent thread.
- At the same time, invoked the `RefreshContainers` by calling `for i in {1..500}; do docker ps -f name="whatever your ContainerNameRefreshFilter is set to"; done`

I see the below errors.
Those are caused on concurrent deletes that are part of the test harness cleanup, so by the time we call `updateContainer` the container is getting deleted, hence the error.
It is benign, but worth noting in case it can cause a bigger issue.
 
```
time="2018-03-10T04:22:19Z" level=error msg="Unable to update state of container \"de21fada7cf50d6ef9b105a1fafe121c7b616b1d12a2793909a9eaaae7ee5a06\": Error: No such container: de21fada7cf50d6ef9b105a1fafe121c7b616b1d12a2793909a9eaaae7ee5a06" id="XYT3:LNNK:JGPX:RZEH:GJSD:2FE7:LPY2:V753:D2HV:LVEB:BEO2:KSQ6|10.0.0.132:12376" name=dani-dev-ucp-226-ubuntu-2
time="2018-03-10T04:22:19Z" level=error msg="Unable to update state of container \"96e83ebb72b5ee4c81314fa7ba882065496b4f11bddf512ce425e3fe8c5a6f8f\": Error: No such container: 96e83ebb72b5ee4c81314fa7ba882065496b4f11bddf512ce425e3fe8c5a6f8f" id="XYT3:LNNK:JGPX:RZEH:GJSD:2FE7:LPY2:V753:D2HV:LVEB:BEO2:KSQ6|10.0.0.132:12376" name=dani-dev-ucp-226-ubuntu-2
time="2018-03-10T04:22:28Z" level=error msg="Unable to update state of container \"83dd1059de40e22ca9dc859b6fbb574712798739ce06b4324ace8719fcafab6b\": Error: No such container: 83dd1059de40e22ca9dc859b6fbb574712798739ce06b4324ace8719fcafab6b" id="Z4ZG:JQDK:CWQM:ZJFO:ILI2:XNKJ:WKTZ:ISZU:VRT2:T2LH:GPK2:RO3D|10.0.0.159:12376" name=dani-dev-ucp-226-ubuntu-0
time="2018-03-10T04:22:28Z" level=error msg="Unable to update state of container \"3c47618162d54847fcc2edfbda393b1796e36e7c02dc1347776af5e4a6f55352\": Error: No such container: 3c47618162d54847fcc2edfbda393b1796e36e7c02dc1347776af5e4a6f55352" id="Z4ZG:JQDK:CWQM:ZJFO:ILI2:XNKJ:WKTZ:ISZU:VRT2:T2LH:GPK2:RO3D|10.0.0.159:12376" name=dani-dev-ucp-226-ubuntu-0
```

@wsong if you can ack the above

cluster/engine.go
wsong: Yeah, sounds fine. Thank you for doing a benchmark.

wsong:
Those "Unable to update state" error messages are going to cause a ton of spam in user workloads where they create and delete containers very rapidly.

Perhaps we should check the error message and see if it contains "No such container"; if it does, then we should just ignore the error.

Alternately, if we want to be a little more conservative, when we get a "No such container" error, we could grab the read lock and check `e.containers` to see if the container ID is still in there; if it's not, we can assume a user deleted the container. If it is still in `e.containers` but we got a "No such container" error, we can log an error then.

dani-docker:
Just to be clear the "Unable to update state"  is not a result of the initial changes made in this PR.
I agree, I will implement the "conservative" approach and checking `e.containers` before eating the exception/err, you ok if we handle it in a separate PR so we can differentiate the 2 issues?

wsong:
I think it would make more sense in this PR, if it's easy to add.

dani-docker:
ok, adding it now and running some test

dani-docker:
@wsong 
Fixed the noisy error message and pushed it in its own commit, so it can be easily reviewed.
If a squash is needed for merge, please let me know.

cluster/engine.go
```diff
@@ -739,31 +727,69 @@ func (e *Engine) refreshVolume(IDOrName string) error {
...
 	for _, c := range containers {
-		mergedUpdate, err := e.updateContainer(c, merged, full, true)
+		mergedUpdate, err := e.updateContainer(c, merged, full)
 		if err != nil {
-			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Errorf("Unable to update state of container %q: %v", c.ID, err)
+			if strings.Contains(err.Error(), "No such container") {
+				// The container might have been deleted after ContainerList was returned and before reaching here
+				// Confirm it's benign by checking the e.containers
+				e.RLock()
+				for containerID := range e.containers {
```
wsong: `e.containers` is a map, right? So can't you just do `_, ok := e.containers[c.ID]` to check if `c.ID` is in `e.containers`?

dani-docker: yep :( another oversight, I just pushed an updated.
I will run a test with this changed and confirm everything is ok

cluster/engine.go
```diff
@@ -739,31 +727,66 @@ func (e *Engine) refreshVolume(IDOrName string) error {
...
+
 	merged := make(map[string]*Container)
 	for _, c := range containers {
-		mergedUpdate, err := e.updateContainer(c, merged, full, true)
+		mergedUpdate, err := e.updateContainer(c, merged, full)
 		if err != nil {
-			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Errorf("Unable to update state of container %q: %v", c.ID, err)
+			if strings.Contains(err.Error(), "No such container") {
+				// The container might have been deleted after ContainerList was returned and before reaching here
+				// Confirm it's benign by checking the e.containers
```
wsong: Make sure you update this comment with more detail.

dani-docker:
done and verified the latest changes

rchourasia:
@wsong : just spoke to @dani-docker and he feels PR is ready to be merged. Is there something else pending? I am trying to make a determination if this can be included in 2.2.8. wdyt?

cc @dani-docker @marcusmartins

cluster/engine.go
```diff
@@ -739,31 +727,67 @@ func (e *Engine) refreshVolume(IDOrName string) error {
...
 	ctx, cancel := context.WithTimeout(context.TODO(), requestTimeout)
 	defer cancel()
 	containers, err := e.apiClient.ContainerList(ctx, opts)
-	e.checkConnectionErr(err, true)
+	e.CheckConnectionErr(err)
 	if err != nil {
 		return err
 	}
 
+	var missingContainerIDs []string
```
nishanttotla: @dani-docker can you add a comment here just briefly describing the logic and reasoning for doing it this way? It's fairly clear in the context of the PR, but it would help in the future.

PR #2858, opened 2018-05-03
Update copyright time to 2018

lioncruise:
Update copyright time from 2016 to 2018.


thaJeztah:
Change looks okay to me, given that we already used a range here

@lioncruise can you update your sign-off to use your full name, instead of only `Shijun` ?

lioncruise:
@thaJeztah Thanks for your remind! I have fixed the commit's sign-off to my full name.

PR #2860, opened 2018-05-05
Force resume refreshLoop when engine status changes to healthy

dani-docker:
Swarm classic randomly shows stale containers, the data eventually gets updated, but that would take hours (depending on `failureCount`). 
Below is the scenario and repro steps needed to get into the bad state.

The problem is that the refreshLoop wait is proportional to the `failureCount`, even after a successful connection, the loop might take hours to resume.
This fix forces the loop to resume after an engine becomes `Healthy`.
Note: I made some changes into `CheckConnectionErr` to avoid race condition and unnecessary blocks.

## Summary
1- simulate a connection failure to a worker node using iptables
2- generate enough failure counts ( for ex: up to 415), which will cause the refresh loop to wait for around 3.5 hours. The 415 failures causes the engine to be marked as `unhealthy`
3- start a container
4- Open the firewall, this causes event monitor to resume and eventually runs `CheckConnectionErr` due to false/positive event `exec_start` that runs `refreshContainer`
5- now that the `refreshLoop` is hung/waiting for 3.5 hours, simulate another connection failure to the same engine, now go ahead and delete the container created in 3 and reopen the firewall (the key here is not to get the engine `unhealthy`, otherwise an `engine_reconnect` event is emitted and this will result of full containers refresh).
6- 1,2,3,4,5 leads to stale data, container 3 still showing, although it's deleted.

## Detail
For simplicity, these steps can be reproduced on a cluster of one manager and one or more worker
1- On the worker node, simulate engine disconnect by running the following iptable rules (important to use `--reject-with tcp-rese`):
`sudo iptables -I FORWARD -p tcp --dport 12376 -j REJECT --reject-with tcp-reset`
`sudo iptables -I FORWARD -p tcp --dport 2376 -j REJECT --reject-with tcp-reset`
2- On the manager node configure your docker client to query swarm classic directly
3- Assuming you have `container-name-refresh-filter` set to `test123` . Run the following `docker ps` with a the filter that forces a container refresh that eventually fails and increases the failure count (in this example, this will generate 900 failures)
`for i in {1..900}; do docker ps -f name="test123"; done`
4- On the worker node, create a container:
`docker run -d alpine ping 127.0.0.1`
5- Reopen the firewall
`sudo iptables -D FORWARD 1`
`sudo iptables -D FORWARD 1`
and `sudo iptables -L --line-numbers` to confirm that the rules from step 1 are gone 
6- run `docker ps` against swarm classic and confirm that the container in step 4 is listed 
7- go back to the worker node and repeat step 1
8- delete the container created in step 4
9- repeat step 5 to open the firewall
10- run `docker ps` against swarm classic, notice how the deleted container is still listed. Wait few hours until the refresh loop is released, notice how the container list is up to date.

### DIFF
--- a/cluster/engine.go
+++ b/cluster/engine.go
@@ -71,19 +71,22 @@ var (
 	testErrImageNotFound = errors.New("TEST_ERR_IMAGE_NOT_FOUND_SWARM")
 )
 
-// delayer offers a simple API to random delay within a given time range.
+// delayer offers a simple API to random delay within a given time range,
+// and to force resume delays
 type delayer struct {
 	rangeMin time.Duration
 	rangeMax time.Duration
 
-	r *rand.Rand
-	l sync.Mutex
+	resumeCh chan bool
+	r        *rand.Rand
+	l        sync.Mutex
 }
 
 func newDelayer(rangeMin, rangeMax time.Duration) *delayer {
 	return &delayer{
 		rangeMin: rangeMin,
 		rangeMax: rangeMax,
+		resumeCh: make(chan bool, 1),
 		r:        rand.New(rand.NewSource(time.Now().UTC().UnixNano())),
 	}
 }
@@ -417,10 +420,21 @@ func (e *Engine) CheckConnectionErr(err error) {
 	if err == nil {
 		e.setErrMsg("")
 		// If current state is unhealthy, change it to healthy
+		// A lock is needed to avoid race condition, which can lead to block on resumeCh
+		// We should not block on slow operations, ex: emitEvent that ends up calling the engine
+		// instead, we'll check the changed flag which is set to true in the lock block
+		changed := false
+		e.Lock()
 		if e.state == stateUnhealthy {
-			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Infof("Engine came back to life after %d retries. Hooray!", e.getFailureCount())
+			changed = true
+			e.state = stateHealthy
+			log.WithFields(log.Fields{"name": e.Name, "id": e.ID}).Infof("Engine came back to life after %d retries. Hooray!", e.failureCount)
+		}
+		e.Unlock()
+		if changed {
+			// signal the refresh loop so we don't wait too long, especially if the failure count was high
+			e.refreshDelayer.resumeCh <- true
 			e.emitEvent("engine_reconnect")
-			e.setState(stateHealthy)
 		}
 		e.resetFailureCount()
 		return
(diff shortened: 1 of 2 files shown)


dani-docker:
ping @nishanttotla

dani-docker:
PR is updated with a test case.

rchourasia:
ping @marcusmartins for assigning reviewer.

cluster/engine.go
```diff
@@ -417,10 +420,20 @@ func (e *Engine) CheckConnectionErr(err error) {
 	if err == nil {
 		e.setErrMsg("")
 		// If current state is unhealthy, change it to healthy
+		// A lock is needed to avoid race condition, which can lead to block on resumeCh
+		// "changed" flag will take care of slow operations, ex: emitEvent
+		changed := false
+		e.Lock()
```
wsong: Why are you locking here?

dani-docker: I thought of it, but the error message was a key to figure out the issue and saw the race condition possibility here. So figured why not fixing it as well as making sure only one refresh perform when the engine health changes.
What's the concerns with the lock?

wsong: If anything else is holding the lock, `CheckConnectionErr` will get blocked.

dani-docker: If this is an issue then we have a bigger problem to deal with , as subcommands in the same block also holds the lock, ex: `getFailureCount` and `resetFailureCount`

wsong: Fair enough.

cluster/engine.go
```diff
@@ -417,10 +420,20 @@ func (e *Engine) CheckConnectionErr(err error) {
 	if err == nil {
 		e.setErrMsg("")
 		// If current state is unhealthy, change it to healthy
+		// A lock is needed to avoid race condition, which can lead to block on resumeCh
+		// "changed" flag will take care of slow operations, ex: emitEvent
```
wsong: I don't really understand this comment. `changed` doesn't affect how we call `emitEvent` at all.

dani-docker: `changed` which is set to true inside the `if e.state == stateUnhealthy` block and in the thread safe block, will be used to decide whether to run (outside the lock) slow operations : ex:  emitEvent (which ends up calling the engine)

wsong: Can we update this comment to be more clear, then?

dani-docker: Sure, I guess It can be a bit clearer

PR #2901, opened 2018-09-24
Setting the http status code based on known error messages

dani-docker:

Swarm uses the docker client API to communicate with the engine.
As you can see [here](https://github.com/moby/moby/blob/master/client/container_create.go#L45-L51 ) and in case of an error, the `serverResp.statusCode` is not returned.
Currently swarm guesses the return code which makes it inconsistent with the status code documented [here](https://docs.docker.com/engine/api/v1.30/#operation/ContainerCreate) 

This is not limited to `ContainerCreate` but because of the scope of the issue, the fix in this PR only addresses the container create api call only.

The error messages used in the helper function `GetHTTPErrorStatusCode` are from this engine method https://github.com/moby/moby/blob/17.05.x/api/server/httputils/errors.go#L56-L68

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -690,6 +690,8 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 	if err != nil {
 		if strings.HasPrefix(err.Error(), "Conflict") {
 			httpError(w, err.Error(), http.StatusConflict)
+		} else if strings.Contains(err.Error(), "does not exist") {
+			httpError(w, err.Error(), http.StatusNotFound)
 		} else {
 			httpError(w, err.Error(), http.StatusInternalServerError)
 		}


dani-docker:
cc: @wsong

api/utils.go
```diff
@@ -428,3 +428,46 @@ func normalizeEvent(receivedEvent *cluster.Event) ([]byte, error) {
...
+
+	errStr := strings.ToLower(errMsg)
+	for _, status := range []struct {
+		keyword string
+		code    int
+	}{
+		{"does not exist", http.StatusNotFound},
+		{"not found", http.StatusNotFound},
+		{"cannot find", http.StatusNotFound},
+		{"no such", http.StatusNotFound},
```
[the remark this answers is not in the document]
dani-docker: updated

api/utils.go
```diff
@@ -428,3 +428,48 @@ func normalizeEvent(receivedEvent *cluster.Event) ([]byte, error) {
 
 	return data, nil
 }
+
+// GetHTTPErrorStatusCode takes an error, checks a list of known errors,
+// and return the HTTP status code.
+// Returns 500 if there is no match
+// List of errors are extracted from https://github.com/moby/moby/blob/17.05.x/api/server/httputils/errors.go#L56-L68
+// In engine > 18.x, we moved to typed errors check, which is hard to extract the err strings for
+func GetHTTPErrorStatusCode(err error) int {
```
thaJeztah: Wonder if, instead of copying this, we could use the upstream package (possibly would need some small refactoring to prevent pulling in too much code?)

api/handlers.go
```diff
@@ -688,11 +688,8 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 
 	container, err := c.cluster.CreateContainer(containerConfig, name, authConfig)
 	if err != nil {
-		if strings.HasPrefix(err.Error(), "Conflict") {
-			httpError(w, err.Error(), http.StatusConflict)
-		} else {
-			httpError(w, err.Error(), http.StatusInternalServerError)
-		}
+		responseStatus := GetHTTPErrorStatusCode(err)
```
thaJeztah: Note that the example you linked to actually returns a well-defined error for the 404 case; https://github.com/moby/moby/blob/7d62e40f7e4f3c17d229a7687d6fcca5448de813/client/container_create.go#L46-L51

The [`objectNotFoundError`](https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb/client/errors.go#L40-L58) can be converted to a HTTP status code with the `GetHTTPErrorStatusCode ` function; https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb/api/server/httputils/errors.go#L20-L73

Would it work instead to bump the vendored version of github.com/docker/docker ?

cpuguy83:
From the mentioned moby tree about this style of error checking:

> 		// FIXME: this is brittle and should not be necessary, but we still need to identify if
>		// there are errors falling back into this logic.
>		// If we need to differentiate between different possible error types,
>		// we should create appropriate error types that implement the httpStatusError interface.

Note that we scrapped the `httpStatusError` mentioned there as well because it did not work well either (error types that encode an HTTP status code).

api/utils.go
dani-docker: @thaJeztah 
I did look into this approach, but the scope was larger for what we are trying to accomplish for this one request.
I am happy to start using errdefs if there is more need to set the http status code for other errors

dani-docker:
@cpuguy83 

There is no question that the typed check error you mentioned is the correct approach; 
but due to the current state of this project and the actual request (ie: just setting the correct http status code for `CreateContainer`)  I decided to take a simpler route and avoid bumping vendored dependencies and bringing in new packages.

cpuguy83:
In this case "simpler" is simply incorrect and brittle error handling... likely which doesn't even work if these are error messages from from docker.

dani-docker:
@cpuguy83 

This incorrectness is widespread across the swarm code base
@thaJeztah @wsong 
If the goal if to revamp swarm error handling and then I will backup my changes and we can decide on how to proceed.

api/handlers.go
dani-docker: @thaJeztah 
quick correction, the API client swarm is using is dated, it's `imageNotFoundError` instead of `objectNotFoundError`
https://github.com/dani-docker/swarm/blob/b44836b5b6cfd3262c5e7eee7208099d6dcdd018/vendor/github.com/docker/docker/client/container_create.go#L48

dani-docker:
@wsong , @thaJeztah and @trapier had a discussion and we all agree that string error check is NOT the ideal fix and we agreed that the ultimate fix is to use type error check as suggested previously. 

Due to the scope of changes, the error type check will not be implemented in this PR, instead, this PR will follow the same practice already in use across the swarm code base, and simply does a string match to when the image is missing and set the http code accordingly. 

I updated the PR and simplified the change to a single error comparison.

dani-docker:
@thaJeztah 

For this particular change, the test you linked uses docker client which can't validate the fix and bubble up the http status code .
We'll need to expend the bats and helper functions to curl the endpoints and read the http status code.

@thaJeztah , let me know If an integration test is necessary , I can start looking into the scope of this change

thaJeztah:
I'm ok with merging as-is, but it would be good to have a follow up with a test, to prevent future regressions (this code/check being brittle as it is)

PR #2920, opened 2019-01-23
Update vendoring

dperny:
Updates the vendoring and transitive dependencies of `github.com/sirupsen/logrus` and `github.com/docker/docker`. Makes changes to the Swarm code base to accommodate API changes.

Split out from #2919 for ease of review.

PR #2932, opened 2019-03-20
Buildkit support

dperny:
I decided that, instead of wrestling with git, it would be easier for me to make a new PR. This PR includes the changes from #2920 and #2919.

Tested manually, it seems to work

/cc @tonistiigi @tiborvass @dhiltgen, who were all at various points pinged in that thread.


tonistiigi:
LGTM

api/handlers.go
```diff
@@ -26,11 +25,19 @@ import (
 	"github.com/docker/swarm/experimental"
 	"github.com/docker/swarm/version"
 	"github.com/gorilla/mux"
+	log "github.com/sirupsen/logrus"
 )
 
 const (
 	// APIVERSION is the default API version supported by swarm manager
-	APIVERSION = "1.30"
+	APIVERSION = "1.39"
```
tiborvass: Should be 1.40

dperny: fixed

PR #2941, opened 2019-04-30
Strip off node name network prefixes in endpoint configs in networking configs

wsong:
Previously, if you specified something like:

```
docker run --rm \
  --net=mynode/testnet
  --network-alias=foobar \
  alpine
```

Swarm classic would fail because it would try to pass an endpoint config to the
daemon with `mynode/testnet` as the network name

### DIFF
--- a/api/handlers.go
+++ b/api/handlers.go
@@ -18,6 +18,7 @@ import (
 	apitypes "github.com/docker/docker/api/types"
 	containertypes "github.com/docker/docker/api/types/container"
 	dockerfilters "github.com/docker/docker/api/types/filters"
+	"github.com/docker/docker/api/types/network"
 	typesversions "github.com/docker/docker/api/types/versions"
 	volumetypes "github.com/docker/docker/api/types/volume"
 	"github.com/docker/docker/pkg/parsers/kernel"
@@ -698,6 +699,7 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 		httpError(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	containerConfig.NetworkingConfig = stripNodeNamesFromNetworkingConfig(containerConfig.NetworkingConfig, c.cluster.EngineNames())
 
 	container, err := c.cluster.CreateContainer(containerConfig, name, authConfig)
 	if err != nil {
@@ -1644,4 +1646,27 @@ func (h headerFlusher) WriteHeader(status int) {
 // header immediately.
 func postContainersWait(c *context, w http.ResponseWriter, r *http.Request) {
 	proxyContainerAndForceRefresh(c, headerFlusher{w}, r)
+}
+
+// stripNodeNamesFromNetworkingConfig goes through a container's
+// NetworkingConfig and, if a user specified a Swarm classic-style network
+// name in the endpoints config (i.e. <node name>/<network name>), strips off
+// the node name so that the daemon will recognize the network.
+func stripNodeNamesFromNetworkingConfig(n network.NetworkingConfig, nodeNames []string) network.NetworkingConfig {
+	endpointsConfig := map[string]*network.EndpointSettings{}
+	nodeNamesMap := map[string]struct{}{}
+	for _, name := range nodeNames {
+		nodeNamesMap[name] = struct{}{}
+	}
+	for networkName, v := range n.EndpointsConfig {
+		parts := strings.SplitN(networkName, "/", 2)
+		if len(parts) > 1 {
+			if _, ok := nodeNamesMap[parts[0]]; ok {
+				networkName = parts[1]
+			}
+		}
+		endpointsConfig[networkName] = v
+	}
+	n.EndpointsConfig = endpointsConfig
+	return n
 }
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -965,6 +965,15 @@ func (c *Cluster) TotalCpus() int64 {
 	return totalCpus
 }
 
+// EngineNames returns the names of all the engines in the cluster.
+func (c *Cluster) EngineNames() []string {
+	ret := make([]string, len(c.engines))
+	for _, engine := range c.engines {
+		ret = append(ret, engine.Name)
+	}
+	return ret
+}
+
 // Info returns some info about the cluster, like nb or containers / images.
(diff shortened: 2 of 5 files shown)


cluster/config_test.go
```diff
@@ -28,6 +28,18 @@ func TestBuildContainerConfig(t *testing.T) {
...
+		EndpointsConfig: map[string]*network.EndpointSettings{
+			"testnet1":           &network.EndpointSettings{},
+			"nodename/testnet2":  &network.EndpointSettings{},
+			"nodename/test/net3": &network.EndpointSettings{},
+		},
+	}
+	config = BuildContainerConfig(container.Config{}, container.HostConfig{}, networkingConfig)
+	assert.Contains(t, config.NetworkingConfig.EndpointsConfig, "testnet1")
+	assert.Contains(t, config.NetworkingConfig.EndpointsConfig, "testnet2")
+	assert.Contains(t, config.NetworkingConfig.EndpointsConfig, "test/net3")
```
dperny: does `assert.Contains` not mean that the string `"nodename/testnet2"` would pass the assertion?

wsong: `config.NetworkingConfig.EndpointsConfig` is a map, so we're checking for whether or not this exact key is in the map.

dperny:
OK yes that makes sense.

I feel like the CI should not be failing on every single version, though.

wsong:
@dperny  and I looked at the failures and it looks like most of the failures are actually from DinD failing to start. Not sure what's going on there but it seems clearly unrelated to this PR.

dperny:
ok, that's good enough for me. i don't have write permissions, though.

cluster/config.go
```diff
@@ -140,6 +140,19 @@ func BuildContainerConfig(c container.Config, h container.HostConfig, n network.
...
 	}
 
+	// If a user specified a Swarm classic-style network name in the
+	// endpoints config (i.e. <node name>/<network name>), strip off the
+	// node name so that the daemon will recognize the network.
+	endpointsConfig := map[string]*network.EndpointSettings{}
+	for networkName, v := range n.EndpointsConfig {
+		parts := strings.SplitN(networkName, "/", 2)
+		if len(parts) > 1 {
+			networkName = parts[1]
```
thaJeztah: I think this we'll be a backward breaking change, as I can create custom networks having a `/`;

```
 docker network create foo/bar/baz

docker network ls

NETWORK ID          NAME                DRIVER              SCOPE
5ce5654a460e        bridge              bridge              local
81d2f269d9ff        docker_gwbridge     bridge              local
7296e31838b7        foo/bar/baz         bridge              local
26d3887604d2        host                host                local
59e5e41d3a7b        none                null                local
```

wsong: Hmm, I guess we'll have to iterate through and see if `parts[0]` is the name of a node.

wsong:
Okay, I refactored this so that we check the list of node names and make sure that the part we're stripping off is the name of a node. PTAL

api/handlers.go
```diff
@@ -698,6 +698,7 @@ func postContainersCreate(c *context, w http.ResponseWriter, r *http.Request) {
 		httpError(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+	containerConfig.NetworkingConfig = cluster.CleanupNetworkingConfig(containerConfig.NetworkingConfig, c.cluster.EngineNames())
```
thaJeztah: I was a bit on the fence if we should move this into `cluster.CreateContainer()`; that way `CreateContainer` would responsible for dealing with this (and we wouldn't have to modify the Interface to add `EngineNames()`).

I went looking for uses of `CreateContainer()` and the other location it's used is in [`rescheduleContainers`](https://github.com/docker/swarm/blob/0a268e4925fb65c51a13f1482082ca5b4028aeb0/cluster/watchdog.go#L144-L145), which copies the configuration of existing containers, so in that case the nodename would be stripped _twice_

We could consider making it a local utility and move it here, given that it's fixing something specific to this endpoint, but happy to get your thought on that

wsong: I think it makes sense to just strip the node names when the container configs first come into the API; as you point out, we should definitely not be stripping node names in `rescheduleContainers`. I'll move `CleanupNetworkingConfig` (aka `StripNodeNames`) into this file.

cluster/config.go
```diff
@@ -143,6 +143,29 @@ func BuildContainerConfig(c container.Config, h container.HostConfig, n network.
 	return &ContainerConfig{c, h, n}
 }
 
+// CleanupNetworkingConfig goes through a container's NetworkingConfig and, if a
+// user specified a Swarm classic-style network name in the endpoints config
+// (i.e. <node name>/<network name>), strips off the node name so that the
+// daemon will recognize the network.
+func CleanupNetworkingConfig(n network.NetworkingConfig, nodeNames []string) network.NetworkingConfig {
```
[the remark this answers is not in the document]
wsong: Good idea, done.

cluster/config_test.go
```diff
@@ -30,6 +30,22 @@ func TestBuildContainerConfig(t *testing.T) {
 	assert.Len(t, config.Labels, 2)
 }
 
+func TestCleanupNetworkingConfig(t *testing.T) {
+	networkingConfig := network.NetworkingConfig{
+		EndpointsConfig: map[string]*network.EndpointSettings{
+			"testnet1":           &network.EndpointSettings{},
```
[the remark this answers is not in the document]
wsong: Done.

cluster/config_test.go
```diff
@@ -30,6 +30,22 @@ func TestBuildContainerConfig(t *testing.T) {
...
+			"nodename/testnet2":  &network.EndpointSettings{},
+			"nodename/test/net3": &network.EndpointSettings{},
```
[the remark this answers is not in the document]
wsong: Done.

cluster/swarm/cluster.go
```diff
@@ -965,6 +965,15 @@ func (c *Cluster) TotalCpus() int64 {
 	return totalCpus
 }
 
+// EngineNames returns the names of all the engines in the cluster.
+func (c *Cluster) EngineNames() []string {
+	ret := []string{}
```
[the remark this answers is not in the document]
wsong: Done.

thaJeztah:
17.06 is green now; let's merge

PR #2961, opened 2019-09-03
Add better OS detection

dperny:
Adds better detection of the OS type of the engine by getting the supported platforms defined in the manifest and creating an ostype constraint using them.

This is #2959 done the right way.

### DIFF
--- a/cluster/swarm/cluster.go
+++ b/cluster/swarm/cluster.go
@@ -138,6 +138,11 @@ func (c *Cluster) StartContainer(container *cluster.Container) error {
 
 // CreateContainer aka schedule a brand new container into the cluster.
 func (c *Cluster) CreateContainer(config *cluster.ContainerConfig, name string, authConfig *types.AuthConfig) (*cluster.Container, error) {
+	// engines newer than api version 1.30 have a /distribution/{name:.*}/json
+	// endpoint, which can be used to contact a registry and determine the
+	// image platforms. before starting a container, fill in the constraint.
+	c.setOSTypeConstraint(config, authConfig)
+
 	container, err := c.createContainer(config, name, false, authConfig)
 
 	if err != nil {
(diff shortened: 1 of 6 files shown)


cluster/swarm/cluster.go
```diff
@@ -177,6 +182,74 @@ func (c *Cluster) CreateContainer(config *cluster.ContainerConfig, name string,
...
+		for os := range ostypes {
+			osStrings = append(osStrings, fmt.Sprintf("(%s)", os))
+		}
+
+		// then, string join the resulting slice with |
+		osStringsJoined := strings.Join(osStrings, "|")
+
+		// and finally, pack it in between / characters to denote that it's a
+		// regular expression
+		osRegex := fmt.Sprintf("/%s/", osStringsJoined)
```
jlhawn: I didn't know we could have a regex in a constraint. 👍

cluster/swarm/cluster.go
```diff
@@ -177,6 +182,74 @@ func (c *Cluster) CreateContainer(config *cluster.ContainerConfig, name string,
...
+	platforms, err := engine.GetImagePlatforms(config, authConfig)
+	if err != nil {
+		return err
+	}
+
+	// now extract the OSes. use a map to deduplicate, in case the image is
+	// available on several architectures with the same platform.
+	ostypes := map[string]struct{}{}
+	for _, p := range platforms {
+		ostypes[p.OS] = struct{}{}
```
ddebroy: If you are depending on the [OS](https://github.com/moby/moby/blob/8e610b2b55bfd1bfa9436ab110d311f5e8a74dcb//vendor/github.com/opencontainers/image-spec/specs-go/v1/descriptor.go#L51) field alone, it may not work for Windows as Windows images are sensitive to OS versions. If you try to run a 2016 image on a 2019 OS or vice versa, Windows HCS will reject the image and the container won't run. To make this work correctly on Windows, the OS versions need to be matched as well (besides OS type)

zongweidsun:
After this fix in, we will enable test TestRunWindowsContainerOnController() and make sure it passes (https://github.com/docker/orca/blob/master/integration/basic/c1_ww1/acceptance_test.go#L165).

cluster/swarm/cluster.go
dperny: I don't see where the engine reports OS version, though. Swarm relies on the output of `docker info` to discover information like the OS, and `info` does have a particular field that appears to map 1:1 to the OSVersion field in the Platform.

zongweidsun: Then I think this fix is sufficient if we can document the expected behaviors for the customers: 1. we recommend the customers to explicitly specify OS-info in the image manifest (and swarm matches it correctly). 2. we make error messages as useful as possible when mismatch happens.

dperny:
It seems I don't actually have write access on this repo...

